PR #36, opened 2021-12-28
refactor(List): rewrite in TypeScript

### LINKED ISSUE #13: Refactor List to TypeScript

korvin89:
Resolves #13 

Move `List` to typescript.

PS: I mentioned couple of arguable moments in commentaries.


src/components/List/List.d.ts
```diff
@@ -1,72 +0,0 @@
...
-    itemHeight?: number | ((item: T) => number);
-    /** Высота списка элементов (или функция, возвращающая значение высоты для списка). Может быть полезно, чтобы задавать высоту списка динамически. */
-    itemsHeight?: number | ((items: T[]) => number);
-    /** Кастомное имя класса которое будет добавлено к контейнеру элемента */
-    itemClassName?: string;
-    itemsClassName?: string;
-
-    theme?: string;
-    view?: string;
-    tone?: string;
```
korvin89, line 16: theme, view and tone didn`t use

src/components/List/List.tsx
```diff
@@ -1,168 +1,80 @@
...
 
-    state = {
+    state: ListState<T> = {
         items: this.props.items,
         filter: '',
     };
-    componentDidUpdate(prevProps) {
+
+    refFilter = React.createRef<HTMLInputElement | HTMLTextAreaElement>();
+    refContainer = React.createRef<any>();
```
korvin89, line 70: There is no `getWrappedInstance` method in [SortableContainerProps](https://github.com/clauderic/react-sortable-hoc/blob/master/types/index.d.ts#L61). Therefore I used `any`.

src/components/List/List.tsx
```diff
@@ -171,58 +83,57 @@ export class List extends React.Component {
...
-        const container = this.getContainer();
-        if (container) {
-            container.scrollToItem(index);
-        }
-    };
-    activateItem(index, scrollTo = true) {
-        if (scrollTo) {
+
+    activateItem(index?: number, scrollTo = true) {
+        if (index && scrollTo) {
```
korvin89, line 127: I added `index` to condition because of uselessness `scrollToIndex` without it.

src/components/List/List.tsx
```diff
@@ -261,18 +172,71 @@ export class List extends React.Component {
...
     };
-    onFocus = () => {
+
+    private getContainer() {
+        const ref = this.refContainer.current;
+        const wrappedInstance =
+            ref &&
+            'getWrappedInstance' in ref &&
+            typeof ref.getWrappedInstance === 'function' &&
+            ref.getWrappedInstance();
```
korvin89, line 182: I added more honest check for `getWrappedInstance` invocation

src/components/List/List.tsx
```diff
@@ -1,168 +1,80 @@
...
         if (oldIndex !== newIndex) {
             const [item] = list.splice(oldIndex, 1);
             list.splice(newIndex, 0, item);
         }
+
         return list;
     }
-    static findNextIndex(list, index, step) {
+
+    static findNextIndex(list: ListItemData<unknown>[], index: number, step: number) {
```
amje, line 50: Maybe we turn static methods into generics?
```ts
static findNextIndex<T = unknown>(list: ListItemData<T>[], index: number, step: number)
```

src/components/List/List.tsx
```diff
@@ -1,168 +1,80 @@
...
     sortable: false,
     virtualized: true,
-
     deactivateOnLeave: true,
 };
 
-export class List extends React.Component {
-    static moveListElement(list, oldIndex, newIndex) {
+export class List<T = unknown> extends React.Component<ListProps<T>, ListState<T>> {
+    static defaultProps: Partial<ListProps<ListItemData<unknown>>> = listDefaultProps;
```
amje, line 39: Seems there should be T instead of unknown

src/components/List/List.tsx
```diff
@@ -1,168 +1,80 @@
...
-    state = {
+    state: ListState<T> = {
         items: this.props.items,
         filter: '',
     };
-    componentDidUpdate(prevProps) {
+
+    refFilter = React.createRef<HTMLInputElement | HTMLTextAreaElement>();
+    refContainer = React.createRef<any>();
+    blurTimer: NodeJS.Timeout | null = null;
```
amje, line 71: NodeJS??

src/components/List/List.tsx
amje: 0 will be treated as false, better change to `typeof index === 'number'`

src/components/List/List.tsx
```diff
@@ -233,11 +144,11 @@ export class List extends React.Component {
                 break;
             }
             case 'PageDown': {
-                this.handleKeyMove(event, pageSize);
+                this.handleKeyMove(event, pageSize!);
```
amje, line 147: Are we confident there is a value here?

src/components/List/List.tsx
```diff
@@ -261,18 +172,71 @@ export class List extends React.Component {
...
+        }
+    };
+
+    private handleKeyMove(event: React.KeyboardEvent, step: number, defaultItemIndex = 0) {
+        event.preventDefault();
+        const {activeItem = defaultItemIndex} = this.state;
+        this.activateItem(List.findNextIndex(this.state.items, activeItem + step, Math.sign(step)));
+    }
+
+    private onFocus = () => {
```
amje, line 226: lets name handler methods like this: `handle*`, `on*` functions are usually come from component props

src/components/List/List.tsx
```diff
@@ -286,56 +250,71 @@ export class List extends React.Component {
...
         if (typeof itemHeight === 'function') {
             const {items} = this.state;
             return itemHeight(items[index]);
         } else {
             return virtualized ? Number(itemHeight) || 28 : itemHeight;
         }
     };
-    renderItem = ({index, style}) => {
+
+    private renderItem = ({index, style}: {index: number; style?: React.CSSProperties}) => {
```
amje, line 310: Let's put inner render methods right after the main render for convenience

src/components/List/components/ListItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+    itemClassName?: string;
+    sortable?: boolean;
+    sortHandleAlign?: ListSortHandleAlign;
+    style?: React.CSSProperties;
+    renderItem?: ListProps<T>['renderItem'];
+    onClick?: ListProps<T>['onItemClick'];
+    onMouseMove?: (index: number) => void;
+};
+
+const defaultRenderItem = <T extends unknown>(item: T) => String(item);
```
amje, line 22: Does `extends unknown` do anything?

src/components/List/components/ListItem.tsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react';
+import {block} from '../../utils/cn';
+import {DragHandleIcon} from './DragHandleIcon';
+import type {ListProps, ListItemData, ListSortHandleAlign} from '../types';
+
+const b = block('list');
+
+type ListItemProps<T> = {
```
amje, line 8: Maybe add default type like in other interfaces?

src/components/List/components/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './ListItem';
+export * from './SimpleContainer';
```
amje, line 2: Do we need SimpleContainer as exported component? It is not documented anywhere

amje (changes requested):

src/components/List/List.tsx
```diff
@@ -1,168 +1,80 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import AutoSizer from 'react-virtualized-auto-sizer';
 import {VariableSizeList as ListContainer} from 'react-window';
+// react-sortable-hoc actually has exported members "sortableElement" and "sortableContainer"
+// https://github.com/clauderic/react-sortable-hoc/blob/master/src/index.js#L5
+// @ts-ignore
```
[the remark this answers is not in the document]
korvin89: I will add types, but they are still wrong

src/components/List/List.tsx
korvin89: We will have an error `Static members cannot reference class type parameters.` if change `unknown` to `T`

src/components/List/List.tsx
korvin89: Because `setTimeout` has this type:
```typescript
declare function setTimeout(callback: (...args: any[]) => void, ms?: number, ...args: any[]): NodeJS.Timeout;
```

korvin89: I think that it is an error, but typescript have another opinion

src/components/List/List.tsx
korvin89: Page size can be `undefined` in case of initial state. It is bad idea to change logic in case of moving code from js to ts

src/components/List/List.tsx
korvin89: Ok. But someone is already using `onKeyDown` public method according to this [type](https://github.com/yandex-cloud/uikit/blob/main/src/components/List/List.d.ts#L65). I'll leave it as it is.

src/components/List/List.tsx
korvin89: Ok. I`ll add all of these render methods after public methods

src/components/List/components/ListItem.tsx
korvin89: Yes. It`s avoiding [this](https://stackoverflow.com/a/45576880) trouble

src/components/List/components/ListItem.tsx
korvin89: Why? We don`t export this type. Type `T` come from `ListItem`

src/components/List/components/index.ts
korvin89: Actually we don`t need

src/components/List/List.tsx
amje: Where this type comes from? Which file?

src/components/List/List.tsx
amje: Leave a comment to change it in the next major please

src/components/List/components/ListItem.tsx
amje: ok

src/components/List/List.tsx
amje: and render method should remain the last of public methods

src/components/List/List.tsx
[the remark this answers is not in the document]
amje: `@types/react-sortable-hoc` is deprecated, let's remove it. `SortableContainer ` is an alias for `sortableContainer`, same works for `SortableElement`. We can simply use them.

amje (approved):

PR #40, opened 2022-01-12
perf: small Table performance fixes

krozhkov:
* remove linear complexity from getRowId function in a compatible way
* add enhanced columns memoization to table with settings hoc
* cache css styles for cells.

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -18,8 +18,8 @@ interface TableState {
     // activeScrollElement нужен, чтобы listner на скролл таблицы не срабатывал, когда скроллится скролл-бар и наоборот
     // в противном случае таблица будет дергаться при прокрутке
     activeScrollElement: ActiveScrollElementType;
-    columnsWidth: (number | undefined)[];
-    columnHeaderRefs: React.RefObject<HTMLTableHeaderCellElement>[];
+    columnsStyles: React.CSSProperties[];
+    columnHeaderRefs: React.RefObject<HTMLTableCellElement>[];
 }
 
 export interface TableColumnConfig<I> {
@@ -101,9 +101,9 @@ export class Table<I extends TableDataItem = Record<string, string>> extends Rea
     };
 
     // Static methods may be used by HOCs
-    static getRowId<I extends TableDataItem>(props: TableProps<I>, item: I) {
+    static getRowId<I extends TableDataItem>(props: TableProps<I>, item: I, rowIndex?: number) {
         const {data, getRowId} = props;
-        const index = data.indexOf(item);
+        const index = rowIndex ?? data.indexOf(item);
 
         if (typeof getRowId === 'function') {
             return getRowId(item, index);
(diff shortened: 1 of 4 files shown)


src/components/Table/Table.tsx
```diff
@@ -101,9 +101,8 @@ export class Table<I extends TableDataItem = Record<string, string>> extends Rea
     };
 
     // Static methods may be used by HOCs
-    static getRowId<I extends TableDataItem>(props: TableProps<I>, item: I) {
-        const {data, getRowId} = props;
-        const index = data.indexOf(item);
+    static getRowId<I extends TableDataItem>(props: TableProps<I>, item: I, index: number) {
```
amje, line 104: This is a breaking change, not sure we are ready to next major now

src/components/Table/Table.tsx
```diff
@@ -18,8 +18,8 @@ interface TableState {
     // activeScrollElement нужен, чтобы listner на скролл таблицы не срабатывал, когда скроллится скролл-бар и наоборот
     // в противном случае таблица будет дергаться при прокрутке
     activeScrollElement: ActiveScrollElementType;
-    columnsWidth: (number | undefined)[];
-    columnHeaderRefs: React.RefObject<HTMLTableHeaderCellElement>[];
+    columnsStyles: React.CSSProperties[];
+    columnHeaderRefs: React.RefObject<HTMLTableCellElement>[];
```
amje, line 22: Why narrowing down the type?

src/components/Table/Table.tsx
```diff
@@ -428,17 +428,21 @@ export class Table<I extends TableDataItem = Record<string, string>> extends Rea
...
+    private updateColumnStyles() {
+        const columnsWidth = this.state.columnHeaderRefs.map((ref) =>
+            ref.current === null ? undefined : ref.current.getBoundingClientRect().width,
+        );
+
+        const columnsStyles = this.props.columns.map((_, index) =>
+            this.getColumnStyles(index, columnsWidth),
+        );
+
+        this.setState({columnsStyles});
```
amje, line 440: Better use function version of setState (`setState((prevState) => newState)`) if you need previous state to build a new state

src/components/Table/Table.tsx
krozhkov: HTMLTableHeaderCellElement is deprecated in typescript 4.5.4 definitions.

src/components/Table/Table.tsx
krozhkov: I can add an index as optional parameter and if it is not present fall back to the previous logic.

src/components/Table/Table.tsx
```diff
@@ -101,9 +101,9 @@ export class Table<I extends TableDataItem = Record<string, string>> extends Rea
     };
 
     // Static methods may be used by HOCs
-    static getRowId<I extends TableDataItem>(props: TableProps<I>, item: I) {
-        const {data, getRowId} = props;
-        const index = data.indexOf(item);
+    static getRowId<I extends TableDataItem>(props: TableProps<I>, item: I, rowIndex?: number) {
+        const {getRowId} = props;
+        const index = rowIndex ?? props.data.indexOf(item);
```
amje, line 106: let's keep vars as they was: `const {data, getRowId} = props;`

src/components/Table/Table.tsx
amje: I think so, in the next major we might want to make the parameter required

amje (approved):

PR #53, opened 2022-01-20
add Control for Select component

korvin89:
Add `Control` component into `Select`. Also add temporary showcase. There are no logic and any handlers yet.


src/components/Select/components/Control/Control.tsx
```diff
@@ -0,0 +1,45 @@
...
+        ...(typeof width === 'string' && {width}),
+    };
+    const controlInlineStyles: React.CSSProperties = {};
+
+    if (typeof width === 'number') {
+        controlInlineStyles.width = width;
+    }
+
+    return (
+        <div className={b(controlMods)} style={controlInlineStyles} tabIndex={1}>
```
ValeraS, line 38: Are we planning to make it accessibility friendly?
Why use `div` instead of `button`?
What about `role="combobox"`, `aria-expanded="..."`, `aria-haspopup="listbox"`, e.t.c. ?

references:
-  https://w3c.github.io/aria/#combobox
- https://w3c.github.io/aria-practices/examples/combobox/combobox-select-only.html

korvin89: Change wrapper to `button`, add `role` and `aria-haspopup`

src/components/Select/components/Control/Control.scss
```diff
@@ -0,0 +1,273 @@
...
+    @include button-text-color(var(--yc-color-text-primary));
+
+    box-sizing: border-box;
+    display: inline-flex;
+    align-items: center;
+    min-width: 34px;
+    border: 1px solid var(--yc-color-line-generic);
+    background-color: transparent;
+    white-space: nowrap;
+    transition: transform 0.1s ease-out, border-color 0.15s linear, background-color 0.15s linear,
```
amje, line 29: Should be exact same behaviour as in TextInput, where no transition at all

src/components/utils/cn.ts
```diff
@@ -2,6 +2,8 @@ import bemBlock from 'bem-cn-lite';
 
 export type CnBlock = ReturnType<typeof bemBlock>;
 
+export type CnMods = Record<string, string | boolean>;
```
amje, line 5: value can be `undefined` as well

src/components/Select/types.ts
```diff
@@ -1,6 +1,12 @@
-import {TextInputSize, TextInputView} from '../TextInput';
+import React from 'react';
+import {TextInputView, TextInputSize, TextInputPin} from '../TextInput';
 
 export type SelectProps = {
     view?: TextInputView;
     size?: TextInputSize;
+    pin?: TextInputPin;
+    controlWidth?: 'unset' | 'auto' | 'max' | number;
```
amje, line 8: Is there `disabled` prop? Can we implement it now?

amje, line 8: `'unset'` is not the same as `undefined` here?

src/components/Select/components/Control/Control.scss
```diff
@@ -0,0 +1,273 @@
...
+
+    &_view_clear {
+        border-color: transparent;
+    }
+
+    &_size_s {
+        height: 24px;
+        padding: 0 8px;
+
+        &#{$block}_pin_round-round {
```
amje, line 41: Looks like candidate for a mixin. Can be used in TextInput component as well.

src/components/Select/components/Control/Control.scss
```diff
@@ -0,0 +1,273 @@
...
+
+    &:hover {
+        color: $hoverColor;
+    }
+}
+
+$block: '.#{$ns}select-control';
+
+#{$block} {
+    @include button-text-color(var(--yc-color-text-primary));
```
amje, line 20: We should apply `button-reset` mixin here

amje (changes requested):

src/components/Select/types.ts
korvin89: It is not the same because we want have `controlWidth : 'auto'` by default. This means that `undefined` value automatically converted to `'auto'`

PR #58, opened 2022-01-21
chore: add checks on precommit hook

ValeraS:


package.json
```diff
@@ -19,14 +19,15 @@
...
+        "init-git-hooks": "npx simple-git-hooks",
         "test": "exit 0",
         "clean": "gulp clean",
         "build": "gulp",
         "start": "start-storybook -p 7007",
         "build-storybook": "build-storybook -c .storybook -o storybook-static",
         "lint:js": "eslint --ext .js,.jsx,.ts,.tsx .",
         "lint:styles": "stylelint '{styles,src}/**/*.scss'",
-        "lint:prettier": "prettier --check --loglevel=error '**/*.{js,jsx,ts,tsx,scss}'",
+        "lint:prettier": "prettier --check --loglevel=error '**/*.md'",
```
ValeraS, line 30: `eslint` and `stylelint` is configured to fix prettier errors

package.json
```diff
@@ -68,6 +69,7 @@
         "gulp-dart-sass": "^1.0.2",
         "gulp-replace": "^1.1.3",
         "gulp-typescript": "^5.0.1",
+        "nano-staged": "^0.5.0",
```
amje, line 72: Why not `lint-staged` that we already familiar with?

package.json
```diff
@@ -87,5 +90,19 @@
...
+            "stylelint --fix --quiet"
+        ],
+        "*.{js,jsx,ts,tsx}": [
+            "eslint --fix --quiet"
+        ],
+        "*.md": [
+            "prettier --write"
+        ]
+    },
+    "simple-git-hooks": {
```
amje, line 105: Let's consider husky as a script runner?

package.json
ValeraS: `nano-staged` has identical functionality and config, but is much smaller, has only one small dependency and faster. https://github.com/usmanyunusov/nano-staged#benchmarks

package.json
```diff
@@ -19,17 +19,20 @@
         "styles"
     ],
     "scripts": {
+        "prepare": "husky install",
```
ValeraS, line 22: husky doesn't install hooks automatically, husky documentation recommends use `prepare` script, but I'm not sure. `prepare` is executed during install, pack and publish.

amje: I don't think it is a problem

package.json
```diff
@@ -19,17 +19,20 @@
...
         "lint:js": "eslint --ext .js,.jsx,.ts,.tsx .",
         "lint:styles": "stylelint '{styles,src}/**/*.scss'",
-        "lint:prettier": "prettier --check --loglevel=error '**/*.{js,jsx,ts,tsx,scss}'",
+        "lint:prettier": "prettier --check --loglevel=error '**/*.md'",
         "lint": "run-p lint:*",
         "typecheck": "tsc --noEmit",
-        "prepublishOnly": "npm run build"
+        "prepublishOnly": "npm run build",
+        "pre-commit": "npx nano-staged",
+        "commit-msg": "npx commitlint --edit $1"
```
amje, line 35: Let's use commands directly from husky scripts. There is no practical benefit of listing them in npm scripts

ValeraS: fixed

amje (approved):

PR #65, opened 2022-01-25
feat: border-radius tokens

amje:


src/components/Button/Button.tsx
```diff
@@ -74,7 +74,7 @@ const PureButton = React.forwardRef<HTMLButtonElement | HTMLAnchorElement, Butto
         {
             view = 'normal',
             size = 'm',
-            pin,
+            pin = 'round-round',
```
amje, line 77: Shouldn't be this change considered as breaking?

korvin89: Seems ok

korvin89 (approved):

PR #66, opened 2022-01-25
fix: use activeItem in List

dividead:
- allow to select 0th element in list
- pre-activate list item from props
- allow Tab keypress to select item

### DIFF
--- a/src/components/List/List.tsx
+++ b/src/components/List/List.tsx
@@ -68,6 +68,7 @@ export class List<T = unknown> extends React.Component<ListProps<T>, ListState<T
     state: ListState<T> = {
         items: this.props.items,
         filter: '',
+        activeItem: this.props.activeItemIndex,
     };
 
     refFilter = React.createRef<HTMLInputElement | HTMLTextAreaElement>();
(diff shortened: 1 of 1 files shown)


src/components/List/List.tsx
```diff
@@ -164,8 +165,9 @@ export class List<T = unknown> extends React.Component<ListProps<T>, ListState<T
                 this.handleKeyMove(event, -(activeItem || 0) - 1);
                 break;
             }
+            case 'Tab':
```
ValeraS, line 168: Why does `Tab` have to select an element? The user wanted to focus on the next control and pressed Tab, but as a side effect, select the list item. IMHO, this behaviour is unexpected.

src/components/List/List.tsx
```diff
@@ -183,7 +185,7 @@ export class List<T = unknown> extends React.Component<ListProps<T>, ListState<T
         const {items, activeItem} = this.state;
         const item = items[index];
         const sortable = this.props.sortable && items.length > 1 && !this.getFilter();
-        const active = index === activeItem || index === this.props.activeItemIndex;
+        const active = index === activeItem;
```
ValeraS, line 188: `activeItemIndex` prop allows you to fix the active element, it is needed, for example, if there is a dropdown menu with actions for each element and it is not visible for inactive elements. When the dropdown is opened, the active element should be fixed.

dividead: why would something with "active" in name freeze an element?
and it doesn't even relates to activeItem, which can still be moved

PR #68, opened 2022-01-26
add Listbox for Select component

korvin89:
Listbox for Select component

NB:
- there are no any optimizations like `React.useMemo` and etc _yet_
- there are no `uncontrolled` mode handling _yet_

### DIFF
--- a/src/components/List/List.tsx
+++ b/src/components/List/List.tsx
@@ -203,7 +203,7 @@ export class List<T = unknown> extends React.Component<ListProps<T>, ListState<T
                 itemClassName={this.props.itemClassName}
                 active={active}
                 selected={index === this.props.selectedItemIndex}
-                onMouseMove={this.onItemMouseMove}
+                onMouseEnter={this.onItemMouseEnter}
                 onClick={this.props.onItemClick}
             />
         );
@@ -401,7 +401,7 @@ export class List<T = unknown> extends React.Component<ListProps<T>, ListState<T
         });
     };
 
-    private onItemMouseMove = (index: number) => {
+    private onItemMouseEnter = (index: number) => {
         if (!this.state.sorting) {
             this.activateItem(index, false);
         }
@@ -431,7 +431,7 @@ export class List<T = unknown> extends React.Component<ListProps<T>, ListState<T
 
         if (typeof itemHeight === 'function') {
             const {items} = this.state;
-            return itemHeight(items[index]);
+            return itemHeight(items[index], index);
         }
         return itemHeight;
     };
--- a/src/components/List/List.tsx
+++ b/src/components/List/List.tsx
@@ -401,7 +401,7 @@ export class List<T = unknown> extends React.Component<ListProps<T>, ListState<T
         });
     };
 
-    private onItemMouseEnter = (index: number) => {
+    private onItemMouseEnter = (index?: number) => {
         if (!this.state.sorting) {
             this.activateItem(index, false);
         }
--- a/src/components/List/types.ts
+++ b/src/components/List/types.ts
@@ -18,7 +18,7 @@ export type ListProps<T = unknown> = {
     filter?: string;
     activeItemIndex?: number;
     selectedItemIndex?: number;
-    itemHeight?: number | ((item: ListItemData<T>) => number);
+    itemHeight?: number | ((item: ListItemData<T>, itemIndex: number) => number);
     itemsHeight?: number | ((items: T[]) => number);
     virtualized?: boolean;
     filterable?: boolean;
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -1,29 +1,62 @@
 import React from 'react';
+import {List} from '../List';
+import {reducer, initialState} from './store';
 import {SelectProps} from './types';
-import {Control} from './components';
+import {FlattenOption} from './utils';
+import {Control, Listbox} from './components';
 
-export const Select = (props: SelectProps) => {
+export const Select = <T extends unknown>(props: SelectProps<T>) => {
     const {
+        renderOption,
+        getOptionHeight,
+        className,
         label,
         placeholder,
         disabled,
+        listboxWidth,
         view = 'normal',
         size = 'm',
         pin = 'round-round',
         controlWidth = 'auto',
+        multiple = false,
         value = [],
+        options = [], // remove default after adding children handling logic
     } = props;
+    const [{controlRect, active}, dispatch] = React.useReducer(reducer, initialState);
+    const controlRef = React.useRef<HTMLButtonElement>(null);
+    const listboxRef = React.useRef<List<FlattenOption<T>>>(null);
 
     return (
-        <Control
-            view={view}
-            size={size}
-            pin={pin}
-            width={controlWidth}
-            label={label}
-            placeholder={placeholder}
-            value={value}
-            disabled={disabled}
-        />
+        <React.Fragment>
+            <Control
+                ref={controlRef}
+                listboxRef={listboxRef}
+                className={className}
+                view={view}
+                size={size}
+                pin={pin}
+                width={controlWidth}
+                label={label}
+                placeholder={placeholder}
+                value={value}
+                active={active}
+                disabled={disabled}
+                dispatch={dispatch}
+            />
+            <Listbox
+                ref={listboxRef}
+                controlRef={controlRef}
+                size={size}
+                value={value}
+                options={options}
+                listboxWidth={listboxWidth}
+                controlRect={controlRect}
+                active={active}
+                multiple={multiple}
+                dispatch={dispatch}
+                renderOption={renderOption}
+                getOptionHeight={getOptionHeight}
+            />
+        </React.Fragment>
     );
 };
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -13,7 +13,7 @@ export const Select = <T extends unknown>(props: SelectProps<T>) => {
         label,
         placeholder,
         disabled,
-        listboxWidth,
+        popupWidth,
         view = 'normal',
         size = 'm',
         pin = 'round-round',
@@ -49,7 +49,7 @@ export const Select = <T extends unknown>(props: SelectProps<T>) => {
                 size={size}
                 value={value}
                 options={options}
-                listboxWidth={listboxWidth}
+                popupWidth={popupWidth}
                 controlRect={controlRect}
                 active={active}
                 multiple={multiple}
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -5,7 +5,7 @@ import {SelectProps} from './types';
 import {FlattenOption} from './utils';
 import {Control, Listbox} from './components';
 
-export const Select = <T extends unknown>(props: SelectProps<T>) => {
+export const Select = (props: SelectProps) => {
     const {
         renderOption,
         getOptionHeight,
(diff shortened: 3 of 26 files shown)


src/components/List/List.tsx
```diff
@@ -136,7 +136,7 @@ export class List<T = unknown> extends React.Component<ListProps<T>, ListState<T
     }
 
     // FIXME: BREAKING CHANGE. Rename to "handleKeyDown"
-    onKeyDown: React.KeyboardEventHandler<HTMLDivElement> = (event) => {
+    onKeyDown: React.KeyboardEventHandler<HTMLElement> = (event) => {
```
korvin89, line 139: There is no need in such precision type

src/components/List/List.tsx
```diff
@@ -401,7 +401,7 @@ export class List<T = unknown> extends React.Component<ListProps<T>, ListState<T
         });
     };
 
-    private onItemMouseMove = (index: number) => {
+    private onItemMouseEnter = (index: number) => {
```
korvin89, line 404: Change handler to `mouseenter` to prevent useless rerenderings

src/components/List/types.ts
```diff
@@ -18,7 +18,7 @@ export type ListProps<T = unknown> = {
     filter?: string;
     activeItemIndex?: number;
     selectedItemIndex?: number;
-    itemHeight?: number | ((item: ListItemData<T>) => number);
+    itemHeight?: number | ((item: ListItemData<T>, itemIndex: number) => number);
```
korvin89, line 21: Add extra argument for convenience

src/components/Select/components/Control/Control.tsx
```diff
@@ -36,18 +58,54 @@ export const Control = (props: ControlProps) => {
...
 
+    const getControlNode = React.useCallback(() => {
+        return (ref as React.RefObject<HTMLButtonElement>)?.current;
+    }, [ref]);
+
+    const handleClick = () => {
+        dispatch({type: 'SET_ACTIVE', payload: {active: !active}});
+    };
+
+    const handleKeyDown = (e: React.KeyboardEvent<HTMLButtonElement>) => {
```
korvin89, line 69: This handler helps to handle keyboard actions in case of control node focused

src/components/Select/components/Listbox/Listbox.scss
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+
+    &__group-label {
+        box-sizing: border-box;
+        position: relative;
+        width: 100%;
+        font-weight: 500;
+        white-space: nowrap;
+        text-overflow: ellipsis;
```
korvin89, line 39: We cannot use `@include overflow-ellipsis()` cause of we do not need style `overflow: hidden`

src/components/Select/components/Listbox/Listbox.tsx
```diff
@@ -0,0 +1,122 @@
...
+        value,
+        listboxWidth,
+        controlRect,
+        active,
+        multiple,
+        controlRef,
+    } = props;
+    const flattenOptions = getFlattenOptions(options);
+    const listboxHeight = getListboxHeight({options: flattenOptions, getOptionHeight, size});
+    const popupVerticalOffset = getPopupVerticalOffset({listboxHeight, controlRect});
```
korvin89, line 57: It is experimental feature :)

src/components/Select/Select.tsx
```diff
@@ -1,29 +1,62 @@
...
-export const Select = (props: SelectProps) => {
+export const Select = <T extends unknown>(props: SelectProps<T>) => {
     const {
+        renderOption,
+        getOptionHeight,
+        className,
         label,
         placeholder,
         disabled,
+        listboxWidth,
```
amje, line 16: `popupWidth`?

src/components/Select/components/Control/Control.tsx
```diff
@@ -1,33 +1,55 @@
...
+import {Action} from '../../store';
 import {SelectProps} from '../../types';
+import {FlattenOption} from '../../utils';
 
 import './Control.scss';
 
 const b = block('select-control');
 
 type ControlProps = {
+    dispatch: React.Dispatch<Action>;
```
amje, line 16: Replace with `Dispatch` from `../../store`?

src/components/Select/components/Control/Control.tsx
```diff
@@ -36,18 +58,54 @@ export const Control = (props: ControlProps) => {
...
+    const handleClick = () => {
+        dispatch({type: 'SET_ACTIVE', payload: {active: !active}});
+    };
+
+    const handleKeyDown = (e: React.KeyboardEvent<HTMLButtonElement>) => {
+        listboxRef?.current?.onKeyDown(e);
+    };
+
+    const handleResize = React.useCallback(() => {
+        const controlRect = getControlNode()?.getBoundingClientRect();
```
amje, line 74: `ref` can be used directly here without accessor function, `ref` is always the same object

src/components/Select/components/Control/Control.tsx
```diff
@@ -36,18 +58,54 @@ export const Control = (props: ControlProps) => {
...
+        dispatch({type: 'SET_CONTROL_RECT', payload: {controlRect}});
+    }, [dispatch, getControlNode]);
+
+    React.useLayoutEffect(() => {
+        const resizeObserver = new ResizeObserver(handleResize);
+        const element = getControlNode();
```
amje, line 80: Same here, can be shorten to `const element = ref.current`;

src/components/Select/components/Control/Control.tsx
```diff
@@ -36,18 +58,54 @@ export const Control = (props: ControlProps) => {
...
+        };
+    }, [getControlNode, handleResize]);
+
     return (
         <button
-            className={b(controlMods)}
+            ref={ref}
+            className={b(controlMods, className)}
             style={controlInlineStyles}
-            role="combobox"
```
amje, line 43: Why deleted the prop?

src/components/Select/components/Listbox/GroupLabel.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+const b = block('select-listbox');
+
+type GroupLabelProps = {
+    size: NonNullable<SelectProps['size']>;
+    label: string;
+};
+
+export const GroupLabel = ({size, label}: GroupLabelProps) => {
+    return <div className={b('group-label', {size})}>{label}</div>;
```
amje, line 13: Is it worth extracting a "oneliner"?

src/components/Select/components/Listbox/Listbox.scss
```diff
@@ -0,0 +1,154 @@
+@import '../../../../../styles/mixins';
+@import '../../../variables';
+
+$block: '.#{$ns}select-listbox';
+$popupBlock: '.#{$ns}popup';
+$s-height: '28px';
+$s-hor-padding: '8px';
```
amje, line 7: Just `$s-padding` maybe? There is no other paddings.

src/components/Select/components/Listbox/Listbox.scss
amje: It works without it?

src/components/Select/components/Listbox/Listbox.tsx
```diff
@@ -0,0 +1,122 @@
...
+const b = block('select-listbox');
+const VIRTUALIZE_THRESHOLD = 50;
+
+type ListForwardedRef<T> =
+    | ((instance: List<FlattenOption<T>> | null) => void)
+    | React.MutableRefObject<List<FlattenOption<T>> | null>
+    | null;
+
+type ListboxProps<T = unknown> = {
+    dispatch: React.Dispatch<Action>;
```
amje, line 28: `Dispatch` alias

src/components/Select/utils/listbox.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    if (!controlRect) {
+        return 0;
+    }
+
+    const vh = window.innerHeight / 100;
+    const heigth5vh = vh * 5;
+    const heigth90vh = vh * 90;
+    const containerHeight = heigth90vh < listboxHeight ? heigth90vh : listboxHeight;
+    const popupPlacement: PopupPlacement =
```
amje, line 67: Popup should do it by default. It has `flip` option enabled.

src/components/Select/components/Listbox/Listbox.scss
korvin89: I added two styles instead
```css
white-space: nowrap;
text-overflow: ellipsis;
```

korvin89: Hmm.. no. I`ll fix that

src/components/Select/components/Control/Control.tsx
korvin89: Yes, the same. It\`s something like shortcut to deal with type cast `ref as React.RefObject<HTMLButtonElement>` case of typescript waits here `ref` from `React.forwardRef`

src/components/Select/components/Control/Control.tsx
korvin89: @ValeraS said that we didn`t need that props :) but I couldn`t find it in documentation

src/components/Select/components/Listbox/GroupLabel.tsx
korvin89: I fixed text collapse. Now it looks like
```typescript
export const GroupLabel = ({label}: GroupLabelProps) => {
    return (
        <div className={b('group-label')}>
            <div className={b('group-label-content')}>{label}</div>
        </div>
    );
};
```

src/components/Select/components/Listbox/Listbox.scss
korvin89: It using like this:
```scss
#{$block}_size_s & {
    height: #{$s-height};
    padding: 8px #{$s-hor-padding} 4px #{$s-hor-padding};
}
```
I think that variable must show that padding exactly horizontal

src/components/Select/utils/listbox.ts
korvin89: This type using there only for two typed constants. It is pure function that knows nothing about popup

src/components/Select/components/Control/Control.tsx
```diff
@@ -24,10 +29,10 @@ type ControlProps = {
     placeholder?: SelectProps['placeholder'];
     active?: boolean;
     disabled?: boolean;
-    listboxRef?: React.RefObject<List<FlattenOption>>;
+    listboxRef?: React.RefObject<List<FlattenOption<T>>>;
 };
 
-export const Control = React.forwardRef<HTMLButtonElement, ControlProps>((props, ref) => {
+const ControlInner = <T extends unknown>(props: ControlProps<T>, ref: ButtonForwardedRef) => {
```
amje, line 35: I wonder why we need the generic here in the first place? It comes from SelectOptions that inherits HTMLSelectOption. Do we actually need that inheritance?

src/components/Select/utils/constants.ts
```diff
@@ -0,0 +1,14 @@
+import {SelectProps} from '../types';
```
amje, line 1: `Select/constants.ts`? No reason to place it into utils

src/components/Select/utils/index.ts
```diff
@@ -1,2 +1,3 @@
+export * from './constants';
 export * from './listbox';
 export * from './types';
```
amje, line 3: Maybe put all of these into one file `Select/utils.ts`?

PR #86, opened 2022-02-09
feat(ClipboardButton): improve a11y

ogonkov:
Add focus and keyboard handlers to `<ClipboardButton/>`

### DIFF
--- a/src/components/ClipboardButton/FakeButton.tsx
+++ b/src/components/ClipboardButton/FakeButton.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
+import React, {MouseEventHandler} from 'react';
 import {block} from '../utils/cn';
+import {useButtonHandlers} from '../utils/useButtonHandlers';
 import {ClipboardIcon} from '../ClipboardIcon';
 import {CopyToClipboardStatus} from '../CopyToClipboard';
 import {ClipboardButtonProps} from './ClipboardButton';
@@ -12,6 +13,7 @@ const DEFAULT_ICON_SIZE = 24;
 interface FakeButtonProps extends Pick<ClipboardButtonProps, 'className' | 'size'> {
     status: CopyToClipboardStatus;
     'data-qa': ClipboardButtonProps['qa'];
+    onClick?: MouseEventHandler<HTMLSpanElement>;
 }
 
 export function FakeButton({
@@ -19,9 +21,19 @@ export function FakeButton({
     className,
     'data-qa': qa,
     size = DEFAULT_ICON_SIZE,
+    onClick,
 }: FakeButtonProps) {
+    const {onKeyDown} = useButtonHandlers<HTMLSpanElement>(onClick);
+
     return (
-        <span className={b(null, className)} data-qa={qa}>
+        <span
+            tabIndex={0}
+            role="button"
+            className={b(null, className)}
+            data-qa={qa}
+            onClick={onClick}
+            onKeyDown={onKeyDown}
+        >
             <ClipboardIcon status={status} size={size} className={b('icon')} />
         </span>
     );
--- a/src/components/ClipboardButton/FakeButton.tsx
+++ b/src/components/ClipboardButton/FakeButton.tsx
@@ -1,6 +1,6 @@
 import React, {MouseEventHandler} from 'react';
 import {block} from '../utils/cn';
-import {useButtonHandlers} from '../utils/useButtonHandlers';
+import {useActionHandlers} from '../utils/useActionHandlers';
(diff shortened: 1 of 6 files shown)


src/components/utils/useButtonHandlers.ts
```diff
@@ -0,0 +1,18 @@
+import {KeyboardEvent, KeyboardEventHandler, useCallback} from 'react';
+
+interface UseButtonHandlers<T> {
+    onKeyDown: KeyboardEventHandler<T>;
+}
+
+export function useButtonHandlers<T>(onClick?: (...args: any[]) => any): UseButtonHandlers<T> {
```
amje, line 7: There is nothing about button, maybe better `useActionKeyDown`? Also arg should be called `callback` of smth

src/components/ClipboardButton/FakeButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const DEFAULT_ICON_SIZE = 24;
+
+interface FakeButtonProps extends Pick<ClipboardButtonProps, 'className' | 'size'> {
+    status: CopyToClipboardStatus;
+    'data-qa': ClipboardButtonProps['qa'];
+    onClick?: MouseEventHandler<HTMLSpanElement>;
+}
+
+export function FakeButton({
```
amje, line 19: What is the point of extracting it into separate component if we proxy all props? The original component wasn't so big

amje (changes requested):

src/components/ClipboardButton/FakeButton.tsx
ogonkov: I need to call hook inside, that adds keyboard handlers. onClick here is passed via cloning node, that returned from clipboard children callback

src/components/ClipboardButton/FakeButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+import {ClipboardButtonProps} from './ClipboardButton';
+import './ClipboardButton.scss';
+
+const b = block('clipboard-button');
+
+const DEFAULT_ICON_SIZE = 24;
+
+interface FakeButtonProps extends Pick<ClipboardButtonProps, 'className' | 'size'> {
+    status: CopyToClipboardStatus;
+    'data-qa': ClipboardButtonProps['qa'];
```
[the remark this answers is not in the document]
ogonkov: May be. I personally prefer to have prop with same name as dom attribute, but i can change it, no problem

src/components/ClipboardButton/FakeButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+    status,
+    className,
+    'data-qa': qa,
+    size = DEFAULT_ICON_SIZE,
+    onClick,
+}: FakeButtonProps) {
+    const {onKeyDown} = useButtonHandlers<HTMLSpanElement>(onClick);
+
+    return (
+        <span
```
ValeraS, line 29: Why not use the `button` tag instead of emulating it?

src/components/utils/useActionHandlers.ts
```diff
@@ -0,0 +1,22 @@
+import {KeyboardEvent, KeyboardEventHandler, useCallback} from 'react';
+
+interface UseButtonHandlers<T> {
```
amje, line 3: `UseActionHandlersResult<T>`?

src/components/ClipboardButton/FakeButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+    status: CopyToClipboardStatus;
+    onClick?: MouseEventHandler<HTMLSpanElement>;
+}
+
+export function FakeButton({
+    status,
+    className,
+    qa,
+    size = DEFAULT_ICON_SIZE,
+    onClick,
```
amje, line 23: Where are we passing onClick prop? I cannot see.

amje (changes requested):

src/components/ClipboardButton/FakeButton.tsx
ogonkov: It is passed by `react-copy-to-clipboard` via cloning element

ogonkov: It is passed by `react-copy-to-clipboard` via cloning element

ogonkov: Here

https://github.com/nkbt/react-copy-to-clipboard/blob/fa8b7e2b1cfdcda5908b2d30f91f02c6dd008aba/src/Component.js#L58

amje (approved):

PR #93, opened 2022-02-16
chore: remove SVGs from a11y tree

ogonkov:


src/components/utils/svgA11yHidden.ts
```diff
@@ -0,0 +1,6 @@
+export const svgA11yHidden = {
+    // Hides element from a11y tree
+    'aria-hidden': true,
+    // Prevents focus in IE
+    focusable: false,
```
amje, line 5: We don't support IE, so there is no need to add this.

src/components/utils/svgA11yHidden.ts
```diff
@@ -0,0 +1,6 @@
+export const svgA11yHidden = {
```
amje, line 1: Not sure about standalone file and object itself cause we have only one prop needed (`aria-hidden`)

src/components/utils/svgA11yHidden.ts
ogonkov: I guess it is not a big deal, we improve UX for IE users, they will receive redundant focus instead

amje: Redundant focus will be the least problem. JS simply won't work in IE due to modern features we are using in the code. No support for IE - no legacy code

src/components/utils/svgA11yHidden.ts
ogonkov: Code itself is clear about it purpose, that could be not clear why svgs receive `aria-hidden`

amje: Ok, but let's name it accordingly, `a11yHiddenSvgProps`, more accent on props. And also let's rename file to `utils/svg.ts` for future svg code

ogonkov: Done

amje (approved):

PR #94, opened 2022-02-19
feat(Tabs): improve a11y

ogonkov:


tsufiev:
@ogonkov Are you certain that this change qualify as new feature - i.e., will create in new entry in 'Features' section and cause minor version of library to be bumped? Looks more like a refactor or a fix (in that case it's better to write down what was broken before) to me.

ogonkov:
For keyboard users it is definitely a feature. But i guess we can qualify it as fix, if you feel it is more preferable.

src/components/Tabs/TabsItem/TabsItem.tsx
```diff
@@ -43,7 +43,15 @@ export const TabsItem: React.FC<TabsItemProps> = ({
     }, [hint, title]);
 
     return (
-        <div className={b({active, disabled, direction})} title={htmlTitle} onClick={handleClick}>
+        <div
+            role="tab"
+            aria-selected={active === true}
+            aria-disabled={disabled === true}
```
mfyodorov, line 49: These properties are exactly what I told you about recently when we were talking about improvements of a11y in common components

PR #104, opened 2022-03-21
Add content border-radius variable [DATAUI-1130]

yuberdysheva:


src/components/Modal/Modal.scss
```diff
@@ -99,3 +99,7 @@ $block: '.#{$ns}modal';
         transform: scale(1);
     }
 }
+
+:root {
+    --yc-modal-content-border-radius: 5px;
```
amje, line 104: Let's name it without `-content-`

amje (changes requested):

amje (approved):

PR #119, opened 2022-03-31
feat(Tooltip): new tooltip component

Ayub-Begimkulov:


src/components/Tooltip2/Tooltip.tsx
```diff
@@ -0,0 +1,84 @@
...
+import isFunction from 'lodash/isFunction';
+import isObject from 'lodash/isObject';
+
+import {Popup, PopupPlacement} from '../Popup';
+import {block} from '../utils/cn';
+
+import './Tooltip.scss';
+
+export interface TooltipProps {
+    text: React.ReactNode;
```
amje, line 11: Let's name it `content`. We usually use this name.

amje (changes requested):

src/components/Tooltip2/Tooltip.tsx
Ayub-Begimkulov: Update the name in the latest commit

amje (approved):

PR #120, opened 2022-04-02
feat: allow use layers for Dialog/Modal/Popup with different root

krozhkov:
allow display modals with a different root element. this may be useful in case when we need to determine the source of modals or place modals into container other than the document root. the default behaviour remains the same.

Example: (define the new modals root)
```
const ref = useRef();

<LayerProvider container={ref}>
  ...components where we need new root for modals
</LayerProvider>
<div ref={ref} className="modal-root"></div>
```

### DIFF
--- a/src/components/Portal/Portal.tsx
+++ b/src/components/Portal/Portal.tsx
@@ -1,11 +1,13 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
+import {usePortalContainer} from '../utils/usePortalContainer';
 
 export interface PortalProps {
     container?: HTMLElement;
     children?: React.ReactNode;
 }
 
-export function Portal({container = document.body, children}: PortalProps) {
-    return ReactDOM.createPortal(children, container);
+export function Portal({container, children}: PortalProps) {
+    const defaultContainer = usePortalContainer();
+    return ReactDOM.createPortal(children, container ?? defaultContainer);
 }
--- a/src/components/utils/PortalProvider.tsx
+++ b/src/components/utils/PortalProvider.tsx
@@ -1,0 +1,14 @@
+import React from 'react';
+import type {RefObject} from 'react';
+
+export const PortalContext = React.createContext<RefObject<HTMLElement>>({current: null});
+
+PortalContext.displayName = 'PortalContext';
+
+export type PortalProviderProps = {
+    container: RefObject<HTMLElement>;
+};
+
+export const PortalProvider: React.FC<PortalProviderProps> = ({container, children}) => {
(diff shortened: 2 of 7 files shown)


src/components/utils/LayerProvider.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+const LayerContext = React.createContext(new LayerManager());
+
+export type LayerProviderProps = {
+    container: React.RefObject<HTMLElement>;
+};
+
+export const LayerProvider: React.FC<LayerProviderProps> = ({container, children}) => {
+    const context = React.useMemo(() => {
+        return new LayerManager(container);
```
ogonkov, line 12: each invocation will (in theory) create new instance of LayerManager

it is better to explicitly add some kind of `destroy` method and invoke it on ref change

src/components/utils/LayerProvider.tsx
```diff
@@ -0,0 +1,20 @@
...
+    }, [container]);
```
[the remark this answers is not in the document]
krozhkov: I don't want to recreate the context often so a mutable RefObject is a best choice.

src/components/utils/LayerProvider.tsx
krozhkov: creating another LayerManager instance in same LayerProvider is very rare. so i don't want complicate code for this.

ogonkov: That mean you will pass through memory leak, since `window` listeners is not cleared

src/components/Portal/Portal.tsx
```diff
@@ -1,11 +1,13 @@
...
 
 export interface PortalProps {
     container?: HTMLElement;
     children?: React.ReactNode;
 }
 
-export function Portal({container = document.body, children}: PortalProps) {
-    return ReactDOM.createPortal(children, container);
+export function Portal({container, children}: PortalProps) {
+    const defaultContainer = usePortalContainer();
```
amje, line 11: I think hook is not needed for now. It is useful only in Portal.

src/components/utils/PortalProvider.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import type {RefObject} from 'react';
+
+export const PortalContext = React.createContext<RefObject<HTMLElement>>({current: null});
```
amje, line 4: Let context value be just HTMLElement, with `document.body` default. And I would name it `PortalContainerContext`

amje (changes requested):

amje (approved):

PR #122, opened 2022-04-06
refactor(ClipboardButton): use `<Button/>` in `<ClipboardButton/>`

ogonkov:


src/components/Button/Button.tsx
```diff
@@ -194,19 +189,38 @@ function prepareChildren(children: React.ReactNode) {
...
-                    rightIcon = (
-                        <span key="icon-right" className={b('icon', {side: 'right'})}>
-                            <span className={b('icon-inner')}>{item}</span>
-                        </span>
-                    );
+                    const key = 'icon-right';
+                    const side = 'end';
+                    if (isIconElement) {
+                        rightIcon = (
+                            <ButtonIcon key={key} side={side}>
```
amje, line 215: Let's use static field `Button.Icon`

src/components/Button/ButtonIcon.tsx
```diff
@@ -0,0 +1,34 @@
+import React, {PropsWithChildren} from 'react';
+import {block} from '../utils/cn';
+
+import './Button.scss';
```
amje, line 4: This will double Button styles, no need to import styles here

amje (changes requested):

amje (approved):

PR #132, opened 2022-04-18
feat(List): add eventBroker for List

Mishnya:
Added `eventBroker` handler for clicking on list items

### DIFF
--- a/src/components/List/components/ListItem.tsx
+++ b/src/components/List/components/ListItem.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import {block} from '../../utils/cn';
+import {eventBroker} from 'src/components/utils/event-broker';
 import {DragHandleIcon} from './DragHandleIcon';
 import type {ListProps, ListItemData, ListSortHandleAlign} from '../types';
 
@@ -23,6 +24,7 @@ const defaultRenderItem = <T extends unknown>(item: T) => String(item);
 
 export class ListItem<T = unknown> extends React.Component<ListItemProps<T>> {
     ref = React.createRef<HTMLDivElement>();
+    _publishEvent = eventBroker.withEventPublisher('button');
 
     render() {
         const {item, style, sortable, sortHandleAlign, itemClassName, selected, active} =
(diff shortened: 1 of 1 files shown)


src/components/List/components/ListItem.tsx
```diff
@@ -23,6 +24,7 @@ const defaultRenderItem = <T extends unknown>(item: T) => String(item);
 
 export class ListItem<T = unknown> extends React.Component<ListItemProps<T>> {
     ref = React.createRef<HTMLDivElement>();
+    _publishEvent = eventBroker.withEventPublisher('button');
```
amje, line 27: why not private?

amje, line 27: button -> list

amje (changes requested):

src/components/List/components/ListItem.tsx
```diff
@@ -69,7 +71,13 @@ export class ListItem<T = unknown> extends React.Component<ListItemProps<T>> {
         return <div className={b('item-content')}>{renderItem(item, active, itemIndex)}</div>;
     }
 
-    private onClick = () => this.props.onClick?.(this.props.item, this.props.itemIndex);
+    private onClick: React.MouseEventHandler<HTMLDivElement> = (event) => {
+        this._publishEvent({
```
korvin89, line 75: Do you need to publish click event in case of handler absence?

src/components/List/components/ListItem.tsx
korvin89: And it looks like constant

src/components/List/components/ListItem.tsx
Mishnya: Yes, i have to publish click event if the item can be clicked.

korvin89: You may have case when item does not have own click handler. Why component should trigger click event in that case?

Mishnya: Because we collecting user clicks, not which handlers have been triggered by the user.

amje (approved):

Issue #135, opened 2022-04-19
Select improvements


korvin89:
1. add optional props `onOpenChange`
2. add optional props `renderControl`

PR #136, opened 2022-04-19
feat(Select): improvements

korvin89:
#135

### DIFF
--- a/src/components/Select/components/SelectControl/SelectControl.tsx
+++ b/src/components/Select/components/SelectControl/SelectControl.tsx
@@ -44,7 +44,7 @@ export const SelectControl = React.forwardRef<HTMLElement, ControlProps>((props,
         disabled,
     } = props;
     const controlRef = React.useRef<HTMLElement>(null);
-    const handleControlRef = useForkRef(ref, controlRef);
+    const handleControlRef = useForkRef<HTMLElement>(ref, controlRef);
     const showOptionsText = Boolean(optionsText.length);
     const showPlaceholder = Boolean(placeholder && !showOptionsText);
     const mods: CnMods = {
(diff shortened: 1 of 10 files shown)


src/components/Select/components/SelectPopup/SelectPopup.scss
```diff
@@ -78,6 +78,7 @@ $xl-hor-padding: '12px';
         display: flex;
         align-items: center;
         width: 100%;
+        height: 100%;
```
korvin89, line 81: It needs to avoid problems such this
Снимок экрана 2022-03-28 в 18 59 20

src/components/Select/types.ts
```diff
@@ -5,6 +5,8 @@ import {Option, OptionGroup} from './tech-components';
 
 export type SelectProps = Pick<ControlGroupProps, 'name' | 'disabled'> & {
     onUpdate?: (value: string[]) => void;
+    onOpenChange?: (args: {open: boolean}) => void;
```
korvin89, line 8: I`ve used object as function`s args to have flexibility in possible future improvements

src/components/Select/utils.ts
```diff
@@ -203,7 +207,7 @@ export const findItemIndexByQuickSearch = (
 
         const optionText = getOptionText(item);
 
-        return new RegExp(quickSearch, 'i').test(optionText);
+        return getEscapedRegExp(quickSearch).test(optionText);
```
korvin89, line 210: Now we crushing on input like `'\'` =/

src/components/Select/types.ts
amje: Not sure about that. We should do it either everywhere in the code or nowhere. There is exactly the same prop in Popover component. Callbacks should be consistent.

src/components/Select/components/SelectControl/SelectControl.tsx
```diff
@@ -59,11 +61,24 @@ export const SelectControl = React.forwardRef<HTMLButtonElement, ControlProps>((
...
-    const handleClick = () => setActive(!active);
+    const handleClick = React.useCallback(() => setActive(!active), [setActive, active]);
+
+    if (renderControl) {
+        return (
+            <div
+                ref={handleControlRef as React.Ref<HTMLDivElement>}
+                className={className}
+                onClick={disabled ? undefined : handleClick}
+                onKeyDown={disabled ? undefined : onKeyDown}
```
amje, line 72: This element is not focusable, keydown wont trigger

amje (approved):

PR #140, opened 2022-04-22
feat(Tooltip): add delay before opening/closing

d3m1d0v:
add support for delay before opening or closing a tooltip


src/components/Tooltip/Tooltip.tsx
```diff
@@ -79,3 +66,38 @@ export const Tooltip = ({children, content, placement = DEFAULT_PLACEMENT}: Tool
...
+            const timeout = setTimeout(showTooltip, openDelay);
+            return () => {
+                clearTimeout(timeout);
+            };
+        }
+        if (!anchorHovered && tooltipVisible) {
+            const timeout = setTimeout(hideTooltip, closeDelay);
+            return () => {
+                clearTimeout(timeout);
+            };
```
amje, line 84: Can we always clearTimeout in hook's cleanup to avoid code duplication? It's no harm to do so.

src/components/utils/useBoolean.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export function useBoolean(
+    initialState: boolean | (() => boolean),
+): [boolean, SetTrue, SetFalse, Toggle] {
+    const [value, setValue] = useState<boolean>(initialState);
+    return [
+        value,
+        useCallback(() => setValue(true), []),
+        useCallback(() => setValue(false), []),
+        useCallback(() => setValue((val) => !val), []),
```
amje, line 15: Aren't we need `val` as a dependency here?

src/components/Tooltip/Tooltip.tsx
d3m1d0v: Fixed

src/components/utils/useBoolean.ts
d3m1d0v: No, we aren't. `toggle` callback should not be recreated when the state changes

amje (approved):

PR #141, opened 2022-04-23
feat(Text): a new component for working with typography

IsaevAlexandr:
Hi, this is related to https://github.com/yandex-cloud/uikit/issues/128

what was done:
- added `Text` component;
- added `text` and `colorText` utilities to work with inline classes in existing codebase;
- removed duplicate code `@mixin text-display-3` in `styles/mixins.scss`;
- all new entities have the ability to show jsdoc on hover
Screenshot 2022-04-23 at 17 33 54
image

also added documentation to storybook

image


src/components/ColorText/ColorText.scss
```diff
@@ -0,0 +1,42 @@
+@import '../variables';
+@import '../../../styles/mixins';
+
+$block: '.#{$ns}color-text';
+
+$textColorMap: (
+    'inherit' inherit,
```
ogonkov, line 7: I guess you don't need component for this

src/components/ColorText/ColorText.tsx
```diff
@@ -0,0 +1,49 @@
...
+ * Storybook: https://preview.yandexcloud.dev/uikit/?path=/story/colors--texts
+ *
+ * ---
+ * For example:
+ * ```jsx
+ * // "color-text color-text_color_inherit some-class"
+ * colorText({color: 'inherit'}, 'some-class')
+ * ```
+ */
+export const colorText = ({color}: ColorTextBase, className?: string) => b({color}, className);
```
ogonkov, line 49: I think export should match file name

src/components/Text/Text.scss
```diff
@@ -0,0 +1,73 @@
+@import '../variables';
+@import '../../../styles/mixins';
```
ogonkov, line 2: Can we replace `@import` with `@use`?

src/components/Text/Text.tsx
```diff
@@ -0,0 +1,153 @@
...
+ *```
+ */
+export const text = ({type = 'body1', ellipsis}: TextBase, className?: string) =>
+    b({type, ellipsis}, className);
+
+export interface TextProps extends TextBase, ColorTextBase {
+    /**
+     * Ability to override default html tag
+     */
+    as?: keyof Pick<
```
ogonkov, line 91: `as` is also attribute of `<a>`, i guess it is better to use `component` instead

src/components/Text/Text.tsx
```diff
@@ -0,0 +1,153 @@
...
+        'span' | 'div' | 'p' | 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'pre' | 'code'
+    >;
+    style?: React.CSSProperties;
+    className?: string;
+    content?: React.ReactNode;
+    children?: React.ReactNode;
+    /**
+     * Use this for locators in your tests
+     */
+    testId?: string;
```
ogonkov, line 102: qa?

src/components/Text/Text.tsx
```diff
@@ -0,0 +1,153 @@
...
+    children,
+    content,
+    type,
+    className,
+    ellipsis,
+    color,
+    testId,
+    ...rest
+}) => {
+    const Tag = ellipsis ? 'div' : as;
```
ogonkov, line 142: Why?

src/components/Text/Text.tsx
IsaevAlexandr: added

src/components/ColorText/ColorText.tsx
IsaevAlexandr: there is no mistake here. `ColorText` is a component that exports the`colorText` utility. This is not a React component, it is component of uikit package

src/components/Text/Text.scss
IsaevAlexandr: thank you for paying attention, I didn't know that this functionality would be dropped in future releases
https://github.com/sass/sass/blob/main/accepted/module-system.md#timeline

src/components/ColorText/ColorText.scss
ValeraS: Why not? In my practice it sometimes helps.

src/components/Text/Text.tsx
```diff
@@ -0,0 +1,151 @@
...
+export const text = ({type = 'body1', ellipsis}: TextBase, className?: string) =>
+    b({type, ellipsis}, className);
+
+export interface TextProps extends TextBase, ColorTextBase {
+    /**
+     * Ability to override default html tag
+     */
+    as?: keyof Pick<
+        JSX.IntrinsicElements,
+        'span' | 'a' | 'div' | 'p' | 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'pre' | 'code'
```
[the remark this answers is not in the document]
IsaevAlexandr: supported all type of tags

src/components/Text/Text.tsx
```diff
@@ -0,0 +1,151 @@
...
+ *```jsx
+ * import {text} from '@yandex-cloud/uikit';
+ *
+ * // textStyles = 'text text_type_display1 some-class-name'
+ * const textStyles = text({type: 'display1'}, 'some-class-name'));
+ *
+ * <span className={textStyles}>some text</span>
+ * ```
+ */
+export const Text: React.FC<TextProps> = ({
```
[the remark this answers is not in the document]
IsaevAlexandr: you don't need to use any type of ref things with this component. If you want to do this just use `text` utility and certain element.
for example:
```jsx
const Component = () => {
    const ref = React.useRef<HTMLDivElement>(null);
    return (
        <div ref={ref} className={text()}>
            some text
        </div>
    );
};
```
The keyword here is `explicit` usage of ref

src/components/Text/Text.tsx
```diff
@@ -0,0 +1,151 @@
...
+     *      - 4: font-size: 48px; line-height: 52px; font-weight: 900;
+     * - code:
+     *      - 1: font-size: 12px; line-height: 18px; font-weight: 400; font-family: var(--yc-font-family-monospace);
+     *      - 2: font-size: 14px; line-height: 20px; font-weight: 400; font-family: var(--yc-font-family-monospace);
+     *      - 3: font-size: 16px; line-height: 14px; font-weight: 400; font-family: var(--yc-font-family-monospace);
+     *      - inline1: font-size: 12px; line-height: 14px; font-weight: 400; font-family: var(--yc-font-family-monospace);
+     *      - inline2: font-size: 14px; line-height: 16px; font-weight: 400; font-family: var(--yc-font-family-monospace);
+     *      - inline3: font-size: 16px; line-height: 20px; font-weight: 400; font-family: var(--yc-font-family-monospace);
+     */
+    type?:
```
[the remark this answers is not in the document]
IsaevAlexandr: done

src/components/Text/Text.tsx
```diff
@@ -0,0 +1,151 @@
...
+        | 'body2'
+        | 'body3'
+        | 'bodyShort'
+        | 'caption1'
+        | 'caption2'
+        | 'header1'
+        | 'header2'
+        | 'subheader1'
+        | 'subheader2'
+        | 'subheader3';
```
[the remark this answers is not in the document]
IsaevAlexandr: done

src/components/ColorText/ColorText.tsx
```diff
@@ -0,0 +1,48 @@
+import {block} from '../utils/cn';
```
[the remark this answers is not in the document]
IsaevAlexandr: colorText moved in `Text` folder

src/components/ColorText/ColorText.tsx
IsaevAlexandr: done

src/components/Text/Text.tsx
```diff
@@ -0,0 +1,151 @@
...
+     * Ability to override default html tag
+     */
+    as?: keyof Pick<
+        JSX.IntrinsicElements,
+        'span' | 'a' | 'div' | 'p' | 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'pre' | 'code'
+    >;
+    style?: React.CSSProperties;
+    className?: string;
+    content?: React.ReactNode;
+    children?: React.ReactNode;
```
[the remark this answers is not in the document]
IsaevAlexandr: removed content prop

src/components/Text/Text.tsx
```diff
@@ -0,0 +1,151 @@
...
+    type,
+    className,
+    ellipsis,
+    color,
+    qa,
+    ...rest
+}) => {
+    return (
+        <Tag
+            className={text({type, ellipsis}, color ? colorText({color}, className) : className)}
```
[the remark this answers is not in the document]
IsaevAlexandr: it's useful if you want to add additional class to element. This technic called [bem mix](https://en.bem.info/methodology/key-concepts/#mix)

src/components/Text/text/text.ts
```diff
@@ -0,0 +1,87 @@
...
+     *      - 4: font-size: 48px; line-height: 52px; font-weight: 900;
+     * - code:
+     *      - 1: font-size: 12px; line-height: 18px; font-weight: 400; font-family: var(--yc-font-family-monospace);
+     *      - 2: font-size: 14px; line-height: 20px; font-weight: 400; font-family: var(--yc-font-family-monospace);
+     *      - 3: font-size: 16px; line-height: 14px; font-weight: 400; font-family: var(--yc-font-family-monospace);
+     *      - inline=1: font-size: 12px; line-height: 14px; font-weight: 400; font-family: var(--yc-font-family-monospace);
+     *      - inline-2: font-size: 14px; line-height: 16px; font-weight: 400; font-family: var(--yc-font-family-monospace);
+     *      - inline-3: font-size: 16px; line-height: 20px; font-weight: 400; font-family: var(--yc-font-family-monospace);
+     */
+    typography?: typeof TYPOGRAPHY_VARIANTS[number];
```
amje, line 63: Maybe call it `variant`?

src/components/Text/text/text.ts
```diff
@@ -0,0 +1,87 @@
...
+     *      - 2: font-size: 24px; line-height: 28px; font-weight: 900;
+     * - subheader:
+     *      - 1: font-size: 13px; line-height: 18px; font-weight: 900;
+     *      - 2: font-size: 15px; line-height: 20px; font-weight: 900;
+     *      - 2: font-size: 17px; line-height: 24px; font-weight: 900;
+     * - display:
+     *      - 1: font-size: 28px; line-height: 36px; font-weight: 900;
+     *      - 2: font-size: 32px; line-height: 40px; font-weight: 900;
+     *      - 3: font-size: 40px; line-height: 48px; font-weight: 900;
+     *      - 4: font-size: 48px; line-height: 52px; font-weight: 900;
```
amje, line 54: `font-weight` is controlled by [variables](https://github.com/yandex-cloud/uikit/blob/main/styles/typography.scss#L9-L15), and can be changed outside.

src/components/Text/text/text.ts
```diff
@@ -0,0 +1,87 @@
...
+    'caption-1',
+    'code-3',
+    'code-inline-3',
+    'code-2',
+    'code-inline-2',
+    'code-1',
+    'code-inline-1',
+] as const;
+
+export interface TextBase {
```
amje, line 31: If it's a props them it should have `Props` prefix

src/components/Text/text/text.ts
```diff
@@ -0,0 +1,87 @@
+import {block} from '../../utils/cn';
+
+import './text.scss';
+
+const b = block('text');
+
+export const TYPOGRAPHY_VARIANTS = [
```
amje, line 7: `TEXT_VARIANTS` is better imo, it shows connection to `Text` component like `TEXT_COLORS` does.

src/components/Text/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './Text';
+export {TYPOGRAPHY_VARIANTS, text} from './text/text';
+export {TEXT_COLORS, colorText} from './colorText/colorText';
```
amje, line 3: Let's name "not a component or hook or class" files in dash-case.

src/components/Text/text/text.scss
```diff
@@ -0,0 +1,73 @@
+@use '../../variables' as variables;
+@use '../../../../styles/mixins' as mixins;
```
amje, line 2: By default its already in these namespaces (ns = filename)

amje (changes requested):

src/components/Text/index.ts
IsaevAlexandr: `colorText` is a component ("block" in BEM paradigm). But yes it's not a "react" component

IsaevAlexandr: and it's better the entity name and the file name match

src/components/Text/text/text.ts
IsaevAlexandr: the main idea of this comments is ability to quick look at the component and understand what it does. Of course we can override every `token` of text variant.

I will add note about it in this comment

src/components/Text/text/text.scss
IsaevAlexandr: removed `as variables` and `as mixins;`;

PR #160, opened 2022-05-16
fix(RadioButton): fixed types

tumenbaev:
RadioButton uses only 4 props from [RadioButtonOption](https://github.com/yandex-cloud/uikit/blob/64440f7e812bdad005eab51618cbf43e491c2287/src/components/RadioButton/RadioButton.tsx#L38)


amje (approved):

amje:
Thanks for fixing this!

PR #172, opened 2022-05-26
fix(Toaster): fix toast paddings

korvin89:
- Add extra padding to toasts in case of icon using

<details>
<summary>Before</summary>

image

</details>

<details>
<summary>After</summary>

image

</details>

- change default `testIdAttribute` in testing-library


src/components/Toaster/Toast/Toast.scss
```diff
@@ -192,9 +207,17 @@ $block: '.#{$ns}toast';
     }
 }
 
+// with extra space for icon placement
 @keyframes toast-set-indents {
     to {
         margin-bottom: 10px;
-        padding: 18px;
+        padding: 16px 16px 16px 48px;
```
amje, line 214: It's better to not animate "reflow" properties, instead it can be done with `transform`

amje (approved):

Issue #169, opened 2022-05-23
Disabled popover cursor is pointer


ierehon1905:
Hi, thank you for your awesome library.

I was working with `Popover` component and discovered that even if I disable it the cursor stays as `pointer`. Is this a desired behaviour.

Here is a [link](https://preview.yandexcloud.dev/uikit/?path=/story/components-popover--default&args=disabled:true) to you storybook with `Popover` disabled.

PR #182, opened 2022-06-04
fix: default cursor on disabled popover

ierehon1905:
Solves #169

### DIFF
--- a/src/components/Popover/Popover.scss
+++ b/src/components/Popover/Popover.scss
@@ -8,6 +8,10 @@ $block: '.#{$ns}popover';
     display: inline-block;
     position: relative;
 
+    &_disabled {
+        cursor: default;
+    }
+
     $class: &;
 
     &__tooltip {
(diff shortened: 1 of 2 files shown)


src/components/Popover/Popover.scss
```diff
@@ -8,6 +8,10 @@ $block: '.#{$ns}popover';
     display: inline-block;
     position: relative;
 
+    &_disabled {
+        cursor: default;
+    }
+
```
ierehon1905, line 14: I am not sure if it is the correct order to put to so notify me if this need to be fixed

amje: What about inverting the rule: `:not(&_disabled) {cursor: pointer}`? So we don't override the cursor of underlying element in Popover whet it's disabled.

ierehon1905: Oh, ok

I slightly modified it by adding an `&` at the beggining like follows:

`&:not(&_disabled) {cursor: pointer}`

without it it was a mess

amje (approved):

PR #186, opened 2022-06-07
feat(I18N): add configure function for manage internal variables like…

dihar:
… language


src/components/utils/configure.ts
```diff
@@ -0,0 +1,31 @@
+export enum LANGS {
```
amje, line 1: Let's name it `Lang`

amje (changes requested):

amje (approved):

PR #190, opened 2022-06-10
fix(List): fix getContainer for sortable container

astandrik:
Looks like List component is currently broken for the sortable-list scenario.

item activation calls scrollToIndex that calls getContainer() that tries to call getWrappedInstance which is forbidden for containers without {withRef: true} option


amje (approved):

astandrik:
https://github.com/yandex-cloud/uikit/issues/199

PR #196, opened 2022-06-15
feat(RadioGroup): improve radio-group components a11y

### LINKED ISSUE #179: RadioButton does not call onFocus and onBlur
It seems like the `RadioButton` component does not call the `onFocus ` and `onBlur` callbacks. This sometimes prevents the error messages in forms to be shown.

Maybe we could make so those method would be called like on a native radio/checkbox

amje:
Closes #179

### DIFF
--- a/src/components/RadioButton/__stories__/RadioButton.stories.tsx
+++ b/src/components/RadioButton/__stories__/RadioButton.stories.tsx
@@ -14,10 +14,12 @@ const options: RadioButtonOption[] = [
     {value: 'Value 3', content: 'Value 3'},
 ];
 
-const DefaultTemplate: Story<RadioButtonProps> = (args) => <RadioButton {...args} />;
+const DefaultTemplate: Story<RadioButtonProps> = (args) => {
+    const [value, setValue] = React.useState<string>(options[0].value);
+    return <RadioButton {...args} value={value} onUpdate={setValue} />;
+};
 export const Default = DefaultTemplate.bind({});
 Default.args = {
-    value: options[0].value,
     options,
 };
 
--- a/src/components/utils/useRadioGroup.ts
+++ b/src/components/utils/useRadioGroup.ts
@@ -4,7 +4,17 @@ import {ControlGroupProps} from '../types';
 import {useUniqId} from './useUniqId';
 
 export function useRadioGroup(props: ControlGroupProps) {
-    const {name, value, defaultValue, options = [], disabled, onUpdate, onChange} = props;
+    const {
+        name,
+        value,
+        defaultValue,
+        options = [],
+        disabled,
+        onUpdate,
+        onChange,
+        onFocus,
+        onBlur,
+    } = props;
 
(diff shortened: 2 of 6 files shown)


src/components/RadioButton/__stories__/RadioButton.stories.tsx
```diff
@@ -14,10 +14,12 @@ const options: RadioButtonOption[] = [
     {value: 'Value 3', content: 'Value 3'},
 ];
 
-const DefaultTemplate: Story<RadioButtonProps> = (args) => <RadioButton {...args} />;
+const DefaultTemplate: Story<RadioButtonProps> = (args) => {
+    const [value, setValue] = React.useState<string>(options[0].value);
```
ogonkov, line 18: Why not use storybook controls?

src/components/utils/useRadioGroup.ts
```diff
@@ -28,14 +38,22 @@ export function useRadioGroup(props: ControlGroupProps) {
         [isControlled, onUpdate, onChange],
     );
 
+    const containerProps = {
+        role: 'radiogroup',
+        'aria-label': props['aria-label'],
+        'aria-labelledby': props['aria-labelledby'],
```
ogonkov, line 44: This role also allows `aria-disabled` https://w3c.github.io/aria/#radiogroup

src/components/RadioButton/__stories__/RadioButton.stories.tsx
ogonkov: Nope, storybook is not intended to be interactive, it is more like UI snapshots than demos.

amje: It's kinda weird when you open a demo and it's not working properly. And the only way to change the value is to do it by hand.

src/components/utils/useRadioGroup.ts
amje: Yeah... Currently the component doesn't implement disabled prop at all. What do you think if it will simply set `disabled` on option components in this case?

amje: Fixed

src/components/utils/useRadioGroup.ts
```diff
@@ -28,14 +38,23 @@ export function useRadioGroup(props: ControlGroupProps) {
         [isControlled, onUpdate, onChange],
     );
 
+    const containerProps = {
+        role: 'radiogroup',
+        'aria-disabled': disabled,
+        'aria-label': props['aria-label'],
+        'aria-labelledby': props['aria-labelledby'],
```
ogonkov, line 45: if `aria-label` set, `aria-labelledby` should not be setted

ogonkov: and vice versa

ogonkov: Looks like i was wrong

ogonkov (approved):

PR #208, opened 2022-06-23
fix(TextInput): set paddings for clear cross depending on size

helenjer:
Clear cross was too close to the right border on l and xl sizes.

This PR:
Sets padding for cross depending on size.


amje:
@helenjer Changed PR's title to be `fix`, not `feat`. Keep in mind on merging.

PR #216, opened 2022-07-01
refactor(ThemeProvider): split component

ogonkov:


src/components/theme/getThemeValue.ts
```diff
@@ -0,0 +1,5 @@
+import {getSystemTheme} from './getSystemTheme';
```
amje, line 1: What is the point of extracting this into single file if the only place it used is ThemeProvider? Same question for getSystemTheme

src/components/theme/updateBodyClassName.ts
```diff
@@ -0,0 +1,18 @@
+import {block} from '../utils/cn';
```
amje, line 1: This file and getDarkMediaMatch can be combined into one `dom-utils.ts`

src/components/theme/ThemeProvider.tsx
```diff
@@ -12,95 +15,46 @@ interface ThemeProviderDefaultProps {
...
-    componentDidUpdate(prevProps: ThemeProviderProps, prevState: ThemeProviderState) {
-        if (prevState.theme !== this.state.theme) {
-            const themeValue = this.getThemeValue(this.state.theme);
-            this.setState({themeValue});
-            this.updateBodyClassName(themeValue);
+        Partial<ThemeProviderDefaultProps>,
+        PropsWithChildren<{}> {}
+
+export function ThemeProvider({theme = DEFAULT_THEME, children}: ThemeProviderProps) {
+    const [_theme, _setTheme] = useState(theme);
```
amje, line 22: Let's use normal names but use alias for theme from props

amje (changes requested):

src/components/theme/getThemeValue.ts
ogonkov: Small modules is easy to refactor, you don't have to move them from one file to another
Changes to small module is easier to code review, because it scoped to module with single responsibility
It is easier to rebase smaller modules

src/components/theme/useSystemTheme.ts
```diff
@@ -0,0 +1,20 @@
...
+    const [theme, setTheme] = useState<'light' | 'dark' | null>(null);
+
+    useEffect(() => {
+        function onChange(event: MediaQueryListEvent) {
+            setTheme(event.matches ? 'dark' : 'light');
+        }
+
+        const matcher = getDarkMediaMatch();
+
+        matcher.addEventListener('change', onChange);
```
ValeraS, line 14: Do we want support Safari 13? In Safari 13 this code will throw.

ogonkov: It is not in our browserslist config

ValeraS: But not in Readme. In most cases, this library works fine in Safari 13. With this changes, it will not work at all.

ogonkov: What readme?..

ValeraS: Does readme contain information about supported browsers? But it's not a main question.

src/components/theme/ThemeProvider.tsx
```diff
@@ -1,106 +1,61 @@
...
+    useEffect(() => {
+        setThemeValue(getThemeValue(theme));
+    }, [theme]);
+    useEffect(() => {
+        updateBodyClassName(themeValue);
+    }, [themeValue]);
+
+    const systemTheme = useSystemTheme();
+    useEffect(() => {
+        if (!systemTheme || theme !== 'system') {
```
amje, line 38: Seems like that we can merge this whole effect into one that's in line 29. Otherwise there will be 2 calls for `getSystemTheme` when `system` theme is activated.

ValeraS: also, I think, we can calculate  `themeValue` during render, so we don't need useState and useEffect for it.

src/components/theme/ThemeProvider.tsx
```diff
@@ -1,106 +1,59 @@
...
+        setTheme(themeProp);
+    }, [themeProp]);
+
+    const [themeValue, setThemeValue] = useState(getThemeValue(theme));
+    const systemTheme = useSystemTheme();
+    useEffect(() => {
+        if (systemTheme && theme === 'system') {
+            setThemeValue(systemTheme);
+
+            return;
```
ValeraS, line 34: let's simplify this:
```suggestion
    const systemTheme = useSystemTheme();
    const themeValue = theme === 'system' ? systemTheme : theme;
```

src/components/theme/ThemeProvider.tsx
```diff
@@ -1,106 +1,53 @@
...
+            setTheme,
+        }),
+        [theme, themeValue],
+    );
+
+    const themeValueContext = useMemo(() => ({themeValue}), [themeValue]);
+
+    return (
+        <ThemeContext.Provider value={contextValue}>
+            <ThemeValueContext.Provider value={themeValueContext}>
```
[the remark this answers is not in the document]
ogonkov: If someone use `ThemeValueContext` directly, that would be breaking change for them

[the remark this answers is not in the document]
ValeraS: Is `ThemeValueContext` part of the public API?

src/components/theme/useSystemTheme.ts
ogonkov: Revert `addListener` usage

src/components/theme/ThemeProvider.tsx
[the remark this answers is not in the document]
ogonkov: I don't think so, just side note

src/components/theme/ThemeContext.ts
```diff
@@ -1,10 +1,11 @@
 import React from 'react';
 
 import {DEFAULT_THEME} from './constants';
+import {Theme} from './types';
 
 export interface ThemeContextProps {
     theme: string;
-    setTheme: (newTheme: string) => void;
+    setTheme: (newTheme: Theme) => void;
```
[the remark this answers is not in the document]
ValeraS: Let's add `string` to `Theme` type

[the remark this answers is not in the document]
ogonkov: Why? Valid strings should works as before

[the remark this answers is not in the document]
ValeraS: What is a valid string? If I define a new theme on my side how can I pass in my theme value?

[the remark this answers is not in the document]
ogonkov: `dark`, `light`, `system`

It doesn't looks like user can define own themes

amje (approved):

src/components/theme/ThemeProvider.tsx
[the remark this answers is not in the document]
amje: @ValeraS  In the next major ThemeValueContext will be removed. There is no need for it. But let's keep it for now.

PR #242, opened 2022-07-14
feat: add animations timing control

ogonkov:
Adds global `--yc-animation-multiplier` that controls components animation speed.

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -33,7 +33,23 @@ const withContextProvider = (Story, context) => {
     );
 };
 
-export const decorators = [withTheme, withMobile, withLang, withContextProvider];
+function withAnimationSpeed(Story, context) {
+    const {animationSpeed} = context.globals;
+
+    if (typeof animationSpeed === 'number') {
+        document.body.style.setProperty('--yc-animation-multiplier', `${1 / animationSpeed}`);
+    }
+
+    return <Story {...context} />;
+}
+
+export const decorators = [
+    withTheme,
+    withMobile,
+    withLang,
+    withContextProvider,
+    withAnimationSpeed,
+];
 
 export const parameters = {
     docs: {
@@ -61,6 +77,8 @@ export const parameters = {
         },
     },
 };
+
+const animationSpeed = [0.2, 0.5, 0.75, 1, 1.25, 1.5, 1.75, 2, 10];
 
 export const globalTypes = {
     theme: {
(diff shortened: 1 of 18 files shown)


pseudopirate (approved):

Marginy605 (approved):

styles/styles.scss
```diff
@@ -22,6 +22,8 @@
     @include yc-brand;
     @include yc-colors-social;
 
+    --yc-animation-multiplier: 1;
```
ValeraS, line 25: Do we need to set it explicitly? We already have the same fallback for it.

.storybook/preview.js
```diff
@@ -62,6 +78,28 @@ export const parameters = {
     },
 };
 
+const animationSpeed = [
+    0.1, 0.2, 0.5, 0.75, 1, 1.25, 1.5, 1.75, 2,
+    // Special value to make speed so fast that it could not be recognized
+    10000,
```
ValeraS, line 84:
```suggestion
    Infinity,
```

ogonkov: We are calculating timing based on this number, we need some real number here

ValeraS: `1 / Infinity === 0`

styles/styles.scss
ValeraS: Of cause not, but we set default in fallback and user can change it by setting the var.

.storybook/preview.js
ogonkov: But we don't need 0, we need some very small value to be sure that animation in JS will trigger

styles/styles.scss
ogonkov: Yes, that is how it should work

.storybook/preview.js
ValeraS: we need it in storybook?

styles/styles.scss
ValeraS: and why are we setting the var?

.storybook/preview.js
ogonkov: I think it nice to have all options to see that components works properly with disabled animations

styles/styles.scss
ogonkov: To debug animations in project, for example

.storybook/preview.js
ValeraS: and animation is disabled when duration is 0. I don't understand, how is this setting in storybook connected to testing animation?

styles/styles.scss
ValeraS: why are we setting it here?

.storybook/preview.js
ogonkov: As i said before
> we need some very small value to be sure that animation in JS will trigger

styles/styles.scss
ogonkov: To see it dev-tools and adjust it if needed

korvin89 (approved):

.storybook/preview.js
ogonkov: How it should look at the component level?

ValeraS: Don't use `--yc-animation-multiplier` for critical animation, as was done in this PR for the `toaster` component. Or, for example, add small amount of time to result duration in critical animations.

PR #248, opened 2022-07-18
feat(Toaster)!: rename `timeout` to `autoHiding`

### LINKED ISSUE #241: Remove `allowAutoHiding` from `<Toast/>`
Prop `allowAutoHiding` is looks redundant, and seems could be replaced with some special value for `timeout`, like `-1` or `null`.

ogonkov:
Fixes #241


amje:
@ogonkov What do you think about using `0` instead of `null`?

ogonkov:
Yes, i think it will work

amje (approved):

PR #262, opened 2022-07-23
fix(Button): add contrast for selected state in dark theme

amje:


src/components/Button/Button.scss
```diff
@@ -194,7 +194,9 @@ $block: '.#{$ns}button';
                 --yc-button-background-color: var(--yc-color-base-selection);
                 --yc-button-background-color-hover: var(--yc-color-base-selection-hover);
 
-                @include button-text-color(var(--yc-color-text-special));
+                @include button-text-color(
+                    var(--yc-my-color-brand-text, var(--yc-color-text-special))
```
ValeraS, line 198: Do we need fallback here, other brand vars don't have it?

ogonkov (approved):

src/components/Button/Button.scss
amje: It's a new brand variable, so we need fallback to keep compatibility, we can remove the fallback later in the next major

ValeraS (approved):

PR #265, opened 2022-07-24
feat(Text): added weight prop to override css font-weight property

IsaevAlexandr:
Where is ability to override `font-weight` css property.

it is very controversial ability because now you can rich more text variants than the design system provides. It's can provide some issues with migrations later.

But, for my opinion, it's facilitates the use of this component and make is more intuitive and developer friendly

p.s. naming is not final, if you have more suitable names , let's use them

### DIFF
--- a/src/components/Text/text/text.ts
+++ b/src/components/Text/text/text.ts
@@ -28,6 +28,15 @@ export const TEXT_VARIANTS = [
     'code-inline-1',
 ] as const;
 
+const TEXT_FONT_WEIGHT_VARIANTS = [
+    'bold',
+    'normal',
+    'lighter',
+    'bolder',
+    'inherit',
+    'unset',
+] as const;
+
 export interface TextBaseProps {
     /**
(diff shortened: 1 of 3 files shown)


src/components/Text/text/text.ts
```diff
@@ -28,6 +28,15 @@ export const TEXT_VARIANTS = [
     'code-inline-1',
 ] as const;
 
+const TEXT_FONT_WEIGHT_VARIANTS = [
```
ogonkov, line 31: I guess design system support only limites set of font  weight

amje: And typically value is a number (100-900)

IsaevAlexandr: adding numeric value is not the problem. I'll do it soon

We need to understand what is the legitimate way to add the `font-weight` property to this component, not taking in mind what we can pass css class with right font-weight in `className` prop - it is provides very ugly dev exp.

Alternative way is have different `variants` for every variant: for example `body-1-bold` but it seems that this is the way to madness.

PR #266, opened 2022-07-25
fix(TextInput)!: remove incorrect resize attribute styles

ValeraS:
textarea does not have resize attribute.

Revert "fix: correctly show textarea resize control"

Reverts yandex-cloud/uikit#33

### DIFF
--- a/src/components/TextInput/TextInput.scss
+++ b/src/components/TextInput/TextInput.scss
@@ -32,10 +32,7 @@ $block: '.#{$ns}text-input';
 
         &_type {
             &_textarea {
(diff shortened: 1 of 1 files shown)


src/components/TextInput/TextInput.scss
```diff
@@ -32,10 +32,7 @@ $block: '.#{$ns}text-input';
 
         &_type {
             &_textarea {
-                &:not([resize]),
```
ogonkov, line 35: May be we should add modifier instead of reverting it?

amje (approved):

PR #270, opened 2022-07-27
refactor (Popover): popover refactoring (-> functional), docs improvement

kseniya57:


src/components/Popover/__tests__/Popover.test.tsx
```diff
@@ -0,0 +1,299 @@
...
+    </Popover>
+);
+
+const waitForTooltipOpenedStateChange = (shouldOpen?: boolean) =>
+    jest.advanceTimersByTime(delayByBehavior[PopoverBehavior.Delayed][shouldOpen ? 0 : 1]);
+
+const checkIfPopoverOpened = () => {
+    const popover = screen.queryByTestId('popover-tooltip');
+
+    expect(popover).toHaveClass('yc-popup_open');
```
ogonkov, line 31: I suggest to check that popover is visible instead

src/components/Popover/__tests__/Popover.test.tsx
```diff
@@ -0,0 +1,299 @@
+import React from 'react';
```
ogonkov, line 1: Can we add this tests via separate PR please? To check that current implementation pass them

src/components/Popover/__tests__/Popover.test.tsx
kseniya57: I've tried that, but it's not working

src/components/Popover/__tests__/Popover.test.tsx
kseniya57: Yes

kseniya57: https://github.com/yandex-cloud/uikit/pull/285

kseniya57:
@ogonkov @ValeraS @amje Hi! Can you please review the changes?

src/components/Popover/README.md
```diff
@@ -43,32 +89,26 @@ Component with rendered raw html (use `content` for plain text) and close on mou
...
-      onClick: () => alert('Lorem ipsum onClick'),
+      text: 'Link with an onClick handler',
+      onClick: () => alert('The link is clicked'),
     },
   ]}
   placement={['right', 'bottom']}
   title="Simple tooltip"
   htmlContent={
-    '<b>Lorem ipsum</b> dolor sit <a href="https://example.com" target="_blank">amet</a>, at scelerisque suspendisse'
+    'Tooltip\'s <b>html</b> content. Learn more <a href="https://storybook.ui.yandex-team.ru/?path=/docs/uikit_components-popover--default" target="_blank">here</a>'
```
amje, line 103: remove ya-team link

src/components/Popover/components/Trigger/Trigger.tsx
```diff
@@ -0,0 +1,71 @@
...
+            if (nextOpen) {
+                openTooltip();
+                closedManually.current = false;
+            } else {
+                closeTooltip();
+                closedManually.current = true;
+            }
+        };
+
+        toggleTooltip();
```
amje, line 67: What is the point of creating and calling function here?

src/components/Popover/hooks/useOpen.ts
```diff
@@ -0,0 +1,109 @@
...
+    const closeTooltipDelayed = useCallback(() => {
+        closingTimeout.current = setTimeout(() => {
+            closingTimeout.current = null;
+            closeTooltip();
+        }, delayClosing ?? defaultDelayClosing);
+    }, [closeTooltip, defaultDelayClosing, delayClosing]);
+
+    return {
+        isOpen,
+        isClosing: Boolean(closingTimeout.current),
```
amje, line 101: This should be a state if we want a truthful value.

amje (changes requested):

src/components/Popover/components/Trigger/Trigger.tsx
kseniya57: Function name explains the code

src/components/Popover/hooks/useOpen.ts
kseniya57: We don't need to force component update when this variable changes

kseniya57: It's used in ` onMouseLeave` listener, that listener is not memoized, because it's passed to html element, so the values inside are always truthful. Maybe better to just return ref `closingTimeout` to make it more clear.

amje (approved):

PR #274, opened 2022-07-27
refactor(Breadcrumbs): refactor more button, separator and showcase

ValeraS:


src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -179,16 +179,21 @@ export class Breadcrumbs extends React.Component<BreadcrumbsProps, BreadcrumbsSt
     }
 
     renderMoreItem() {
-        const {popupStyle, popupPlacement, renderItemDivider} = this.props;
         const {hiddenItems} = this.state;
+        if (hiddenItems.length === 0) {
+            return null;
+        }
```
ogonkov, line 185: not sure that it should be lifted from component here

vrozaev: Same here.

ogonkov: I think about `BreadcrumbsMore` as composite component, than component to render specific dropdown.

I agree that it would be nice to have such component, if it was used somewhere else. But now it looks like more like refactor for refactor to me.

vrozaev (approved):

PR #279, opened 2022-07-28
chore(Breadcrumbs): add generic parameter <T extends BreadcrumbsItem>

ma-efremoff:


src/components/Breadcrumbs/BreadcrumbsItem.tsx
```diff
@@ -3,43 +3,43 @@ import {Link} from '../Link';
...
-    data,
-    isCurrent,
-    isPrevCurrent,
-    renderItem,
-}: Props) {
-    const {text, href, action} = data;
+export class BreadcrumbsItem<
+    T extends IBreadcrumbsItem = IBreadcrumbsItem,
+> extends React.PureComponent<Props<T>> {
+    render() {
```
ogonkov, line 20: Why you convert FC component to class?

ma-efremoff: It is impossible to create generic components with React.memo

vrozaev (approved):

src/components/Breadcrumbs/BreadcrumbsItem.tsx
ogonkov: It is possible https://stackoverflow.com/a/60170425

src/components/Breadcrumbs/BreadcrumbsItem.tsx
```diff
@@ -3,21 +3,23 @@ import {Link} from '../Link';
...
 }
 
 const b = block('breadcrumbs');
 
-export const BreadcrumbsItem = React.memo(function BreadcrumbsItem({
+function BreadcrumbsItem<T extends IBreadcrumbsItem = IBreadcrumbsItem>({
     data,
     isCurrent,
     isPrevCurrent,
     renderItem,
```
ogonkov, line 21: Isn't it enough to make function generic?

```suggestion
export const BreadcrumbsItem = React.memo(function BreadcrumbsItem<T extends IBreadcrumbsItem = IBreadcrumbsItem>({
    data,
    isCurrent,
    isPrevCurrent,
    renderItem,
```

ma-efremoff: Yes, it is. But the last version has less changed rows. But if you insist I will bring back the version with function.

ogonkov: There is also `displayName` disappear, and i personally like named instead of `default`

ma-efremoff: Please suggest your fix for the file.

ogonkov: I suggest (above) to just add generic to function and keep everything other unchanged

```jsx
export const BreadcrumbsItem = React.memo(function BreadcrumbsItem<T extends IBreadcrumbsItem = IBreadcrumbsItem>({
    data,
    isCurrent,
    isPrevCurrent,
    renderItem,
}: Props<T>) {
    //
});

BreadcrumbsItem.displayName = 'Breadcrumbs.Item';
```

ma-efremoff: Please provide full solution, your one has no any generic-component, and there is nothing you can use for type-cast as described in link above.

ogonkov (approved):

PR #280, opened 2022-07-28
feat(Popover): add prop to disable link styles

ogonkov:


src/components/Popover/Popover.tsx
```diff
@@ -260,12 +263,22 @@ export class Popover extends React.Component<
...
             <Popup
                 anchorRef={anchorRef || this.controlRef}
-                className={b('tooltip', {theme, size, ['with-close']: hasClose}, tooltipClassName)}
+                className={b(
+                    'tooltip',
+                    {
+                        theme,
+                        size,
+                        ['with-close']: hasClose,
+                        force_links_appearance: forceLinksAppearance,
```
amje, line 278: `force-links-appearance`

amje (changes requested):

src/components/Popover/Popover.tsx
ogonkov: ouch, fixed

ma-efremoff (approved):

src/components/Popover/Popover.tsx
```diff
@@ -136,6 +138,7 @@ export class Popover extends React.Component<
         hasClose: false,
         initialOpen: false,
         disabled: false,
+        forceLinksAppearance: true,
```
ma-efremoff, line 141: It is better to have names of properties when you don't need to set any values in defaultProps. What do you think about `disableLinksAppearance`?

ogonkov: it looks like "inverted" prop for me, like having `notOpen` instead of `open`

ma-efremoff: Inverted for "notOpen" is 'closed', your example with negation, and I agree that having negation in names is a bad pattern. But there is no negation in names like `disableAnyThing`, also there is standard property `disabled`.

ogonkov: `disableLinksAppearance` will cause same negotiation logic


```js
if (!disableLinksAppearance) {
  //add modifier
}
```

Versus current


```js
if (forceLinksAppearance) {
  //add modifier
```

ma-efremoff: Negation is bad as part of a name, but it's ok for conditions. More important to have intuitive set props.

amje (approved):

PR #282, opened 2022-07-28
chore(theme): export `Theme` type

ogonkov:


src/components/theme/index.ts
```diff
@@ -5,3 +5,4 @@ export * from './useTheme';
 export * from './useThemeValue';
 export * from './withTheme';
 export * from './withThemeValue';
+export {Theme as ThemeType} from './types';
```
amje, line 8: Why alias?

ogonkov:
```js
import {Theme} from '@yandex-cloud/uikit'
```

I think it looks like we import some component

amje: We already have similar exported types (i.e. `Lang`). No need to add *Type

ogonkov: Fixed

amje (approved):

PR #288, opened 2022-08-02
chore(Toaster): add close button label

ogonkov:


korvin89 (approved):

amje (approved):

src/components/utils/registerKeyset.ts
```diff
@@ -0,0 +1,10 @@
+import {i18n} from '../../i18n';
+import {Lang} from './configure';
+
+type KeysData = Parameters<typeof i18n.registerKeyset>[2];
+
+export function registerKeyset<T extends KeysData>(data: Record<Lang, T>, keysetName: string) {
+    Object.entries(data).forEach(([lang, keys]) => i18n.registerKeyset(lang, keysetName, keys));
+
+    return i18n.keyset(keysetName);
```
ValeraS, line 9: How about this?
```suggestion
    return i18n.keyset(keysetName) as (key: keyof T, params?: Params): string;
```

ogonkov: I think about adding it to `i18n` itself

PR #291, opened 2022-08-04
refactor: use `registerKeyset`

ogonkov:


src/components/Breadcrumbs/i18n/index.ts
```diff
@@ -1,11 +1,7 @@
-import {i18n} from '../../../i18n';
-import {Lang} from '../../utils/configure';
+import {registerKeyset} from '../../utils/registerKeyset';
 import en from './en.json';
 import ru from './ru.json';
 
 const COMPONENT = 'Breadcrumbs';
```
ValeraS, line 5: Let's use for ids a name like this `yc-some-component-name`.

ogonkov: Can you propose changes please?

ogonkov: Isn't it a breaking change if someone override it in project?

ValeraS: `i18n` will throw error if you try to use the same name for a keyset https://github.com/yandex-cloud/i18n/blob/4f791c56822dc9d1645cbe852738056412d5d0e6/src/index.ts#L26

vrozaev (approved):

ValeraS (approved):

amje (approved):

PR #292, opened 2022-08-04
refactor(Label): use `<Button/>`

ogonkov:


src/components/Label/Label.scss
```diff
@@ -54,8 +54,7 @@ $transitionTimingFunction: ease-in-out;
             }
 
             & #{$block}__icon {
-                width: 20px;
-                height: 20px;
```
ValeraS, line 58: These changes break a label with type `copy`.

amje (approved):

PR #293, opened 2022-08-04
refactor(TextInput): use `<Button/>`

ogonkov:


src/components/TextInput/i18n/en.json
```diff
@@ -0,0 +1,3 @@
+{
+    "label_clear-button": "Clear input value"
```
amje, line 2: Just `Clear` wont be enough?

ogonkov: I think it wont be clear enough what does it mean

amje (approved):

dustyo-O (approved):

PR #294, opened 2022-08-05
feat: added event broker to form components

Mishnya:
Added event listeners to Radio, RadioButton, RadioGroup, Checkbox and Switch.


src/components/utils/useCheckbox.ts
```diff
@@ -34,6 +35,15 @@ export function useCheckbox({
     }, [indeterminate]);
 
     const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+        eventBroker.publish({
+            componentId: 'Checkbox',
+            eventId: 'click',
```
amje, line 40: This event could be fired not only by user clicks, but also by external updates too. Maybe move it to onClick handler?

amje (approved):

PR #296, opened 2022-08-08
test(tabs): add tests

sofiushko:


src/components/Tabs/Tabs.tsx
```diff
@@ -43,7 +43,7 @@ export const Tabs: React.FC<TabsProps> = ({
     size = 'm',
     activeTab,
     allowNotSelected = false,
-    items = [],
+    items,
```
amje, line 46: Looks like a breaking change

src/components/Tabs/__tests__/Tabs.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+const tabTitle1 = 'Tab 1 title';
+const tab1 = {id: tabId1, title: tabTitle1};
+
+const tabId2 = 'tab2';
+const tabTitle2 = 'Tab 2 title';
+const tab2 = {id: tabId2, title: tabTitle2};
+
+const qaId = 'tabs';
+
+describe.only('Tabs', () => {
```
amje, line 16: only

src/components/Tabs/__tests__/TabsItem.test.tsx
```diff
@@ -0,0 +1,168 @@
+import React from 'react';
+import userEvent from '@testing-library/user-event';
+import {render, screen, fireEvent} from '@testing-library/react';
+import {TabsItem} from '../TabsItem';
+import {GearIcon} from '../../icons/GearIcon';
+
+const tabId = 'tab-id';
+const tabTitle = 'Tab title';
+
+describe.only('TabsItem', () => {
```
amje, line 10: only

amje (changes requested):

src/components/Tabs/__tests__/TabsItem.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+    });
+
+    test('should not call onClick on tab click', async () => {
+        const onClickFn = jest.fn();
+
+        render(<TabsItem id={tabId} title={tabTitle} onClick={onClickFn} />);
+
+        const component = screen.getByRole('tab');
+
+        fireEvent.keyDown(component, {keyCode: 9});
```
ogonkov, line 165: can we use `@testing-library/user-event`?

src/components/Tabs/__tests__/TabsItem.test.tsx
ogonkov: there is actually no need to wrap all tests in common `describe`, jest is adding test name anyway

```
PASS src/components/Tabs/__tests__/TabsItem.test.tsx (14.139 s)
  TabsItem
    ✓ should render tab item by default (29 ms)
```

src/components/Tabs/__tests__/TabsItem.test.tsx
sofiushko: done)

src/components/Tabs/Tabs.tsx
sofiushko: This prop is required by types https://github.com/yandex-cloud/uikit/blob/main/src/components/Tabs/Tabs.tsx#L32
Maybe we should make it optional then?

amje (approved):

ogonkov (approved):

PR #297, opened 2022-08-08
refactor(DropdownMenu): convert DropdownMenu to functional component

axtk:


src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -1,211 +1,195 @@
...
 } from './types';
 import {DropdownMenuPopup} from './DropdownMenuPopup';
 import {MenuProps} from '../Menu';
-
 import './DropdownMenu.scss';
 
 const b = block('dropdown-menu');
 
-interface DropdownMenuGeneralProps<T> {
+export type DropdownMenuProps<T> = {
```
ogonkov, line 27: Why types changed?

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -1,211 +1,195 @@
...
+                onMenuToggle?.();
+                hidePopup();
+            }
         }
 
-        this.props.onMenuToggle();
-
-        if (this.props.onSwitcherClick) {
-            this.props.onSwitcherClick(e);
+        function showPopup() {
```
ogonkov, line 107: It is better to use `useEffect` instead of memoization

```js
const onSwitcherClick = useCallback(() => {
  if (disabled) {
    return;
  }

  onMenuToggle();
  onSwitcherClick?.();

  setPopupShown((isPopupShown) => {
    return !isPopupShown;
  });
}, []);

useEffect(() => {
  if (!isPopupShown || !hideOnScroll) {
    return;
  }

  document.addEventListener('scroll', handleScroll, true);

  return () => {
    document.removeEventListener('scroll', handleScroll, true);
  };
}, [isPopupShown, hideOnScroll]);
```

src/components/DropdownMenu/DropdownMenu.tsx
axtk: the exported type hasn't really changed

before the changes, the exported type was split into a couple of types to represent the props and the state of the class component, now it's no longer necessary

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -1,211 +1,195 @@
...
-     * Useful when rendering lists with context menus.
+     * Switcher icon.
+     */
+    icon?: ReactNode;
+    onMenuToggle?: () => void;
+    hideOnScroll?: boolean;
+    /**
+     * Applied for the switcher and the menu.
+     */
+    size?: DropdownMenuSize;
```
ogonkov, line 42: May be we can move it after rest of the props, to reduce diff?

ogonkov (changes requested):
Readme broken

https://github.com/yandex-cloud/uikit/blob/8305d7f94c9640676bc083ed7e6687881664211f/src/components/DropdownMenu/README.md

src/components/DropdownMenu/DropdownMenu.tsx
axtk: looks better indeed, made an update, thanks

axtk:
> Readme broken
> https://github.com/yandex-cloud/uikit/blob/8305d7f94c9640676bc083ed7e6687881664211f/src/components/DropdownMenu/README.md

seems like prettier fails to properly handle the `|` symbol in the type declaration inside of a table and messes it up a little

src/components/DropdownMenu/DropdownMenu.tsx
axtk: not sure that the diff really matters here
seems better to have the most frequently used props closer to the top of the list

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -1,211 +1,171 @@
...
 
-    private showPopup() {
-        this.setState({popupIsShowed: true});
-
-        if (this.props.hideOnScroll) {
-            document.addEventListener('scroll', this.onScroll, true);
+    const handleMenuItemClick = useCallback(
+        (event: React.MouseEvent<HTMLElement, MouseEvent>, action: DropdownMenuItemAction<T>) => {
+            action(event, data);
+            onMenuToggle?.();
```
ogonkov, line 102: onMenuToggle is called when display state changes, i think it is ok to move it to separate `useEffect`

```js
useEffect(() => {
  onMenuToggle?.();
}, [isPopupShown]);
```

axtk: such an effect will also run right after the component has mounted, we'll have to handle it to prevent an unnecessary `onMenuToggle()` call; I think it's ok to leave it as it is

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -1,211 +1,171 @@
...
+    onSwitcherClick,
+    menuProps,
+    popupClassName,
+    popupPlacement,
+    children,
+}: DropdownMenuProps<T>) => {
+    const [isPopupShown, setPopupShown] = useState(false);
+    const anchorRef = useRef<HTMLDivElement | null>(null);
+
+    const handleSwitcherClick: MouseEventHandler<HTMLDivElement> = (event) => {
```
ogonkov, line 89: Can we memoize this callback with `useCallback`?

axtk: that would be an unnecessary overhead since this callback is used on a plain HTML element

axtk: a callback unwrapped into `useCallback()` won't cause a re-render by itself

the callback will indeed be recreated each re-render, but memoizing it would be pointless since plain HTML elements won't check whether their event handler props have changed before re-rendering

see also this thread:
https://github.com/facebook/react/issues/17055

ogonkov (approved):

PR #304, opened 2022-08-11
feat(HelpPopup)!: remove default ofset

chervyakovru:
Add css rule `left: 4` by default look like not so good solution.
What if i want render this Popover exactly in place? What if i want add padding with `padding` or `margin` rules?
And another important point is the evidence of behavior. How i should to know that this component is shifted on 4 pixels? 

Summing up the above i suggest remove this default offset at all and add the ability for users to manage the margins themselves

### DIFF
--- a/src/components/HelpPopover/HelpPopover.scss
+++ b/src/components/HelpPopover/HelpPopover.scss
@@ -1,7 +1,0 @@
-@use '../variables';
-
-$block: '.#{variables.$ns}help-popover';
-
-#{$block} {
-    position: relative;
-}
--- a/src/components/HelpPopover/HelpPopover.scss
+++ b/src/components/HelpPopover/HelpPopover.scss
@@ -4,5 +4,4 @@ $block: '.#{variables.$ns}help-popover';
 
 #{$block} {
     position: relative;
-    left: 4px;
 }
--- a/src/components/HelpPopover/HelpPopover.tsx
+++ b/src/components/HelpPopover/HelpPopover.tsx
@@ -1,9 +1,12 @@
 import React from 'react';
 
+import {block} from '../utils/cn';
 import {Popover, PopoverProps} from '../Popover';
 import {Icon} from '../Icon';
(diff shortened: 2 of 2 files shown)


ogonkov:
Looks like a breaking change

icekimi23:
Oh, I guess it would be breaking change:) why do you keep 4px in scss file?

src/components/HelpPopover/HelpPopover.scss
```diff
@@ -4,5 +4,4 @@ $block: '.#{variables.$ns}help-popover';
 
 #{$block} {
     position: relative;
-    left: 4px;
```
amje, line 7: No need of this file at all then

amje (changes requested):

src/components/HelpPopover/HelpPopover.tsx
```diff
@@ -18,7 +13,7 @@ export type HelpPopoverProps = DistributiveOmit<PopoverProps, 'children'>;
 
 export function HelpPopover(props: HelpPopoverProps) {
     return (
-        <Popover offset={{left: 4}} {...props} className={b(null, props.className)}>
```
ValeraS, line 21: Don't remove className, only remove import

ValeraS (changes requested):

src/components/HelpPopover/HelpPopover.tsx
ValeraS: I see it as a convention that the component should have its own class name.

chervyakovru: But we don't need any css rules for this component. why do we have add class with empty content?

chervyakovru: @ValeraS, can you take a look please?

chervyakovru: Returned the className for better readability of the components in the devTools

amje (approved):

ValeraS (approved):

PR #307, opened 2022-08-16
feat(LayerManager): add focus trapping

Volodomor:


src/components/Modal/Modal.tsx
```diff
@@ -89,6 +87,10 @@ export function Modal({
         onOutsideClick,
         onClose,
         contentRefs: [contentRef],
+        enableFocusTrap,
+        focusTrapContainersRefs: [contentRef],
+        focusTrapOptions,
```
ValeraS, line 92: I guess, these options should depend on `disableEscapeKeyDown`, `disableOutsideClick` props

src/components/utils/LayerManager.ts
```diff
@@ -18,23 +22,39 @@ export type ContentElement =
...
+}
+
+function isNonNullable<T>(value: T): value is NonNullable<T> {
+    return value !== null && value !== undefined;
 }
 
 class LayerManager {
     private stack: LayerConfig[] = [];
     private mouseDownTarget: HTMLElement | null = null;
+    private focusTraps = new WeakMap<LayerConfig, FocusTrap>();
```
ValeraS, line 39: Do we need to pause/unpause focus trap ourself? From [documentation](https://github.com/focus-trap/focus-trap#one-at-a-time):
> Focus trap manages a queue of traps: if A activates; then B activates, pausing A; then C activates, pausing B; when C then deactivates, B is unpaused; and when B then deactivates, A is unpaused.

src/components/Modal/Modal.tsx
```diff
@@ -53,6 +51,9 @@ export function Modal({
     'aria-label': ariaLabel,
     container,
     qa,
+    enableFocusTrap,
+    focusTrapOptions,
+    focusTrapRef,
```
ValeraS, line 56: I don't think we should give direct access to the `focus-trap` options, we need to use some abstraction here.

ValeraS, line 56: When do we need to control focus trap instance outside the component?

ValeraS (changes requested):

amje:
The feature was implemented in #482 and released in 4.0.0. Closing this PR

PR #326, opened 2022-09-06
docs(Link): impoving Link docs

LakeVostok:


src/demo/ShowcaseGrid/ShowcaseGrid.tsx
```diff
@@ -0,0 +1,123 @@
+import React, {useMemo} from 'react';
+import {Text} from '../../components/Text';
+import {getPropsCombinations, PropCombination, PropSequences} from './getPropsCombinations';
+
+type Combinations<ComponentType extends React.ElementType> = PropCombination<
+    Partial<React.ComponentProps<ComponentType>>
+>;
+
+export function ShowcaseGrid<ComponentType extends React.ElementType>({
```
LakeVostok, line 9: This is an attempt to reduce and standardize the rendering of components with different combinations of props. This is not very optimal, just something like how one would do.

src/components/Link/__stories__/Link.new.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+        title: {
+            control: {type: 'text'},
+            defaultValue: 'title',
+        },
+        id: {
+            control: {type: 'text'},
+        },
+    },
+    parameters: {
+        order: -100,
```
LakeVostok, line 41: Not sure if I should use it here, it is taken from Button

PR #329, opened 2022-09-07
fix: relax RealTheme type for compatibility

resure:
Current value of `RealTheme` type is not compatible with some existing code in services and related libraries. I think it would be better to keep it as `string` for now and move to a more narrow type in the next major release.

### DIFF
--- a/src/components/theme/types.ts
+++ b/src/components/theme/types.ts
@@ -1,4 +1,4 @@
-export type RealTheme = string;
+export type RealTheme = 'light' | 'light-hc' | 'dark' | 'dark-hc' | string;
 export type ThemeType = 'light' | 'dark';
 
 export type Theme = 'system' | RealTheme;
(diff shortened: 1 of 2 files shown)


amje (approved):

src/components/theme/types.ts
```diff
@@ -1,4 +1,4 @@
-export type RealTheme = 'light' | 'light-hc' | 'dark' | 'dark-hc';
+export type RealTheme = string;
 export type ThemeType = 'light' | 'dark';
 
 export type Theme = 'system' | RealTheme;
```
ValeraS, line 4:
```suggestion
export type RealTheme = 'light' | 'light-hc' | 'dark' | 'dark-hc';
export type ThemeType = 'light' | 'dark';

export type Theme = 'system' | RealTheme  | string;
```

resure: That's not enough, problem is in RealTheme type itself

PR #331, opened 2022-09-08
feat!: transfer package from yandex-cloud

amje:


package.json
```diff
@@ -34,6 +34,7 @@
         "prepublishOnly": "npm run build"
     },
     "dependencies": {
+        "@gravity-ui/i18n": "^1.0.0",
```
ogonkov, line 37: Why it is not a peer anymore?

amje: We are no more enforcing it. It's now only for internal use.

ogonkov (approved):

PR #341, opened 2022-09-09
chore: add title props for Text component

GermanVor:
Add `title` property for `Text` component.
`title` is [global attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes), there is no conflicts with `as` property.

### DIFF
--- a/src/components/Text/Text.tsx
+++ b/src/components/Text/Text.tsx
@@ -10,6 +10,7 @@ export interface TextProps extends TextBaseProps, ColorTextBaseProps {
     style?: React.CSSProperties;
     className?: string;
     children?: React.ReactNode;
+    title?: string;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/components/Text/Text.tsx
```diff
@@ -10,6 +10,7 @@ export interface TextProps extends TextBaseProps, ColorTextBaseProps {
     style?: React.CSSProperties;
     className?: string;
     children?: React.ReactNode;
+    title?: string;
```
teleginzhenya, line 13: I think it is better to name it `titleAttribute`, because in some components we use `title` property for text rendering

teleginzhenya, line 13: And we might want to add examples in storybook

GermanVor: [0ccd469](https://github.com/gravity-ui/uikit/pull/341/commits/0ccd4699fea51ffbdb4f2118da6f3467ff145b3d)

IsaevAlexandr (approved):

PR #351, opened 2022-10-06
fix(Label): change hover style for label with copy, fix label with single icon

mfyodorov:
- change hover style for label with copy: [Storybook](https://preview.yandexcloud.dev/uikit/351/?path=/story/components-label--interactions)
- fix label with a single icon: [Storybook](https://preview.yandexcloud.dev/uikit/351/?path=/story/components-label--icon)

### DIFF
--- a/src/components/Label/Label.scss
+++ b/src/components/Label/Label.scss
@@ -22,22 +22,29 @@ $transitionTimingFunction: ease-in-out;
     }
 
     &__icon {
-        position: absolute;
-        top: 0;
         display: flex;
         align-items: center;
         justify-content: center;
 
-        &_left {
+        &_side_left,
+        &_side_right {
+            position: absolute;
+            top: 0;
+        }
+
+        &_side_left {
             left: 2px;
         }
 
-        &_right {
+        &_side_right {
             right: 0;
         }
 
-        &_cross {
+        &_interactive {
+            --yc-button-background-color: transparent;
+
             cursor: pointer;
+            color: inherit;
             transition: color, background-color;
             transition-duration: $transitionDuration;
             transition-timing-function: $transitionTimingFunction;
@@ -72,7 +79,7 @@ $transitionTimingFunction: ease-in-out;
                     border-radius: 4px;
 
                     & #{$block}__icon {
-                        &_cross {
+                        &_interactive {
                             --yc-button-height: 20px;
                             --yc-button-border-radius: 4px;
                         }
@@ -107,7 +114,7 @@ $transitionTimingFunction: ease-in-out;
                     border-radius: 4px;
 
                     & #{$block}__icon {
-                        &_cross {
+                        &_interactive {
                             --yc-button-height: 28px;
                             --yc-button-border-radius: 4px;
                         }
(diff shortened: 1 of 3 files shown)


src/components/Label/Label.scss
```diff
@@ -22,21 +22,25 @@ $transitionTimingFunction: ease-in-out;
...
+        }
+
+        &_side_left {
             left: 2px;
         }
 
-        &_right {
+        &_side_right {
             right: 0;
         }
```
mfyodorov, line 41: like in the Button

amje (approved):

src/components/Label/Label.tsx
```diff
@@ -58,60 +65,62 @@ export const Label = React.forwardRef<HTMLDivElement, LabelProps>(function Label
...
 
-    const content = <div className={b('text')}>{children}</div>;
+    const content = hasContent && <div className={b('text')}>{children}</div>;
 
     const renderCopyButton = (status?: CopyToClipboardStatus) => {
         return (
-            typeCopy && (
-                <div className={b('icon', {right: true, copy: true})}>
+            <Button
+                onClick={onClose}
```
ValeraS, line 97: Why `onClose` here for copy button?

ValeraS (changes requested):

src/components/Label/Label.tsx
mfyodorov: this's typo, thanks, i'll fix it

src/components/Label/Label.tsx
```diff
@@ -58,14 +60,18 @@ export const Label = React.forwardRef<HTMLDivElement, LabelProps>(function Label
...
     } = props;
 
+    const hasContent =
+        children && Array.isArray(children) ? children.length > 0 : Boolean(children);
+
     const typeDefault = type === 'default';
-    const typeClose = type === 'close';
-    const typeCopy = type === 'copy';
+    const typeClose = type === 'close' && hasContent;
+    const typeCopy = type === 'copy' && hasContent;
```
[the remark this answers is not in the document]
mfyodorov: Exactly. There is some issue to implement that behaviour and it makes no sense as it seems to me.

src/components/Label/Label.tsx
```diff
@@ -151,16 +157,15 @@ export const Label = React.forwardRef<HTMLDivElement, LabelProps>(function Label
         event.stopPropagation();
     };
 
-    const labelWithCopy =
-        typeCopy && copyText && !disabled ? (
+    if (hasCopy && copyText) {
```
ValeraS, line 160:
```suggestion
    if (hasCopy) {
```

ValeraS (changes requested):

src/components/Label/Label.tsx
mfyodorov: I check copyText value because typescript highlights text prop of the CopyToClipboard below

ValeraS (approved):

src/components/Label/Label.tsx
mfyodorov: I thought about this double check but I'm think it's ok

PR #357, opened 2022-10-14
feat(Popover): support VirtualElement as anchor

makhnatkin:


src/components/Popover/types.ts
```diff
@@ -17,9 +18,11 @@ export type PopoverButtonProps = {
     onClick: (event: React.MouseEvent<HTMLButtonElement>) => void;
 };
 
+export type VirtualElement = popper.VirtualElement;
+
 export type PopoverExternalProps = {
     /** Allows to use custom anchor and disables `openByHover` and `onClick` for the component */
-    anchorRef?: React.RefObject<HTMLElement>;
+    anchorRef?: React.RefObject<HTMLElement | VirtualElement>;
```
amje, line 25: Let's refactor to `PopupProps['anchorRef']`

src/components/Popover/types.ts
```diff
@@ -17,9 +18,11 @@ export type PopoverButtonProps = {
     onClick: (event: React.MouseEvent<HTMLButtonElement>) => void;
 };
 
+export type VirtualElement = popper.VirtualElement;
```
amje, line 21: There is no need to export this

amje (changes requested):

src/components/Popover/types.ts
makhnatkin: ok

src/components/Popover/types.ts
makhnatkin: ok

PR #372, opened 2022-10-28
feat(Popover): updated anchorRef prop, added strategy prop to Popover

makhnatkin:
The Popover component calls Popup internally. Updated the Popover props type so that it can accept the strategy prop and pass it to the Popup. In addition, replaced the Anchor property type so that a virtual anchor could be set. Also, in order to be able to type the transmitted value, created an exported Anchor type

### DIFF
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -13,7 +13,6 @@ import {Trigger} from './components/Trigger/Trigger';
 import {useOpen} from './hooks/useOpen';
 import {cnPopover} from './Popover.classname';
 import type {PopoverInstanceProps, PopoverProps} from './types';
-import {PopperProps} from '../utils/usePopper';
 import {PopoverBehavior} from './config';
 import './Popover.scss';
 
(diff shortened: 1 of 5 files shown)


src/components/Popover/Popover.tsx
```diff
@@ -47,8 +48,9 @@ export const Popover = forwardRef(function (
         onCloseClick,
         onClick,
         anchorRef,
+        strategy,
         qa,
-    }: PopoverProps & QAProps,
+    }: Omit<PopperProps, 'offset' | 'modifiers'> & PopoverProps & QAProps,
```
makhnatkin, line 53: Can I also pass modifiers prop at the same time?

src/components/Popover/types.ts
```diff
@@ -107,7 +107,8 @@ export type PopoverDefaultProps = {
     size: 's' | 'l';
 };
 
-export type PopoverProps = PopoverExternalProps &
+export type PopoverProps = Omit<PopperProps, 'offset' | 'modifiers'> &
```
kseniya57, line 110: What if new properties are added to the `PopperProps`?  Typescript will think that `Popover` supports them, but, in fact, it won't.

makhnatkin: I agree with the remark, I will correct it. But in general Popover should support all Popup properties in my opinion

DarkGenius (approved):

amje (approved):

kseniya57 (approved):

PR #376, opened 2022-11-07
test(Table): add table tests

niktverd:


src/components/Table/utils/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+export interface DataItem {
+    name: string;
+    city?: string;
+    phone: string;
+    count: number;
+    date: string;
+    disabled?: boolean;
+}
+
+export const data: DataItem[] = [
```
amje, line 12: Why you moved the file? This should be placed inside `__stories__`, otherwise it will be compiled and packaged.

amje (changes requested):

amje (approved):

PR #380, opened 2022-11-11
feat(Label)!: new xs size and click action for more custom label

goshander:
I want to customize label for create new TagInput component with Label and TextInput collaboration

1. hasOnClick behaviour very strange only for default type
2. Fix Label component border radius as TextInput
3. Add correct XS, S, M sizes as TextInput component

### DIFF
--- a/src/components/Label/Label.scss
+++ b/src/components/Label/Label.scss
@@ -95,12 +95,12 @@ $transitionTimingFunction: ease-in-out;
             &#{$block}_style {
                 &_default,
                 &_rounded {
-                    border-radius: 4px;
+                    border-radius: var(--yc-border-radius-s);
 
                     & #{$block}__icon {
                         &_interactive {
                             --yc-button-height: 20px;
-                            --yc-button-border-radius: 4px;
+                            --yc-button-border-radius: var(--yc-border-radius-s);
                         }
                     }
                 }
@@ -130,12 +130,12 @@ $transitionTimingFunction: ease-in-out;
 
             &#{$block}_style {
                 &_default {
-                    border-radius: 4px;
+                    border-radius: var(--yc-border-radius-m);
 
                     & #{$block}__icon {
                         &_interactive {
                             --yc-button-height: 28px;
-                            --yc-button-border-radius: 4px;
+                            --yc-button-border-radius: var(--yc-border-radius-m);
                         }
                     }
                 }
--- a/src/components/Label/Label.scss
+++ b/src/components/Label/Label.scss
@@ -71,7 +71,7 @@ $transitionTimingFunction: ease-in-out;
     }
 
     &_size {
-        &_s {
+        &_xs {
             height: 20px;
 
             & #{$block}__text {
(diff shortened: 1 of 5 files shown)


goshander:
Template for new TagInput component
image

src/components/Label/Label.scss
```diff
@@ -95,12 +95,48 @@ $transitionTimingFunction: ease-in-out;
             &#{$block}_style {
                 &_default,
                 &_rounded {
-                    border-radius: 4px;
+                    border-radius: var(--yc-border-radius-s);
```
amje, line 98: For size `xs` used `border-radius-s`

src/components/Label/Label.tsx
```diff
@@ -63,7 +67,7 @@ export const Label = React.forwardRef<HTMLDivElement, LabelProps>(function Label
     const {
         type = 'default',
         theme = 'normal',
-        size = 's',
+        size = 'xs',
```
amje, line 70: This would be a breaking change

src/components/Label/Label.tsx
```diff
@@ -80,27 +84,33 @@ export const Label = React.forwardRef<HTMLDivElement, LabelProps>(function Label
...
+    const {copyIconSize, closeIconSize, buttonSize} = sizeMap[size];
 
     const leftIcon = icon && (
         <div className={b('icon', {side: hasContent ? 'left' : undefined})}>{icon}</div>
     );
     const content = hasContent && <div className={b('text')}>{children}</div>;
 
+    const handleCloseClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+        /* preventing event from bubbling */
+        event.stopPropagation();
```
amje, line 102: This could affect closing behaviour of Popups. Let's use another way

src/components/Label/Label.scss
```diff
@@ -211,5 +247,23 @@ $transitionTimingFunction: ease-in-out;
...
+
+            // hover on action button
+            &:not(#{$disabled}) #{$block}__icon_interactive {
+                --yc-button-background-color: transparent;
+                --yc-button-background-color-hover: var(--yc-color-text-secondary);
+
+                &:hover {
+                    color: var(--yc-color-text-inverted-primary);
+                }
+            }
```
mfyodorov, line 266: This code copy-pasted from `themeState` mixin with a tiny change, maybe there is a way to avoid it?

src/components/Label/Label.scss
```diff
@@ -71,7 +71,7 @@ $transitionTimingFunction: ease-in-out;
     }
 
     &_size {
-        &_s {
+        &_xs {
```
mfyodorov, line 74: This is a breaking change, don't forget about it

mfyodorov (approved):

src/components/Label/Label.scss
```diff
@@ -95,12 +95,48 @@ $transitionTimingFunction: ease-in-out;
...
+
+            &#{$block}_style {
+                &_default,
+                &_rounded {
+                    border-radius: var(--yc-border-radius-s);
+
+                    & #{$block}__icon {
+                        &_interactive {
+                            --yc-button-height: 24px;
+                            --yc-button-border-radius: var(--yc-border-radius-s);
```
amje, line 139: Maybe introduce a new `xs` size for button, and not use these variables here in this component?

amje (changes requested):

src/components/Label/Label.scss
goshander: Maybe work with button in other PR?

src/components/Label/Label.tsx
goshander: Add hasOnClick check to save old behaviour

src/components/Label/Label.scss
goshander: In design XS Label has same border radius 5px as S. I added another variable with equal value

Marginy605:
@goshander Looks like last label (see screenshot) doesn't depend on size:
telegram-cloud-photo-size-2-5273751362091468440-y
I suppose it should use this prop in showcase

goshander:
> @goshander Looks like last label (see screenshot) doesn't depend on size. I suppose it should use this prop in showcase

Thanks, fixed

amje (approved):

PR #384, opened 2022-11-22
feat: run tests with coverage and thresholds

helenjer:
This PR:
- Commits current reached thresholds: not to decrease coverage by code changes
- Runs test script on PR with coverage

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -13,6 +13,14 @@ module.exports = {
         '!**/__stories__/**/*',
         '!**/*/*.stories.{ts,tsx}',
     ],
+    coverageThreshold: {
+        global: {
+            branches: 45,
+            functions: 45,
+            lines: 65,
+            statements: 65,
+        },
+    },
     testEnvironment: 'jsdom',
     setupFiles: ['<rootDir>/test-utils/setup-tests.ts'],
     setupFilesAfterEnv: ['<rootDir>/test-utils/setup-tests-after.ts'],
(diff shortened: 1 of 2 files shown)


jest.config.js
```diff
@@ -13,6 +13,14 @@ module.exports = {
         '!**/__stories__/**/*',
         '!**/*/*.stories.{ts,tsx}',
     ],
+    coverageThreshold: {
```
amje, line 16: I'm not sure we need coverage tests on CI. These metrics are inconsistent.

helenjer: If we can automate it - why not?

amje: Because it can break thresholds even if you add code where tests are not applicable or not critical. Such CI fails will be a pain for a PR's author.

helenjer: > > It looks like we don't have culture to update/write new tests on PRs. It's up to help this practice.
> 
> Maybe it's better to integrate this aspect of culture in some sort of guidelines? And gently enforce it by maintainers during PR reviews — they should know what tests should really be added / updated in each PRs.

Are there any suitable guidelines for maintainers/contributors? The last one contains legal information.

amje: @helenjer We can add a new job for CI that will place comment with coverage changes in PRs using Github Action.

amje:
@helenjer Closing this. Reopen if it's still relevant

PR #386, opened 2022-11-27
feat(Select): add filter section

korvin89:


Marginy605:
Concerning new setting `filterOption`: it would be useful to have different options for filtering.
For example, we have list of options `['Value 1', 'val', 'Value', 'value']`:
1) default filtering (as it works) - ex. in case of searching `val` or `Val` for both we recieve full list with `['Value 1', 'val', 'Value', 'value']`
2) case sensitive filtering - ex.  in case of searching `val` we receive `['val', 'value']` and as for `Val` the result is `['Value 1', 'Value']`
3) precise search - ex.  in case of searching `val` we receive `['val']` and for `Val` there would be empty result.

The last option with precise search would be better to show in ui (as checkbox or etc) so that user could change filter optionally.

And one thing more, it would be useful to see this in storybook showcase with similar examples.

korvin89:
> Concerning new setting `filterOption`: it would be useful to have different options for filtering. For example, we have list of options `['Value 1', 'val', 'Value', 'value']`:
> 
> 1. default filtering (as it works) - ex. in case of searching `val` or `Val` for both we recieve full list with `['Value 1', 'val', 'Value', 'value']`
> 2. case sensitive filtering - ex.  in case of searching `val` we receive `['val', 'value']` and as for `Val` the result is `['Value 1', 'Value']`
> 3. precise search - ex.  in case of searching `val` we receive `['val']` and for `Val` there would be empty result.
> 
> We need to use last variant with precise search as optional select setting in ui for user (as checkbox inside select or etc). And I need to see this in storybook showcase.

O

amje (approved):

PR #387, opened 2022-11-27
feat: new component StoriesGroup

GermanVor:
Guys, I have questions about `Stories` and `StoriesGroup` components.

I can not understand the philosophy of the original component about fault tolerance.

The original `Stories` component lock the user screen if we pass empty items array - user will see unclosed modal.
Should we protect user in this case ? For example `return null`  And maybe there is point add also `onOutsideClick` and `onEscapeKeyDown`.

The original `Stories` component [check](https://github.com/gravity-ui/uikit/blob/main/src/components/Stories/Stories.tsx#L44) `initialStoryIndex` first time of render `Stories` and make it equal zero in mistake. But maybe better do it in `React.useEffect(() => { if (!isComponentDidMount) {...} }, [items])` too or do not even try handle others users mistakes and return in non valid cases `null` or throw Error ?

What is the `StoriesGroup` behaviour for such case ? 
```
groups = {[
    {items: [some_item_1, some_item_2]}, {items: []}, {items: [some_item_3]}
]}
```

And `Stories` do nothing with `storyIndex` when `React.useEffect(() => {}, [items])`, for example `items.lenght === 3`, `initialStoryIndex === 2` than `items.lenght === 2` and it locks forever the user screen.

Also this [line](https://github.com/gravity-ui/uikit/blob/main/src/components/Modal/Modal.scss#L17) seems extra like any other styles for modal like `border-radius`. In my opinion `Modal` as any other components should be as simple as it possible and make only functional work without non-layout styles.


GermanVor:
I am not finished work yet

GermanVor:
And I wonder about `block` cause every `className={b(string)}` will call concatenation function and it is extra load.
You can easy improve performance by moving static classNames outside the component like this 

```
const strCn = b(string)
...
const A = () => {
...
    className={strCn}
...
}
```

And you can also optimize memory and clean up component classNames when there is no a single instance in the DOM and generate them once when at least one instance still appears in the DOM.

DarkGenius:
> The original Stories component lock the user screen if we pass empty items array - user will see unclosed modal.
Should we protect user in this case ? For example return null

Good point, better to fix it here.

> And maybe there is point add also onOutsideClick and onEscapeKeyDown.

Existing event `onClose` contains reason for closing.

> But maybe better do it in React.useEffect(() => { if (!isComponentDidMount) {...} }, [items]) too or do not even try handle others users mistakes and return in non valid cases null or throw Error ?

I still don’t understand what is the advantage of this approach and what other errors are we talking about.

> What is the StoriesGroup behaviour for such case ?

You can ignore empty item sets.

> And Stories do nothing with storyIndex when React.useEffect(() => {}, [items]), for example items.lenght === 3, initialStoryIndex === 2 than items.lenght === 2 and it locks forever the user screen.

In a good way, we should prevent the user's screen lock.

> Also this [line](https://github.com/gravity-ui/uikit/blob/main/src/components/Modal/Modal.scss#L17) seems extra like any other styles for modal like border-radius. In my opinion Modal as any other components should be as simple as it possible and make only functional work without non-layout styles.

Auxiliary styles have been added to match the design of the component

GermanVor:
seems like we should ask user to memo `items` and `groups` to avoid undefined behaviours

DarkGenius:
> You can easy improve performance by moving static classNames outside the component like this

I think this optimization is practically negligible

DarkGenius:
Slide title is broken
Screen Shot 2022-11-29 at 6 52 00 PM

src/components/StoriesGroup/types.ts
```diff
@@ -0,0 +1,7 @@
+import {StoriesItem, StoriesItemMedia} from '../Stories/types';
+
+export interface StoriesGroupItem {
+    items: StoriesItem[];
+    /** Custom image for preview, otherwise first StoriesItem.media from StoriesItem will be used */
+    previewMedia?: StoriesItemMedia;
```
teleginzhenya, line 6: There is `selectedFromThumbnail` in https://github.com/gravity-ui/uikit/pull/387/files#diff-d7c93a4851676d09aa02706b71ceadeb04226b246e0050535331bcffae7d3812R23. I think it is better to name it `thumbnailMedia`

src/components/StoriesGroup/StoriesGroup.tsx
```diff
@@ -0,0 +1,124 @@
...
+    groups: StoriesGroupItem[];
+    initialStoryIndex?: [groupIndex: number, itemIndex: number];
+    disableOutsideClick?: boolean;
+    onClose?: (
+        event: MouseEvent | KeyboardEvent | React.MouseEvent<HTMLElement, MouseEvent>,
+        reason: ModalCloseReason | 'closeButtonClick',
+    ) => void;
+    onItemSelect?: (
+        itemIndexes: [groupIndex: number, itemIndex: number],
+        selectedFromThumbnail?: boolean,
```
teleginzhenya, line 23: Lets make it required argument

src/components/StoriesGroup/StoriesGroup.tsx
```diff
@@ -0,0 +1,124 @@
...
+    onItemSelect,
+    disableOutsideClick,
+    initialStoryIndex = [0, 0],
+    onClose,
+}: StoriesGroupProps) => {
+    const [[groupIndex, itemIndex], setCurrentStoryIndex] = React.useState(initialStoryIndex);
+
+    const handleClose = React.useCallback<NonNullable<StoriesGroupProps['onClose']>>(
+        (event, reason) => {
+            if (onClose) onClose(event, reason);
```
teleginzhenya, line 39: Let's keep it in one style, either `if (function) function(args)`, either `function?.(args)`

src/components/StoriesGroup/StoriesGroup.tsx
```diff
@@ -0,0 +1,124 @@
...
+            onItemSelect?.([newGroupIndex, 0], true);
+        },
+        [onItemSelect],
+    );
+
+    const currentGroup = groups[groupIndex];
+    const currentItems = currentGroup?.items || [];
+
+    const indexType =
+        ((currentGroup === undefined || currentItems[itemIndex] === undefined) &&
```
teleginzhenya, line 98: Same here, don't think we need `IndexType.Invalid`

src/components/StoriesGroup/components/StoriesPreview/StoriesPreview.scss
```diff
@@ -0,0 +1,71 @@
...
+#{$block} {
+    user-select: none;
+    margin-bottom: 16px;
+
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    gap: 8px;
+
+    &__slider-previe-list-wrapper {
```
teleginzhenya, line 14: `previe` typo

src/components/StoriesGroup/components/StoriesPreview/StoriesPreview.tsx
```diff
@@ -0,0 +1,132 @@
...
+};
+const StoriesPreviewWithSlider = ({
+    groupIndex,
+    groups,
+    onGroupSelect,
+}: StoriesPreviewWithSliderProps) => {
+    const [offset, setOffset] = React.useState(0);
+
+    React.useEffect(() => {
+        const currentOffset = 12 * Math.floor(groupIndex / 12);
```
teleginzhenya, line 58: 12 is [magic number](https://eslint.org/docs/latest/rules/no-magic-numbers) here, lets make constant with this value or even make it property

src/components/StoriesGroup/components/StoriesPreview/StoriesPreview.scss
```diff
@@ -0,0 +1,71 @@
...
+        width: 568px; // widht of 12 preview items
```
teleginzhenya, line 15: image
image

Shouldn't we keep in in the middle? I would prefer do not hardcode it here

src/components/StoriesGroup/components/StoriesPreview/StoriesPreview.tsx
```diff
@@ -0,0 +1,132 @@
...
+import './StoriesPreview.scss';
+import {Button} from '../../../Button';
+
+const b = block('stories-group-preview');
+
+type StoriesPreviewListProps = {
+    groups: StoriesGroupItem[];
+    groupIndex: number;
+    onGroupSelect: (groupIndex: number) => void;
+    offset?: number;
```
teleginzhenya, line 16: Should it be customized? I cannot imagine scenario when we need to use custom offset

src/components/Stories/Stories.scss
```diff
@@ -1,126 +1,9 @@
...
-        }
-    }
-
-    &__media-block {
-        position: relative;
-        display: flex;
-        width: 100%;
-        height: 100%;
+    &__modal-content {
+        border-radius: 20px;
```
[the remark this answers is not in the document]
GermanVor: now, we do not cause `StoriesLayout` is the an independent component that cannot affect the styles of the parent.

src/components/Stories/Stories.tsx
```diff
@@ -69,94 +51,50 @@ export function Stories({
...
             const newIndex = currentStoryIndex + 1;
-            setCurrentStoryIndex(newIndex);
             onNextClick?.(newIndex);
-        }
-    }, [currentStoryIndex, items, onNextClick]);
+            return newIndex;
+        });
+    }, [items, onNextClick]);
+
+    const indexType =
```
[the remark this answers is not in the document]
GermanVor: here cause of `StoriesGroup`

src/components/StoriesGroup/components/StoriesPreview/StoriesPreview.tsx
GermanVor: https://github.com/gravity-ui/uikit/blob/ca146d5d6acc826c00f08185a2340b48a4a4c5b2/src/components/StoriesGroup/components/StoriesPreview/StoriesPreview.tsx#L90

src/components/Stories/Stories.tsx
```diff
@@ -69,94 +51,50 @@ export function Stories({
...
+        (items[storyIndex] === undefined && IndexType.Invalid) ||
```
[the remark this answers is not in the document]
GermanVor: line 35 is for once checking

We can't guarantee that the index is always valid because we don't own the original array.
We can recalculate `storyIndex` if `items` will be memorized.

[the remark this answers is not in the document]
GermanVor: and also `Invalid` - it is literally one of the statuses for this case. 
`Invalid` - `(-infinity, start) and (end, +infinity)`
`Start` - `start`
`End` - `end`
`InProgress` - `(start, end)`

src/components/StoriesGroup/README.md
```diff
@@ -0,0 +1,65 @@
+## StoriesGroup
+
+Component for displaying stories. It looks like a carousel in a modal with given places to display text and media.
```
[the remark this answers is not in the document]
GermanVor: [ca146d5](https://github.com/gravity-ui/uikit/pull/387/commits/ca146d5d6acc826c00f08185a2340b48a4a4c5b2)

src/components/StoriesGroup/StoriesGroup.tsx
GermanVor: [37ebf42](https://github.com/gravity-ui/uikit/pull/387/commits/37ebf42123aef15bb3632269ddde6af48e838464)

src/components/StoriesGroup/components/StoriesPreview/StoriesPreview.scss
GermanVor: [3a02cb5](https://github.com/gravity-ui/uikit/pull/387/commits/3a02cb502aca456c09fd81b139198d5bf8ba13cd)

src/components/StoriesGroup/components/StoriesPreview/StoriesPreview.tsx
GermanVor: [eacd251](https://github.com/gravity-ui/uikit/pull/387/commits/eacd251693e128d07a54cafe132fef58e30d1f15)

src/components/StoriesGroup/StoriesGroup.tsx
teleginzhenya: Nope, I've meant itemIndexes and selectedFromThumbnail both required, not both optional

src/components/Stories/Stories.tsx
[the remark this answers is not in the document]
teleginzhenya: Yep, forgot to delete my comment, sorry

src/components/Stories/Stories.tsx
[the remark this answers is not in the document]
teleginzhenya: @DarkGenius what do you think about `Invalid` status? I would prefer to check it in line 35 in state initialization and in `useEffect` if `items` changes. Then `console.warn` that provided index is not in valid range and call `setCurrentStoryIndex` with [0, 0].

Everything else LGTM

teleginzhenya (changes requested):

src/components/Stories/Stories.tsx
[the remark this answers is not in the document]
GermanVor: We can not do `useEffect` for `items` until `items` is not memorised. 
And if users pass this property like this `items={[...]}` it will destroy front end cause of infinite recursion.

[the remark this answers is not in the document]
DarkGenius: > @DarkGenius what do you think about `Invalid` status? I would prefer to check it in line 35 in state initialization and in `useEffect` if `items` changes. Then `console.warn` that provided index is not in valid range and call `setCurrentStoryIndex` with [0, 0].
> 
> Everything else LGTM

I think status `Invalid` is overcomplicated. Let's reset the index to a reasonable default value if the value passed is invalid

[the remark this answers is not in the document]
teleginzhenya: @GermanVor yes, for better usage it should be memoized.
What do you mean by `And if users pass this property like this items={[...]} it will destroy front end cause of infinite recursion.`? If reference of items changes it will trigger `useEffect`, however if everything fine nothing would happen.

Another scenario how we could remove `Invalid` status is making `indexType` property optional, if it does not exists, then indexes are wrong and we won't render any buttons or render only close button.

teleginzhenya (approved):

teleginzhenya:
@DarkGenius @amje @ValeraS please take a look, because you're code owners

src/components/Stories/Stories.tsx
```diff
@@ -69,94 +51,61 @@ export function Stories({
...
+    if (items.length === 0) {
+        return null;
+    }
+
+    // case when items has changed and index has ceased to be valid
+    if (items[storyIndex] === undefined) {
+        const correctIndex = items[initialStoryIndex] === undefined ? 0 : initialStoryIndex;
+        setStoryIndex(correctIndex);
+
+        return null;
```
DarkGenius, line 86: Why does it return null here?

GermanVor: cause current `storyIndex` is invalid and we expect that `setStoryIndex(correctIndex);` will call render with valid index after `return null`.

GermanVor: it seems like recursion, but not really
cause of [this lines](https://github.com/gravity-ui/uikit/blob/d3e88f692c195db489ea79debbbf53c568c621ca/src/components/Stories/Stories.tsx#L78)

src/components/Stories/components/StoriesLayout/StoriesLayout.tsx
```diff
@@ -0,0 +1,98 @@
+import React from 'react';
+
+import {Button} from '../../../Button';
+import {ButtonClose} from '../../../Dialog/ButtonClose/ButtonClose';
+import {Link} from '../../../Link';
+import {MediaRenderer} from '..';
+import i18n from '../../i18n';
+
+import './StoriesLayout.scss';
```
DarkGenius, line 9: Let's move the styles import to the end

DarkGenius (changes requested):

src/components/StoriesGroup/StoriesGroup.tsx
```diff
@@ -0,0 +1,168 @@
...
+    ) => void;
+}
+
+export const StoriesGroup = ({
+    open,
+    groups,
+    onItemSelect,
+    disableOutsideClick,
+    initialStoryIndex = [0, 0],
+    maxSliderItemsCount = 12,
```
DarkGenius, line 34: Let's make it constant `DEFAULT_MAX_SLIDER_ITEMS_COUNT`

DarkGenius (changes requested):

src/components/StoriesGroup/StoriesGroup.tsx
GermanVor: I wouldn't do this because it seems that the memory area for this number will never be released on the client side.

DarkGenius: We don't skimp on constants, readability and clarity are a priority.

GermanVor: [89ef897](https://github.com/gravity-ui/uikit/pull/387/commits/89ef897cae4b92bca252fd2c0c29a14c304cb450)

DarkGenius (approved):

src/components/StoriesGroup/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './StoriesGroup';
+export * from './types';
```
amje, line 2: you should alse add reexport to main index.ts

amje (changes requested):

src/components/StoriesGroup/index.ts
GermanVor: [009fe86](https://github.com/gravity-ui/uikit/pull/387/commits/009fe86e3dcc26336bda65bc6b983645d8f1d9c1)

PR #393, opened 2022-12-01
fix(DropdownMenu): close opened menu on disable

ogonkov:


src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -135,6 +135,18 @@ export const DropdownMenu = <T,>({
...
     }, [isPopupShown, hideOnScroll, handleScroll]);
 
+    useEffect(() => {
+        if (disabled) {
+            setPopupShown((popupShown) => {
+                if (popupShown) {
+                    return false;
+                }
+
+                return popupShown;
```
axtk, line 145: возвращаемое значение функции внутри `setPopupShown` всегда `false`
достаточно `setPopupShown(false)`

проверку текущего значения `popupShown` стоит вынести в условие выше: `if (disabled && popupShown)`

axtk, line 145: возвращаемое значение функции внутри `setPopupShown` всегда `false`
достаточно `setPopupShown(false)`

проверку текущего значения `popupShown` стоит вынести в условие выше: `if (disabled && popupShown)`

axtk (changes requested):

src/components/DropdownMenu/DropdownMenu.tsx
ogonkov: popupShown check will require to add to useEffect dependencies, that what i wanted to avoid

axtk: > popupShown check will require to add it to useEffect dependencies

it will indeed, that shouldn't be a problem

in the current code, there's effectively no check for the `popupShown` value, since you always set it to `false` regardless of its current value

adding `popupShown` to the outer `if` and to the list of dependencies will make the check work

axtk: another option is to remove the `popupShown` value check from this effect altogether
react should bail out from a rerender triggered by a state setter with the same value

axtk (approved):

amje (approved):

PR #394, opened 2022-12-02
fix(DropdownMenu): trigger callback on each state change

ogonkov:


axtk (approved):

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -86,12 +87,20 @@ export const DropdownMenu = <T,>({
     const [isPopupShown, setPopupShown] = useState(false);
     const anchorRef = useRef<HTMLDivElement | null>(null);
 
+    const previousIsPopupShown = usePreviousValue(isPopupShown);
+    useEffect(() => {
```
ValeraS, line 91: it's breaking change

ValeraS (changes requested):

src/components/DropdownMenu/DropdownMenu.tsx
ogonkov: Why?

ValeraS: Thus we oblige the user to memoize `onMenuToggle` callback

ValeraS: IMHO, onMenuToggle should be called on open/close events, not in useEffect. It would be nice to read this article https://beta.reactjs.org/learn/you-might-not-need-an-effect#sharing-logic-between-event-handlers

ogonkov: In other words you propose to remember to call `onMenuToggle` each time state is changed?

ValeraS: not when state changes, but when toggle event happens).

ogonkov: Can you give some example?

ValeraS: for example, deleted code in this PR.

ogonkov: "Toggle event" happens 6 times, it is very error prone to remember to add onMenuToggle invocatio  each time

axtk: we can introduce a new callback that combines `setPopupShown` and `onMenuToggle` since they always occur together and use this callback instead of `setPopupShown(value)`

```ts
const togglePopup = useCallback<typeof setPopupShown>((value) => {
    onMenuToggle?.();
    setPopupShown(value);
}, [onMenuToggle]);
```

maybe not a big difference in this particular case, but this one is in fact more straightforward and more readable

ogonkov: i have replaced useEffect with custom hook

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -139,7 +136,7 @@ export const DropdownMenu = <T,>({
         if (disabled) {
```
ValeraS, line 136:
```suggestion
        if (disabled && isPopupShown) {
```

ValeraS (changes requested):

src/components/DropdownMenu/DropdownMenu.tsx
ogonkov: Why?

ValeraS: if disabled and not isPopupShown then no toggle event

ogonkov: useEffect was so much more simple

ValeraS (approved):

PR #395, opened 2022-12-04
fix: use native scrollbar on mobile devices

Avol-V:
On Android devices a custom scrollbar is not ignored, we need to disable it.

- MobileContext automatically adds the `yc-root_mobile` class.
- The existence of the `yc-root_mobile` or `yc-root_native-scrollbar` class disables the custom scrollbar.

### DIFF
--- a/src/components/theme/ThemeProvider.tsx
+++ b/src/components/theme/ThemeProvider.tsx
@@ -14,6 +14,7 @@ interface ThemeProviderDefaultProps {
     theme: Theme;
     systemLightTheme: RealTheme;
     systemDarkTheme: RealTheme;
+    nativeScrollbar: boolean;
 }
 
 export interface ThemeProviderProps
@@ -25,6 +26,7 @@ export function ThemeProvider({
     theme: themeProp = DEFAULT_THEME,
     systemLightTheme: systemLightThemeProp = DEFAULT_LIGHT_THEME,
     systemDarkTheme: systemDarkThemeProp = DEFAULT_DARK_THEME,
+    nativeScrollbar = false,
     children,
 }: ThemeProviderProps) {
     const [theme, setTheme] = useState<Theme>(themeProp);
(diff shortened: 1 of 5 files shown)


resure (approved):

amje (approved):

styles/styles.scss
```diff
@@ -104,7 +104,7 @@
     }
 }
 
-body.yc-root {
+body.yc-root:not(.yc-root_mobile):not(.yc-root_native-scrollbar) {
```
ValeraS, line 107: As a user, I need to create this class name (`.yc-root_native-scrollbar`) manually? How can I find out about this class name and what it does?

ValeraS (changes requested):

styles/styles.scss
Avol-V: From documentation. It's just an additional ability to force native scrollbar.
Automatically it's applied only on mobile devices.

ValeraS: Where is the documentation? Maybe export this class name?

Avol-V: Where I can place this export? Is there an example of such type of class name placement?

ValeraS (changes requested):

src/components/theme/ThemeProvider.tsx
```diff
@@ -47,8 +49,11 @@ export function ThemeProvider({
     const themeValue = theme === 'system' ? systemTheme : theme;
 
     useEffect(() => {
-        updateBodyClassName(themeValue);
-    }, [themeValue]);
+        updateBodyClassName(
+            themeValue,
+            nativeScrollbar === undefined ? undefined : {'native-scrollbar': nativeScrollbar},
```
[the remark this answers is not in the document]
Avol-V: The idea was to leave arguments the same as before if we did not use new prop. But OK, may be it is not necessary.

ValeraS (approved):

PR #399, opened 2022-12-06
feat(ShareTooltip): add Email as one of default sharing options and r…

niktverd:
…efactor sharetooltip


src/components/ShareTooltip/index.ts
```diff
@@ -1,4 +1,4 @@
-export {ShareSocialNetwork} from './constants';
+export {ShareOptions} from './constants';
```
Lunatic174, line 1: It looks like breaking change. To avoid it, it's better to reexport it with previous name too.
```ts
export {ShareOptions, ShareOptions as ShareSocialNetwork} from './constants';
```

niktverd: ok

amje (approved):

PR #406, opened 2022-12-06
fix(DropdownMenu.Item): allow usage w/out `text`

ogonkov:


axtk (changes requested):

src/components/DropdownMenu/DropdownMenuItem.tsx
```diff
@@ -1,11 +1,9 @@
...
+import React, {MouseEventHandler, useCallback, useContext} from 'react';
+import {Menu, MenuItemProps} from '../Menu';
 import {DropdownMenuContext} from './DropdownMenuContext';
-import type {DropdownMenuItem as DropdownMenuItemType} from './types';
+import type {DropdownMenuItemRequiredProps} from './types';
 
-type Props<T> = DropdownMenuItemType<T> & {
-    children?: ReactNode;
-};
+type Props<T> = Omit<MenuItemProps, 'onClick'> & Partial<DropdownMenuItemRequiredProps<T>>;
```
axtk, line 6: if the original `DropdownMenuItem` type changes in the way it picks props from `MenuItemProps` or some other way, we'll have to sync it with this type, which is inconvenient, if not a source of errors

why not do exactly what the title of the PR reads:

```ts
type Props<T> = Omit<DropdownMenuItem<T>, 'text'> & {
    text?: DropdownMenuItem<T>['text'];
    children?: ReactNode;
};
```

ogonkov: This is resolves `text` as required

axtk: it's strange, [tried it out](https://codesandbox.io/s/musing-platform-mj97b1?file=/src/index.ts), seems to work
can you show your code?

ogonkov: seems like i was confused by some WS checks

axtk (approved):

Issue #115, opened 2022-03-25
Weird ArrowToggle rotation


amje:
By default it follows CSS transition rules: angle is increasing - move clockwise, angle is decreasing - move counterclockwise. We should able to control this.

PR #419, opened 2022-12-13
fix(ArrowToggle): normal flip behavior #115

kiraind:


src/components/ArrowToggle/ArrowToggle.tsx
```diff
@@ -14,9 +14,56 @@ export interface ArrowToggleProps {
...
+        targetDirection,
+    );
+    const [transitions, setTransitions] = useState(true);
+
+    useEffect(() => {
+        if (direction === targetDirection) {
+            return;
+        }
+
+        if (direction === 'full-rotation') {
```
Marginy605, line 34: It's better to join this and previous condition with single return

kiraind: sure!

src/components/ArrowToggle/ArrowToggle.scss
```diff
@@ -7,6 +7,10 @@ $block: '.#{variables.$ns}arrow-toggle';
     vertical-align: middle;
     transition: transform 0.1s ease-out;
 
+    &_no-transition {
+        transition: transform 0s;
+    }
+
     &_direction {
```
ValeraS, line 14: IMHO, it can be done with css, for example 
```css
    &_direction {
        &_bottom {
            transform: matrix(1, 0, 0, 1, 0, 0);
        }

        &_left {
            transform: matrix(0, 1, -1, 0, 0, 0);
        }

        &_top {
            transform: matrix(-1, 0, 0, -1, 0, 0);
        }

        &_right {
            transform: matrix(0, -1, 1, 0, 0, 0);
        }
    }
```

ValeraS (changes requested):

src/components/ArrowToggle/ArrowToggle.scss
kiraind: that's genius, thank you

PR #426, opened 2022-12-19
feat(TextInput): ability to show innerLabel in InputControl

refactor256:
add ability to show inner label in InputControl variant of TextInput
inner label sample

### DIFF
--- a/src/components/TextInput/InputControl/InputControl.tsx
+++ b/src/components/TextInput/InputControl/InputControl.tsx
@@ -5,7 +5,7 @@ import {TextInputProps} from '../types';
 export interface InputControlProps
     extends Omit<TextInputProps, 'autoComplete'>,
         Pick<React.InputHTMLAttributes<HTMLInputElement>, 'autoComplete'> {
-    innerLabelWidth?: number;
+    labelWidth?: number;
 }
 
 const b = block('text-input');
@@ -19,8 +19,8 @@ export function InputControl(props: InputControlProps) {
         autoFocus,
         autoComplete,
         placeholder,
-        innerLabel,
-        innerLabelWidth,
+        label,
+        labelWidth,
         value,
         defaultValue,
         onChange,
@@ -35,8 +35,8 @@ export function InputControl(props: InputControlProps) {
     } = props;
 
     const style: React.CSSProperties = React.useMemo(
-        () => (innerLabel && innerLabelWidth ? {paddingLeft: `${innerLabelWidth}px`} : {}),
-        [innerLabel, innerLabelWidth],
+        () => (label && labelWidth ? {paddingLeft: `${labelWidth}px`} : {}),
+        [label, labelWidth],
     );
 
     return (
--- a/src/components/TextInput/InputControl/InputControl.tsx
+++ b/src/components/TextInput/InputControl/InputControl.tsx
@@ -4,7 +4,9 @@ import {TextInputProps} from '../types';
 
 export interface InputControlProps
     extends Omit<TextInputProps, 'autoComplete'>,
-        Pick<React.InputHTMLAttributes<HTMLInputElement>, 'autoComplete'> {}
+        Pick<React.InputHTMLAttributes<HTMLInputElement>, 'autoComplete'> {
+    innerLabelWidth?: number;
+}
 
 const b = block('text-input');
 
@@ -17,6 +19,8 @@ export function InputControl(props: InputControlProps) {
         autoFocus,
         autoComplete,
         placeholder,
+        innerLabel,
+        innerLabelWidth,
         value,
         defaultValue,
         onChange,
@@ -30,11 +34,17 @@ export function InputControl(props: InputControlProps) {
         disabled,
     } = props;
 
+    const style: React.CSSProperties = React.useMemo(
+        () => (innerLabel && innerLabelWidth ? {paddingLeft: `${innerLabelWidth}px`} : {}),
+        [innerLabel, innerLabelWidth],
+    );
+
     return (
         <input
             {...(controlProps as React.InputHTMLAttributes<HTMLInputElement>)}
             ref={controlRef as React.Ref<HTMLInputElement>}
             className={b('control', {type: 'input'}, controlProps?.className)}
+            style={style}
             type={type}
             name={name}
             id={id}
--- a/src/components/TextInput/InputControl/InputControl.tsx
+++ b/src/components/TextInput/InputControl/InputControl.tsx
@@ -4,9 +4,7 @@ import {TextInputProps} from '../types';
 
 export interface InputControlProps
     extends Omit<TextInputProps, 'autoComplete'>,
-        Pick<React.InputHTMLAttributes<HTMLInputElement>, 'autoComplete'> {
-    labelWidth?: number;
-}
+        Pick<React.InputHTMLAttributes<HTMLInputElement>, 'autoComplete'> {}
 
 const b = block('text-input');
 
@@ -19,8 +17,6 @@ export function InputControl(props: InputControlProps) {
         autoFocus,
         autoComplete,
         placeholder,
-        label,
-        labelWidth,
         value,
         defaultValue,
         onChange,
@@ -34,17 +30,11 @@ export function InputControl(props: InputControlProps) {
         disabled,
     } = props;
 
-    const style: React.CSSProperties = React.useMemo(
-        () => (label && labelWidth ? {paddingLeft: `${labelWidth}px`} : {}),
-        [label, labelWidth],
-    );
-
     return (
         <input
             {...(controlProps as React.InputHTMLAttributes<HTMLInputElement>)}
             ref={controlRef as React.Ref<HTMLInputElement>}
             className={b('control', {type: 'input'}, controlProps?.className)}
-            style={style}
             type={type}
             name={name}
             id={id}
--- a/src/components/TextInput/TextInput.scss
+++ b/src/components/TextInput/TextInput.scss
@@ -2,10 +2,12 @@
 @use '../../../styles/mixins';
 
 @mixin block_size($size: s) {
-    height: 24px;
-    padding: 4px 8px;
-    font-size: 13px;
-    line-height: 16px;
+    @if $size == 's' {
+        height: 24px;
+        padding: 4px 8px;
+        font-size: 13px;
+        line-height: 16px;
+    }
 
     @if $size == 'm' {
         height: 28px;
--- a/src/components/TextInput/TextInput.scss
+++ b/src/components/TextInput/TextInput.scss
@@ -84,6 +84,7 @@ $block: '.#{variables.$ns}text-input';
         text-overflow: ellipsis;
         white-space: nowrap;
         max-width: 50%;
+        z-index: 1;
     }
 
     &__error {
--- a/src/components/TextInput/TextInput.scss
+++ b/src/components/TextInput/TextInput.scss
@@ -50,6 +50,12 @@ $block: '.#{variables.$ns}text-input';
         }
     }
 
+    &__label {
+        position: absolute;
+        font-weight: 500;
+        box-sizing: border-box;
+    }
+
     &__error {
         @include mixins.text-body-1();
         color: var(--yc-color-text-danger);
@@ -84,6 +90,14 @@ $block: '.#{variables.$ns}text-input';
                 line-height: 16px;
             }
 
+            #{$block}__label {
+                height: 24px;
+                padding: 4px 8px;
+                padding-right: 4px;
+                font-size: 13px;
+                line-height: 16px;
+            }
+
             &#{$block}_has-clear {
                 #{$block}__control {
                     padding-right: 26px;
@@ -98,7 +112,15 @@ $block: '.#{variables.$ns}text-input';
                 height: 28px;
                 padding: 5px 8px;
                 font-size: 13px;
-                line-height: 16px;
+                line-height: 18px;
+            }
+
+            #{$block}__label {
+                height: 28px;
+                padding: 5px 8px;
+                padding-right: 4px;
+                font-size: 13px;
+                line-height: 18px;
             }
 
             &#{$block}_has-clear {
@@ -118,6 +140,14 @@ $block: '.#{variables.$ns}text-input';
                 line-height: 16px;
             }
 
+            #{$block}__label {
+                height: 36px;
+                padding: 10px 13px;
+                padding-right: 4px;
+                font-size: 13px;
+                line-height: 16px;
+            }
+
             &#{$block}_has-clear {
                 #{$block}__control {
                     padding-right: 36px;
@@ -131,6 +161,14 @@ $block: '.#{variables.$ns}text-input';
             #{$block}__control {
                 height: 44px;
                 padding: 12px 13px;
+                font-size: 15px;
+                line-height: 20px;
+            }
+
+            #{$block}__label {
+                height: 44px;
+                padding: 12px 13px;
+                padding-right: 4px;
                 font-size: 15px;
                 line-height: 20px;
             }
@@ -189,6 +227,10 @@ $block: '.#{variables.$ns}text-input';
             &:hover {
                 border-color: transparent;
             }
+        }
+
+        #{$block}__label {
+            color: var(--yc-color-text-hint);
         }
     }
 
--- a/src/components/TextInput/TextInput.scss
+++ b/src/components/TextInput/TextInput.scss
@@ -78,7 +78,7 @@ $block: '.#{variables.$ns}text-input';
 
     &__label {
         position: absolute;
-        font-weight: bold;
+        font-weight: 500;
         box-sizing: border-box;
         overflow: hidden;
         text-overflow: ellipsis;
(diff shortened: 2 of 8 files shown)


Marginy605:
@refactor256, let's make some improvements:
- add cases with innerLabel to storybook Showcases tab
- add availability for max width (px, %, etc) of innerLabel (if we have short container input would be longer it, see screenshort):
image
- hide long title by `overflow: hidden` + `text-overflow: ellipsis` (there is a bug, see screenshot):
image

refactor256:
@Marginy605 added commits with fixes + improvements + storybook showcases

src/components/TextInput/TextInput.tsx
```diff
@@ -144,10 +158,24 @@ export const TextInput = React.forwardRef<HTMLSpanElement, TextInputProps>(funct
             )}
             data-qa={qa}
         >
-            {props.multiline ? (
+            {isInnerLabelVisible && (
+                <span
+                    {...innerLabelProps}
+                    ref={innerLabelRef}
+                    className={b('label', innerLabelProps?.className)}
```
Marginy605, line 165: When you cut text with '...' you need to add title attribute on element for better ux

refactor256: Agree, will do.
Also I will wrap ResizeObserver in something like throttle/debounce, to minimize impact on performance.

Marginy605 (approved):

src/components/TextInput/InputControl/InputControl.tsx
```diff
@@ -17,6 +19,8 @@ export function InputControl(props: InputControlProps) {
         autoFocus,
         autoComplete,
         placeholder,
+        innerLabel,
+        innerLabelWidth,
```
amje, line 23: Can we do it automatically?

src/components/TextInput/TextInput.tsx
```diff
@@ -42,6 +43,7 @@ export const TextInput = React.forwardRef<HTMLSpanElement, TextInputProps>(funct
         name,
         value,
         defaultValue,
+        innerLabel,
```
amje, line 46: We call this prop in `<Select/>` just `label`. The behaviour should be also identical.

src/components/TextInput/TextInput.scss
```diff
@@ -129,10 +172,12 @@ $block: '.#{variables.$ns}text-input';
...
-                height: 44px;
-                padding: 12px 13px;
-                font-size: 15px;
-                line-height: 20px;
+                @include block_size(xl);
+            }
+
+            #{$block}__label {
+                @include block_size(xl);
+                padding-right: 4px;
```
amje, line 180: Can be lifted up for deduplication

src/components/TextInput/TextInput.scss
```diff
@@ -190,6 +235,10 @@ $block: '.#{variables.$ns}text-input';
                 border-color: transparent;
             }
         }
+
+        #{$block}__label {
+            color: var(--yc-color-text-hint);
```
amje, line 240: Shouldn't it be primary bold?

src/components/TextInput/TextInput.tsx
```diff
@@ -54,16 +56,22 @@ export const TextInput = React.forwardRef<HTMLSpanElement, TextInputProps>(funct
         style,
         className,
         qa,
+        innerLabelProps,
```
amje, line 59: Not sure if we want to allow any customization here

src/components/TextInput/TextInput.tsx
```diff
@@ -54,16 +56,22 @@ export const TextInput = React.forwardRef<HTMLSpanElement, TextInputProps>(funct
...
     } = props;
     const [uncontrolledValue, setUncontrolledValue] = React.useState(defaultValue ?? '');
     const innerControlRef = React.useRef<HTMLTextAreaElement | HTMLInputElement>(null);
+    const innerLabelRef = React.useRef<HTMLSpanElement>(null);
+    const [innerLabelWidth, setInnerLabelWidth] = React.useState(0);
     const [hasVerticalScrollbar, setHasVerticalScrollbar] = React.useState(false);
 
     const isControlled = value !== undefined;
     const inputValue = isControlled ? value : uncontrolledValue;
+    const isInnerLabelVisible = !multiline && Boolean(innerLabel);
```
amje, line 69: We can implement label via `text-indent: <number> px`. Thus we'll support both singleline and multiline inputs

src/components/TextInput/TextInput.tsx
```diff
@@ -144,10 +158,25 @@ export const TextInput = React.forwardRef<HTMLSpanElement, TextInputProps>(funct
...
         >
-            {props.multiline ? (
+            {isInnerLabelVisible && (
+                <span
+                    {...innerLabelProps}
+                    ref={innerLabelRef}
+                    className={b('label', innerLabelProps?.className)}
+                    title={innerLabel}
+                >
+                    {`${innerLabel}:`}
```
amje, line 168: `:` should not be included in label text. It is up to user

src/components/TextInput/TextInput.scss
```diff
@@ -50,6 +76,17 @@ $block: '.#{variables.$ns}text-input';
         }
     }
 
+    &__label {
+        position: absolute;
+        font-weight: 500;
```
amje, line 81: font-weight is configurable in branding. Use `@include mixins.text-accent`

src/components/utils/useElementSize.ts
```diff
@@ -0,0 +1,43 @@
...
+            return undefined;
+        }
+
+        const observer = new ResizeObserver(_throttle(handleResize, RESIZE_THROTTLE));
+        observer.observe(ref.current);
+
+        return () => {
+            observer.disconnect();
+        };
+    }, [ref, handleResize]);
```
amje, line 40: Hook won't know about ref update. HTMLElement will be better as the parameter.

amje (changes requested):

src/components/TextInput/TextInput.scss
refactor256: it's done like this to override `padding` property in mixins

src/components/TextInput/TextInput.scss
refactor256: it is for `disabled` state

src/components/TextInput/TextInput.tsx
refactor256: by default `max-width` is set at 50% parent container. but sometimes at use place we might want to set it in fixed `px` or different `%`.

src/components/TextInput/TextInput.tsx
refactor256: `text-indent` could be an option, but:
- label is not workable in multiline input, as on scroll the content covers label (see attach). I think at this stage Label makes sense only for singleline inputs

if we do not use label for multiline - does it make sense to use `text-indent`?

Screenshot 2022-12-28 at 13 02 55

src/components/TextInput/InputControl/InputControl.tsx
refactor256: yes, moved setting paddingLeft to parent component (TextInput). no need to add extra props here.

src/components/TextInput/TextInput.tsx
```diff
@@ -54,16 +56,23 @@ export const TextInput = React.forwardRef<HTMLSpanElement, TextInputProps>(funct
...
+    const [labelWidth, setLabelWidth] = React.useState(0);
     const [hasVerticalScrollbar, setHasVerticalScrollbar] = React.useState(false);
 
     const isControlled = value !== undefined;
     const inputValue = isControlled ? value : uncontrolledValue;
+    const isLabelVisible = !multiline && Boolean(label);
 
     const handleRef = useForkRef(props.controlRef, innerControlRef);
 
+    const innerControlSize = useElementSize(isLabelVisible ? innerControlRef : null);
```
ValeraS, line 74: Can we get label size here?

src/components/TextInput/TextInput.tsx
```diff
@@ -107,6 +122,17 @@ export const TextInput = React.forwardRef<HTMLSpanElement, TextInputProps>(funct
     const isErrorMsgVisible = typeof error === 'string';
     const isClearControlVisible = Boolean(hasClear && !disabled && inputValue);
 
+    const controlProps: TextInputProps['controlProps'] = React.useMemo(
```
ValeraS, line 125: `useMemo` not needed

src/components/utils/useElementSize.ts
refactor256: in this case, imo, the usage might become less clean.

right now it is hook's user responsibility to make sure he provides new ref, in case element in ref changes (e.g, `isLabelVisible ? innerControlRef : null` does exactly this)

also, i looked in https://www.npmjs.com/package/@react-hook/resize-observer and they use same approach, as i see.

ValeraS (changes requested):

src/components/TextInput/TextInput.tsx
refactor256: yes, looks like `labelSize` / `labelRef` also ok. will change.

src/components/TextInput/TextInput.scss
```diff
@@ -78,10 +110,12 @@ $block: '.#{variables.$ns}text-input';
...
             #{$block}__control {
-                height: 24px;
-                padding: 4px 8px;
-                font-size: 13px;
-                line-height: 16px;
+                @include block_size(s);
+            }
+
+            #{$block}__label {
+                @include block_size(s);
```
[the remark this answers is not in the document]
refactor256: @ValeraS 
- maybe lets rename mixin to `control_styles` and keep font-size in in?
- are you sure we want to use mixins for font-size? in initial `TextInput` they were hardcoded...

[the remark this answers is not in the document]
refactor256: @amje your comments are welcomed as well

src/components/TextInput/TextInput.scss
amje: ok

src/components/TextInput/TextInput.tsx
```diff
@@ -76,6 +91,12 @@ export const TextInput = React.forwardRef<HTMLSpanElement, TextInputProps>(funct
         }
     }, [multiline, inputValue, hasVerticalScrollbar]);
 
+    React.useLayoutEffect(() => {
+        const labelEl = labelRef.current;
+        const width = labelEl?.offsetWidth || 0;
+        setLabelWidth(width);
+    }, [labelSize, label, size, multiline]);
```
ValeraS, line 98: Can we get correct size in `useElementSize`? [ResizeObserverEntry](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry) may have [borderBoxSize](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry/borderBoxSize), but if it doesn't, it has a `target`, we can get `offsetWidth` from it.

src/components/TextInput/TextInput.tsx
```diff
@@ -42,28 +46,39 @@ export const TextInput = React.forwardRef<HTMLSpanElement, TextInputProps>(funct
...
+        controlProps: originalControlProps,
+        labelProps,
     } = props;
     const [uncontrolledValue, setUncontrolledValue] = React.useState(defaultValue ?? '');
     const innerControlRef = React.useRef<HTMLTextAreaElement | HTMLInputElement>(null);
+    const labelRef = React.useRef<HTMLLabelElement>(null);
+    const [labelWidth, setLabelWidth] = React.useState(0);
+    const [id] = React.useState(() =>
+        label ? originalId || _uniqueId(uniqueIdPrefix) : originalId,
+    );
```
ValeraS, line 71:
```suggestion
    const innerId = useUniqId();
    const id = originalId || innerId;
```

ValeraS (changes requested):

src/components/TextInput/TextInput.tsx
```diff
@@ -42,28 +44,37 @@ export const TextInput = React.forwardRef<HTMLSpanElement, TextInputProps>(funct
...
         className,
         qa,
+        controlProps: originalControlProps,
+        labelProps,
     } = props;
     const [uncontrolledValue, setUncontrolledValue] = React.useState(defaultValue ?? '');
     const innerControlRef = React.useRef<HTMLTextAreaElement | HTMLInputElement>(null);
+    const labelRef = React.useRef<HTMLLabelElement>(null);
+    const innerId = useUniqId();
+    const id = label ? originalId || innerId : originalId;
```
ValeraS, line 67: no need to check label, there is nothing wrong with always setting id to the input

src/components/utils/useElementSize.ts
```diff
@@ -0,0 +1,54 @@
...
+            return undefined;
+        }
+
+        const handleResize: ResizeObserverCallback = (entries) => {
+            if (!Array.isArray(entries)) {
+                return;
+            }
+
+            const entry = entries[0];
+            if (entry.borderBoxSize?.length > 0) {
```
ValeraS, line 31: from mdn [example](https://github.com/mdn/dom-examples/blob/main/resize-observer/resize-observer-text.html#L88), there might be an object here.

ValeraS (changes requested):

src/components/TextInput/TextInput.tsx
refactor256: ok, will remove (the idea was to avoid changing current behavior for inputs without explicit id provided, but it should be no problem, i guess)

src/components/TextInput/TextInput.scss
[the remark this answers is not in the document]
amje: I agreed that we should use mixins, but not `subheader`. 13/16 - `body-short`, 15/20 - `body-2`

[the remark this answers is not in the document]
ValeraS: Should the label have a `font-width` like for `subheader` or like for `body`?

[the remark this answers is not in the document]
refactor256: @ValeraS if you mean `font-weight` - for label anyway is used mixin `text-accent`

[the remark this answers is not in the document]
refactor256: @amje what about `m` size? it is now 13/18 for some reason (`font-size/line-height`). my idea would be to use 13/16 `body-short` for it, but change padding from `5px 8px` to `6px 8px`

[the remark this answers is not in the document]
refactor256: @amje I have added commit using these mixins

src/components/utils/useElementSize.ts
```diff
@@ -11,6 +11,10 @@ interface ElementSize {
 
 export function useElementSize<T extends HTMLElement = HTMLDivElement>(
     ref: MutableRefObject<T | null> | null,
+    // can be used, when it is needed to force reassign observer to element
+    // in order to get correct measures. might be related to below
+    // https://github.com/WICG/resize-observer/issues/65
```
refactor256, line 16: @ValeraS ResizeObserver not always firing (for example changing from 'm' to 'l' size it doesn't fire), so for such cases added `key` param (in case of `TextInput` size will be used as key)

src/components/TextInput/TextInput.scss
```diff
@@ -1,6 +1,34 @@
...
+        @include mixins.text-body-short;
+    }
+    @if $size == 'l' {
+        height: 36px;
+        padding: 10px 13px;
+        @include mixins.text-body-short;
+    }
+    @if $size == 'xl' {
+        height: 44px;
+        padding: 12px 13px;
```
amje, line 22: Side padding should be `12px` exactly like in `Select`

src/components/TextInput/TextInput.scss
```diff
@@ -1,6 +1,34 @@
...
+        @include mixins.text-body-short;
+    }
+    @if $size == 'm' {
+        height: 28px;
+        padding: 6px 8px;
+        @include mixins.text-body-short;
+    }
+    @if $size == 'l' {
+        height: 36px;
+        padding: 10px 13px;
```
amje, line 17: Side padding should be `12px` exactly like in `Select`

src/components/TextInput/TextInput.tsx
```diff
@@ -42,28 +44,37 @@ export const TextInput = React.forwardRef<HTMLSpanElement, TextInputProps>(funct
...
         onUpdate,
         onChange,
-        id,
+        id: originalId,
         tabIndex,
         style,
         className,
         qa,
+        controlProps: originalControlProps,
+        labelProps,
```
amje, line 61: Let's remove it for now. We'll be refactoring this later

amje (changes requested):

ValeraS (approved):

src/components/TextInput/TextInput.tsx
refactor256: you mean remove both lines? in this case if someone using `controlProps` to set `style` attribute of `InputControl`, and using label, then the style will be overriden.

amje: I meant labelProps only

refactor256: ok, removed it for now. but for sure we shall need some alternative to adjust styles (as minimum, `max-width`) of label, as it will be needed in actual use places. in order not to forget, maybe open an issue for this?
cc @Marginy605

amje (approved):

PR #430, opened 2022-12-19
feat: layout components

IsaevAlexandr:
About changes:

Briefly:
- introduces the concept of `spacing` and the ability to work with `media query` expressions;
- new `LayoutContext` provided props for base layout and other components;
- mobile-first grids with 12-th column layout;

More details:
- new css variables with `spacing`. Can be used and redefined at the project level;
- basic `media query` expressions are introduced. The values can be redefined in the scss file at the project level;
- new `LayoutProvider` component, where you can set default properties for new layout components;
- `useLayoutContext` hook to access the current active media expression and the calculated props from the topic for this expression;
- new `Container` component for centering content. Gets default props from `LayoutContext` depending on the current media screen value. Knows how to set the "rhythm" between the child components of `Row`;
- new `Row` component. Gets default props from `LayoutContext` depending on the current media screen value. It is a container for `Col` components;
- new `Col` component. Provides an opportunity to declaratively describe how many columns of a 12-column grid the component will occupy. If there are no props, then by default the component takes up all the available space in the row. Used only inside the `Row` component;
- new `Flex` component. Declarative description of the flexbox model in the jsx world. Retrieves active media expressions from contexts, so you can specify property values as functions of media expressions. It is also possible to set margins between child elements via the spacing prop (analogous to the `Stack` component);
- a new utility for generating classes to work with `spacing` indents. Motivation - no need to write a separate file with indents, you can use pre-generated classes;

warnings:
- negative margins for offsets between elements. You need to be careful with the `background-color` properties of such elements;


src/components/Layout/components/getThemeValue.ts
```diff
@@ -0,0 +1,31 @@
+import {LayoutProps, LayoutTheme, MediaPartial} from '../types';
+
+/**
+ * Select provided prop value from LayoutTheme. Depends on the current screen media query
```
resure, line 4: this (and others) should be a simple comment, not jsdoc

src/components/Layout/components/makeDataUiTheme.ts
```diff
@@ -0,0 +1,14 @@
+import merge from 'lodash/merge';
+import {DATA_UI_LAYOUT_THEME} from '../constants';
```
resure, line 2: "data-ui "is not a correct naming for external projects, it's more like "default" theme

resure:
@IsaevAlexandr this PR certainly lacks a description, at least a concise one

src/components/Layout/components/getThemeValue.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export function getThemeValue<T extends keyof LayoutProps>(
+    key: T,
+    theme: Partial<LayoutTheme>,
```
resure, line 8: Why layout is connected with theme at all? Theme should only affect colors, not other things.

resure (changes requested):

src/components/Layout/components/getThemeValue.ts
IsaevAlexandr: why?
all modern code editors support `jsdoc` in references.
You can easily write `md` syntax here like:
image

src/components/Layout/components/getThemeValue.ts
IsaevAlexandr: You can use `LayoutProvider` as a separate provider now. Current `ThemeProvider` only responsible for color app theme and should be called `ColorThemeProvider`.

But I think it will be inconvenient for the consumer to connect several providers to his application, because he will have to figure out what each of these providers is responsible for

src/components/Layout/components/useMediaQuery.tsx
```diff
@@ -0,0 +1,108 @@
...
+    removeEventListener: () => {},
+    dispatchEvent: (_: Event) => true,
+};
+
+const MEDIA_QUERIES: MediaProps<string> = {
+    mobile: `(max-width: ${MEDIA_TO_VALUE.mobile}px)`,
+    tablH: `(min-width: ${MEDIA_TO_VALUE.mobile + 1}px) and (max-width: ${MEDIA_TO_VALUE.tablH}px)`,
+    lptpS: `(min-width: ${MEDIA_TO_VALUE.tablH + 1}px) and (max-width: ${MEDIA_TO_VALUE.lptpS}px)`,
+    lptpM: `(min-width: ${MEDIA_TO_VALUE.lptpS + 1}px) and (max-width: ${MEDIA_TO_VALUE.lptpM}px)`,
+    dsktp: `(min-width: ${MEDIA_TO_VALUE.lptpM + 1}px) and (max-width: ${MEDIA_TO_VALUE.dsktp}px)`,
```
gorgeousvlad, line 21: is `dsktp` is really much shorter than `desktop`?  I think second option looks better, for tabl and lptp too

src/components/Layout/components/Container/Container.tsx
```diff
@@ -0,0 +1,109 @@
...
+ *       Col 1
+ *    </Col>
+ *    <Col>
+ *       Col 2
+ *    </Col>
+ *  </Row>
+ * </Container>
+ * ```
+ */
+export const Container = React.memo(
```
gorgeousvlad, line 66: did we considered of using styled components?

src/components/Layout/constants.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const MEDIA_TO_MOD: Record<MediaType, string> = {
+    mobile: 'm',
+    tablH: 'th',
+    lptpS: 'lps',
+    lptpM: 'plm',
+    dsktp: 'd',
+};
+export const MEDIA_TO_VALUE: Record<MediaType, number> = {
+    mobile: 480,
```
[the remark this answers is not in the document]
IsaevAlexandr: Good note. I'll add the ability to redefine these values through the theme

src/components/Layout/components/useMediaQuery.tsx
IsaevAlexandr: Are you talking about renaming these variables?
like:
`dsktp` -> `desktop`
`lptpM` -> `laptopM`
`tablH` -> `tablet`

gorgeousvlad: yep)

src/components/Layout/components/getThemeValue.ts
IsaevAlexandr: ok, l'll fix this. 
But i think project don't need this rules (@return statement in jsdoc for example) cos we have typescript in the project and use jsdoc only for annotations

src/components/Layout/components/spacing/spacing.tsx
```diff
@@ -0,0 +1,33 @@
+import {block} from '../../../utils/cn';
+import {Space} from '../../types';
+
+import './spacing.scss';
+
+const b = block('s');
+
+export interface SpacingProps {
```
Marginy605, line 8: Let's add comment for interface describing that props names are short versions of styles. Ex.`mr` is for `margin-top`, `pt` is for `padding-top`, plus description that `y` and `x` can be used for one direction spacing. It would be useful.

src/components/Layout/index.ts
```diff
@@ -0,0 +1,14 @@
...
+export * from './components/LayoutProvider';
+export * from './components/spacing/spacing';
+
+export * from './hooks/useLayoutContext';
+
+export * from './utils/makeDefaultTheme';
+
+export {SPACE_TO_PIXEL} from './constants';
+
+export type {Space, MediaType, LayoutTheme} from './types';
```
amje, line 14: Let's reduce the number of exported members. At the beginning we don't need the most, and they might be subjects of change. I would keep just components (Col, Row, Flex, Container, LayoutProvider) for now

src/components/Layout/types.ts
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * Override default breakpoints values.
+     *
+     * @important **you mast override corresponding scss variables**
+     */
+    breakpoints: MediaProps<number>;
+    /**
+     * Component props build in into the theme. You can describe the props depending on the current media expression
+     */
+    smartProps: SmartProps;
```
amje, line 47: Maybe call it `config`?

src/components/Layout/variables.scss
```diff
@@ -0,0 +1,78 @@
...
+$SPACE: (
+    none: var(--yc-spacing-none),
+    nano: var(--yc-spacing-nano),
+    micro: var(--yc-spacing-micro),
+    xs: var(--yc-spacing-xs),
+    s: var(--yc-spacing-s),
+    m: var(--yc-spacing-m),
+    l: var(--yc-spacing-l),
+    xl: var(--yc-spacing-xl),
+    xxl: var(--yc-spacing-xxl),
```
amje, line 40: Can't we use namespace (`$ns`) here?

src/components/Layout/variables.scss
```diff
@@ -0,0 +1,78 @@
+@use '../variables.scss';
+
+$colClassName: '.#{variables.$ns}col';
+$rowClassName: '.#{variables.$ns}row';
+$flexClassName: '.#{variables.$ns}flex';
+$containerClassName: '.#{variables.$ns}container';
+$spacingClassName: '.#{variables.$ns}s';
```
amje, line 7: Let's use block instead of className

amje (changes requested):

src/components/Layout/types.ts
```diff
@@ -0,0 +1,52 @@
+export type ColSize = '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9' | '10' | '11' | '12';
```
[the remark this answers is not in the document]
IsaevAlexandr: like `1 | 2 | 3 | ... | '1' | '2' | '3' | ...`?

src/components/Layout/components/Container/Container.scss
```diff
@@ -0,0 +1,22 @@
+@use '../../variables.scss' as v;
+
+#{v.$containerClassName} {
+    margin-inline: auto;
+    max-width: 100%;
+    height: 100%;
+
+    @each $media, $breakpoint-width in v.$MEDIA_TO_BREAKPOINT_WIDTH {
+        &_mw_#{$media} {
```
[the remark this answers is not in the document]
IsaevAlexandr: `mw` is shot for `max-width`. A can add comment here

src/components/Layout/components/Container/Container.scss
```diff
@@ -0,0 +1,22 @@
...
+            max-width: $breakpoint-width;
+        }
+    }
+
+    @each $space, $value in v.$SPACE {
+        // space row - space between Row components
+        &_sr_#{$space} {
```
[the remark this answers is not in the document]
IsaevAlexandr: we don't need long class names here cos it's a lib code. For prevent bundle size grow better to holt short class names

src/components/Layout/components/Container/Container.tsx
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * Remove gutters
+     */
+    disableGutters?: boolean;
+    /**
+     * Space between child `Row` components
+     *
+     * By default takes props via `LayoutContext`
+     */
+    spaceRow?: ((fn: IsMediaActive) => Space) | Space;
```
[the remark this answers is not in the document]
IsaevAlexandr: cos it's for spacing in vertical direction. More explicit usage

src/components/Layout/components/Container/Container.tsx
```diff
@@ -0,0 +1,109 @@
...
+    className?: (fn: IsMediaActive) => string | string;
+    children?: React.ReactNode;
+    /**
+     * Max width equals max width of current breakpoint
+     */
+    fixed?: boolean;
+    /**
+     * Width of container will never be larger then specified media type width
+     */
+    maxWidth?: MediaType;
```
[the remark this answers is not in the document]
IsaevAlexandr: I didn't quite understand what needs to be done here.
Current usage is:
```tsx
<Container maxWIdth="m">...</Container>
```

src/components/Layout/components/Container/Container.tsx
```diff
@@ -0,0 +1,109 @@
...
+import './Container.scss';
+
+const b = block('container');
+
+export interface ContainerProps {
+    style?: React.CSSProperties;
+    /**
+     * Use function to define different classes in different media queries
+     */
+    className?: (fn: IsMediaActive) => string | string;
```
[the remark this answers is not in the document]
IsaevAlexandr: it's rly looks strange here. 
let's clean it up. If needed in the future, you can always add

src/components/Layout/components/Col/Col.tsx
```diff
@@ -0,0 +1,81 @@
...
+ */
+export const Col = ({children, style, className, ...media}: ColProps) => {
+    const mods = React.useMemo(
+        () =>
+            Object.entries(media).reduce((acc, [mod, modSize]) => {
+                acc[`s-${mod}`] = modSize;
+
+                return acc;
+            }, {} as Record<string, ColSize>),
+        [media],
```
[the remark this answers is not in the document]
IsaevAlexandr: yes it's true. I'll remove this `React.useMemo` and wrap component in `React.memo`

src/components/Layout/components/Row/Row.tsx
```diff
@@ -0,0 +1,68 @@
...
+ *  <Col>col</Col>
+ *  <Col>col</Col>
+ * </Row>
+ * ```
+ */
+export const Row = ({children, style, className, space, spaceRow}: RowProps) => {
+    const {isMediaActive, rowThemeProps} = useRowThemeProps();
+
+    return (
+        <div
```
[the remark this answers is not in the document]
IsaevAlexandr: It's can affect unexpected side effects in future. If it possible we don't need to use this component here

src/components/Layout/components/Row/Row.tsx
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * Space between children `<Col />` components. Gaps are saved if children go to another line
+     *
+     * If not specified takes props via `LayoutContext`
+     */
+    space?: Space | ((fn: IsMediaActive) => Space | undefined);
+    /**
+     * Override default (space) vertical gaps between children if it wrap on next line
+     */
+    spaceRow?: Space | ((fn: IsMediaActive) => Space | undefined);
```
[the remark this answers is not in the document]
IsaevAlexandr: i think what `space` is more abstract name for this feature and can rely on different realisation inside. Instead `gap` can confuse some developers and may affect bugs in future

src/components/Layout/components/Col/Col.tsx
```diff
@@ -0,0 +1,81 @@
...
+ * ```tsx
+ * // wrong
+ * <Col>
+ *      <SomeComponentWIthBackground />
+ * </Col>
+ *
+ * // right
+ * <Col>
+ *   <div>
+ *     <SomeComponentWIthBackground />
```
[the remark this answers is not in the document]
IsaevAlexandr: i'll wrong here. This comment belongs only to `Flex` componet. 
Limitation there is only than we try to set `background-color` to Col component.

IsaevAlexandr:
/ping

any updates here?

IsaevAlexandr:
continue in https://github.com/gravity-ui/uikit/pull/551

PR #439, opened 2022-12-22
feat(Toaster): add "has" method

oynikishin:


src/components/Toaster/Toast/Toast.tsx
```diff
@@ -202,6 +204,17 @@ export function Toast(props: ToastUnitedProps) {
         [type || 'default']: true,
     };
 
+    React.useEffect(() => {
+        if (onMount && heightProps.ref.current) {
+            onMount({toast: props, element: heightProps.ref.current});
```
amje, line 209: I would call it `props` instead of `toast`

src/components/Toaster/types.ts
```diff
@@ -18,6 +18,13 @@ export type ToastAction = {
     removeAfterClick?: boolean;
 };
 
+export interface ToastLifecircleArgs {
```
amje, line 21: typo: `lifecycle`

amje (changes requested):

src/components/Toaster/README.md
```diff
@@ -90,17 +90,19 @@ On initialization it is possible to pass className, that will be assigned to dom
...
+| content         | `node`                   |          | `undefined` | Notification content. [Anything that can be rendered: numbers, strings, elements or an array](https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes)  |
+| type            | `string`                 |          | `undefined` | Notification type. Possible values: `error`, `success`. If `type` is set, icon (success/error) will be added into notification title. _By default there is no icon_ |
+| isClosable      | `boolean`                |          | `true`      | Configuration that manages visibility of cross icon, which allows to close notification                                                                             |
+| actions         | `ToastAction[]`          |          | `undefined` | Array of [actions](./types.ts#L9) which displays after `content`                                                                                                    |
+| onMount         | `ToastLifecycleCallback` |          | `undefined` | Callback. Fired when corresponding toast component mount                                                                                                            |
```
korvin89, line 104: I think `callback` word is redundant

src/components/Toaster/Toast/Toast.tsx
```diff
@@ -202,6 +204,17 @@ export function Toast(props: ToastUnitedProps) {
...
+    React.useEffect(() => {
+        if (onMount && heightProps.ref.current) {
+            onMount({props, element: heightProps.ref.current});
+        }
+        return () => {
+            if (onUnmount && heightProps.ref.current) {
+                onUnmount({props, element: heightProps.ref.current});
+            }
+        };
+    }, []);
```
korvin89, line 216: Why there are no `onMount`, `onUnmount` dependencies here?

oynikishin: I don't want to do smth when `onMount` or `onUnmount` changes.
`onMount` and `onUnmount` should fired only once during corresponding stages.
Using  `React.useEffect` without debs is common pattern for this case as I know.
Actual value of these functions will be used when time comes.

amje (approved):

src/components/Toaster/Toast/Toast.tsx
korvin89: Ok. I'll leave it up to Alexey @ogonkov :)

ValeraS (changes requested):
What task do we solve here? We need callbacks for toast addition and deletion events, right? If so, it is better to call them from appropriate methods from `ToasterProvider` (add, remove, removeAll).

oynikishin:
The main goal is to provide API that helps developers to understand when they cannot reuse (update) old toast and have to create another one.

I chose to provide unmount event because it is easy to understand API and solves my problem not ideal but well enough.
(mount is added for symmetry).

The problem with `remove` is that I don't quite understand what it is.
Technically a toast is removed when it is deleted from the list of toasts.
But in fact a toast becomes "invalid" when it enters the `ToastStatus.Hiding` stage.
We can still change the content but we can no longer affect `allowAutoHiding` and `timeout` props.
And maybe it would be better for the developer to recreate the toast and not try to update the current hiding one.

Maybe it will be better to provide a `ToastStatus` change event with extra status like `removed`..?
I need your advice!

ogonkov:
Can you provide example of usage of this api?

oynikishin:
> Can you provide example of usage of this api?

I need smith like this:
`
    const createUpdatableToast = (name: string, title: string) => {
        let toastExists = false;
        const addToast = (newTitle: string) => {
            toaster.add({
                name,
                title: newTitle,
                onUnmount: () => {
                    toastExists = false;
                }
            });
            toastExists = true;
        };

        const addOrUpdateToast = (newTitle: string) => {
            if (toastExists) {
                toaster.update(name, { title: newTitle});
            } else {
                addToast(newTitle);
            }
        };

        addToast(title);
        return addOrUpdateToast;
    }
`
There is several ways to solve my problem:
 - different variants of `close callback` (at least three))
 - or maybe some method like addOrUpdate (...why not)
 
 As I already mentioned.
 I chose mount/unmount event because it is enough for my case and it is well self-documented api.
 
 But may be you suggest the best way!

src/components/Toaster/__tests__/ToasterProvider.test.tsx
```diff
@@ -250,3 +250,54 @@ describe('api.removeAll', () => {
         expect(screen.queryByRole('heading', {name: 'Test Toast2'})).not.toBeInTheDocument();
     });
 });
+
+describe('api.has', () => {
+    it('should return false when toast is not added', () => {
+        const providerAPI = setup();
+        expect(providerAPI.has('unexisted toasts')).toBeFalsy();
```
ogonkov, line 257: toBe(false)

you interested in returning boolean from api, not some falsy value

oynikishin: Fixed!

ogonkov (approved):

src/components/Toaster/Provider/ToasterProvider.tsx
```diff
@@ -60,20 +60,28 @@ export const ToasterProvider = React.forwardRef<ToasterPublicMethods, Props>(
             [],
         );
 
+        const toastsRef = React.useRef<InternalToastProps[]>();
+        toastsRef.current = toasts;
```
ValeraS, line 64: React documentation recommends not writing or reading `ref.current` during rendering https://beta.reactjs.org/reference/react/useRef#caveats

ValeraS (changes requested):

ValeraS (approved):

PR #440, opened 2022-12-22
feat(Toaster): add ability to customize icon

oynikishin:


src/components/Toaster/types.ts
```diff
@@ -29,6 +31,9 @@ export type ToastProps = {
     isClosable?: boolean;
     isOverride?: boolean;
     actions?: ToastAction[];
+
+    /** Function. Use for toast icon customization. By default type-based behavior is used */
+    iconRenderer?: ToastIconRenderer;
```
amje, line 36: `renderIcon`

src/components/Toaster/Toast/Toast.scss
```diff
@@ -153,6 +140,14 @@ $block: '.#{variables.$ns}toast';
         }
     }
 
+    &__icon-container {
+        flex: 0 0 auto;
+
+        &_not-empty {
```
amje, line 146: Can be implemented without mod: `&:not(:empty) {padding-right: 8px}`

amje (changes requested):

src/components/Toaster/Toast/Toast.scss
oynikishin: Yes I can.
But complex selector especially with pseudo classes like 'not' are a little bit more expensive.

However, if you want, I will rewrite this place.

src/components/Toaster/README.md
```diff
@@ -90,17 +90,18 @@ On initialization it is possible to pass className, that will be assigned to dom
...
+| content         | `node`               |          | `undefined` | Notification content. [Anything that can be rendered: numbers, strings, elements or an array](https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes)  |
+| type            | `string`             |          | `undefined` | Notification type. Possible values: `error`, `success`. If `type` is set, icon (success/error) will be added into notification title. _By default there is no icon_ |
+| isClosable      | `boolean`            |          | `true`      | Configuration that manages visibility of cross icon, which allows to close notification                                                                             |
+| actions         | `ToastAction[]`      |          | `undefined` | Array of [actions](./types.ts#L9) which displays after `content`                                                                                                    |
+| renderIcon      | `RenderIconFunction` |          | `undefined` | Function. Use for toast icon customization. By default type-based behavior is used                                                                                  |
```
korvin89, line 104: I think `Function` is redundant

amje (changes requested):

src/components/Toaster/Toast/Toast.scss
```diff
@@ -14,20 +14,26 @@ $block: '.#{variables.$ns}toast';
...
-    --yc-toaster-padding: 16px 16px 16px 48px;
-
+    display: flex;
+    flex-flow: row nowrap;
     box-sizing: border-box;
     position: absolute;
+    opacity: 0;
     width: inherit;
     margin-bottom: 10px;
-    padding: var(--yc-toaster-padding);
```
ogonkov, line 23: Why is var removed?

src/components/Toaster/Toast/Toast.tsx
```diff
@@ -209,8 +210,10 @@ export function Toast(props: ToastUnitedProps) {
             {...heightProps}
             {...closeOnTimeoutProps}
         >
+            <div className={b('icon-container')}>
+                {renderIcon ? renderIcon(props) : renderIconByType({type})}
+            </div>
```
ogonkov, line 215: Why icon wrapped to div and moved out of the container?

src/components/Toaster/Toast/Toast.scss
oynikishin: In the old layout, the size of the icon was implicitly declared in the padding and this var was redefined for the default toast (without icon).
Now icon has own container and padding is always `16px`.

src/components/Toaster/Toast/Toast.tsx
oynikishin: In the old layout, the size of the icon was implicitly declared in the padding.
It means - we have to provide icons with the same size.
When I discussed the possibility of customizing icons with @korvin89 He advised me to also get away from the predefined size.
So I use flex container and` icon-container` should stretch to the size of the passed icon.

src/components/Toaster/types.ts
```diff
@@ -18,6 +18,8 @@ export type ToastAction = {
     removeAfterClick?: boolean;
 };
 
+export type RenderIconFunction = (toastProps: ToastProps) => React.ReactNode;
```
amje, line 21: Let's remove this type completely, and use function notation: `() => React.ReactNode`

amje (changes requested):

src/components/Toaster/types.ts
oynikishin: Do you really want me to remove the arguments?
I added them to give the user the ability to naturally change the icon depending on the type and (possibly) other properties of the toast.
We don't have a toast object.
I doubt a closure will work here, since all the user has is an immutable object that he pass into the add function.
When changing the properties of the toast through the change method, the old closure will no longer be relevant. He will have to change the render function too.

If you don't like the argument, maybe it will be better to have just Icon ReactNode prop (not a function)?

amje (approved):

src/components/Toaster/Toast/Toast.scss
```diff
@@ -49,24 +54,10 @@ $block: '.#{variables.$ns}toast';
...
     }
 
-    &_show-animation#{&}_default {
-        $indents-duration: 0.35s;
-        $move-duration: 0.25s;
-
-        animation: toast-set-symmetrical-indents $indents-duration ease forwards,
-            toast-move-left $move-duration ease forwards $indents-duration,
-            get-trailing-keyframe(toast-display-end, $indents-duration + $move-duration);
-    }
```
ogonkov, line 59: Why is animation changed in PR about customizing icon?

oynikishin: In the previous version, the size of the icon was taken into account in the paddings. Because of this, it was necessary to handle the default toast type separately using `toast-set-symmetrical-indents`  keyframes. Now paddings always symmetrical.
I just remove duplicate animation rule.

ogonkov (approved):

PR #441, opened 2022-12-22
fix(Toaster): fix animation when toast is updated

oynikishin:


src/components/Toaster/Toast/Toast.tsx
```diff
@@ -37,39 +37,42 @@ enum ToastStatus {
...
+            status === ToastStatus.Hiding
+        ) {
+            return; // Not update height during animation
+        }
 
-        setHeight(height);
-    }, [isOverride]);
+        const newHeight = getHeight(ref);
+        setHeight(newHeight);
+    }, [updatesCounter]);
```
ogonkov, line 75: You want to update height on sequential updates?

oynikishin: Yes. This calculated height is applied on the hiding stage.
If we set this height only on first update, there can be height jump on hiding animation.

src/components/Toaster/types.ts
```diff
@@ -27,12 +27,15 @@ export type ToastProps = {
     content?: React.ReactNode;
     type?: ToastType;
     isClosable?: boolean;
+    // FIXME: BREAKING CHANGE.
+    // It looks like system field. Not part of public API. Maybe it should be deleted.
     isOverride?: boolean;
```
amje, line 32: Can you deprecate this prop please?

amje (approved):

ogonkov (approved):

PR #446, opened 2022-12-29
fix(StoriesGroup): fix click area

GermanVor:
Space between PreviewItems were disabled for user clicking and it is impossible to close StoriesGroup by clicking on area of PreviewItems

### DIFF
--- a/src/components/StoriesGroup/components/StoriesPreview/StoriesPreview.tsx
+++ b/src/components/StoriesGroup/components/StoriesPreview/StoriesPreview.tsx
@@ -8,8 +8,8 @@ import {Button, ButtonProps} from '../../../Button';
 
 import {block} from '../../../utils/cn';
 import './StoriesPreview.scss';
-import {ModalCloseReason} from 'src/components/Modal';
-import {StoriesItemMedia} from 'src/components/Stories/types';
+import {ModalCloseReason} from '../../../Modal';
+import {StoriesItemMedia} from '../../../Stories/types';
 
 const PREVIEW_ITEM_SIZE = 40;
 const PREVIEW_LIST_GAP = 8;
--- a/src/components/StoriesGroup/components/StoriesPreview/StoriesPreview.tsx
+++ b/src/components/StoriesGroup/components/StoriesPreview/StoriesPreview.tsx
@@ -4,12 +4,50 @@ import {MediaRenderer} from '../../../Stories/components';
 
 import {StoriesGroupItem} from '../../types';
 import {Chevron} from '../../../icons/Chevron';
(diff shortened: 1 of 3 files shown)


GermanVor:
and by the way 
upper right semicircular piece is not clickable too
but it is problem of Modal component

image

src/components/StoriesGroup/components/StoriesPreview/StoriesPreview.tsx
```diff
@@ -4,13 +4,51 @@ import {MediaRenderer} from '../../../Stories/components';
 
 import {StoriesGroupItem} from '../../types';
 import {Chevron} from '../../../icons/Chevron';
-import {Button} from '../../../Button';
+import {Button, ButtonProps} from '../../../Button';
 
 import {block} from '../../../utils/cn';
 import './StoriesPreview.scss';
+import {ModalCloseReason} from 'src/components/Modal';
+import {StoriesItemMedia} from 'src/components/Stories/types';
```
amje, line 12: These imports won't work, they should be relative

amje (changes requested):

GermanVor:
@amje 

Maybe we want to add new property for Modal component like `modalContentOutsideClickable: boolean` ?
If this property is true than Modal will call `onClose` for clicking on `.yc-modal__content`.
It will be useful if the Modal content is not in the shape of a rectangle, like on the screen.

src/components/StoriesGroup/components/StoriesPreview/StoriesPreview.tsx
GermanVor: why this imports do not work ?

https://github.com/gravity-ui/uikit/blob/6c37dbe17c8802f06925f6c2fd276505cd87fad3/src/components/StoriesGroup/components/StoriesPreview/StoriesPreview.tsx#L25

https://github.com/gravity-ui/uikit/blob/6c37dbe17c8802f06925f6c2fd276505cd87fad3/src/components/StoriesGroup/components/StoriesPreview/StoriesPreview.tsx#L175

amje: Roots are not configured in the project. This might work for storybook, but not when code is compiled.

GermanVor: [3377e1a](https://github.com/gravity-ui/uikit/pull/446/commits/3377e1a48b8feed7b0a2af732a4650ad39b6a80a)

amje (approved):

GermanVor:
@ValeraS

PR #459, opened 2023-01-09
feat(Select): add defaultOpen prop

DaryaLari:


src/components/Select/__tests__/Select.base-actions.test.tsx
```diff
@@ -40,6 +42,15 @@ describe('Select base actions', () => {
...
         });
+
+        test('default with defaultOpen prop', async () => {
+            await act(async () => {
+                setup({defaultOpen: true});
+            });
+            const selectControl = screen.getByTestId(TEST_QA);
+            expect(selectControl).toHaveClass(SELECT_CONTROL_OPEN_CLASS);
+            screen.getByTestId(SelectQa.POPUP);
+        });
```
korvin89, line 53: Could you add test without using `defaultOpen` property?

DaryaLari: What do you mean? Add a test with checking that popup doesn't appear and Select stays "closed" while rendering without `defaultOpen`  property?

src/components/Select/__tests__/Select.base-actions.test.tsx
```diff
@@ -13,7 +13,9 @@ import {
     setup,
     timeout,
     generateOptions,
+    SELECT_CONTROL_OPEN_CLASS,
 } from './utils';
+import {act, screen} from '@testing-library/react';
 
 describe('Select base actions', () => {
     describe('open popup by', () => {
```
korvin89, line 21: And this test should be outside of "open popup by" describe section

korvin89 (approved):

PR #462, opened 2023-01-10
feat(RadioGroup): add optionClassName prop

trabemz:


src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -51,7 +61,11 @@ export const RadioGroup = React.forwardRef<HTMLDivElement, RadioGroupProps>(func
                 <Radio
                     {...optionProps}
                     key={optionProps.value}
-                    className={b('option')}
+                    className={b(
+                        'option',
+                        optionClassName &&
+                            bemBlock(optionClassName)({checked: optionProps.checked}),
```
amje, line 67: Here we're forcing our BEM naming, this should be up to user. What's your case though?

trabemz: I need to style label of the selected option as well. But I think you're right, I'll remove BEM naming.

amje (approved):

PR #478, opened 2023-01-13
feat(ThemeProvider): add props scoped and modifiers

imsitnikov:


amje (approved):

resure (approved):

imsitnikov:
New properties for `ThemeProvider`:
1) `scoped: boolean` – allows using `ThemeProvider` in localized scope, with `scoped={false}` (default) theme classes will be placed in the `body`, with `scoped={true}` theme classes will be placed in a `div` within `ThemeProvider`.
2) `rootClassName` – allows to add additional classes to the root element (`body` or `div` if we use `scoped={true}`).

PR #482, opened 2023-01-17
feat: add focus trap

ValeraS:


src/components/Modal/Modal.tsx
```diff
@@ -74,6 +90,7 @@ export function Modal({
         contentRefs: [contentRef],
     });
 
+    const _contentRef = useForkRef<HTMLDivElement>(contentRef, setFocusTrap);
```
amje, line 93: Can be named as `handleContentRef`

src/components/Portal/Portal.tsx
```diff
@@ -5,9 +5,15 @@ import {usePortalContainer} from '../utils/usePortalContainer';
...
     children?: React.ReactNode;
+    disablePortal?: boolean;
 }
 
-export function Portal({container, children}: PortalProps) {
+export function Portal({container, children, disablePortal}: PortalProps) {
     const defaultContainer = usePortalContainer();
+
+    if (disablePortal) {
+        return <React.Fragment>{children}</React.Fragment>;
```
amje, line 15: Why wrap in fragment?

ValeraS: Otherwise, when using `Portal`, you get ts error: `Portal` cannot be used as a JSX component. Its return type 'ReactNode' is not a valid JSX element.

src/components/utils/useFocusTrap.ts
```diff
@@ -1,39 +1,72 @@
...
+    return [setFocusTrap, containerProps] as const;
+}
+
+function getFocusElement(root: HTMLElement) {
+    if (
+        !(document.activeElement instanceof HTMLElement) ||
+        !root.contains(document.activeElement)
+    ) {
+        if (!root.hasAttribute('tabIndex')) {
+            if (process.env.NODE_ENV !== 'production') {
```
amje, line 61: We cannot be assured process.env will be defined

src/components/Modal/Modal.tsx
```diff
@@ -88,20 +106,27 @@ export function Modal({
...
                 onExit={() => setInTransition(true)}
-                onEntered={() => setInTransition(false)}
-                onExited={() => setInTransition(false)}
+                onEntered={() => {
+                    setInTransition(false);
+                    onOpenEnd?.();
+                }}
+                onExited={() => {
+                    setInTransition(false);
+                    onCloseEnd?.();
```
amje, line 115: Maybe add lifecycle callbacks for all events? I would name it like so: `onTransitionEnter`, `onTransitionEntered`, `onTransitionExit`, `onTransitionExited`

amje (changes requested):

src/components/utils/useOnFocusOutside.ts
```diff
@@ -0,0 +1,75 @@
...
+ *       <Button onClick={() => {setOpen(true);}}>Select</Button>
+ *       <Popup open={open}>
+ *          ...
+ *       </Popup>
+ *     </span>
+ *   );
+ *  }
+ * }
+ */
+export function useOnFocusOutside(onFocusOutside: (event: FocusEvent) => void, enable = true) {
```
amje, line 37: Let's use object as an argument like in other hooks

amje (changes requested):

amje (approved):

PR #484, opened 2023-01-18
feat!: text semantic colors update

amje:


IsaevAlexandr (approved):

ogonkov:
@amje  That would be nice to publish some migration guide for such subtle changes at least in PR

PR #492, opened 2023-01-20
feat(Layer Manager): support idle layer configs.

yndx-madfriend:
This PR introduces a concept of idle layer configs: an idle config means that layer will be added to the stack only when stack becomes empty. This concept allows to have promotional popovers/popups which would appear only when no other popup is visible to reduce the visual clutter.

idle

In addition to idle, there are also `idleTimeout` (minimum number of milliseconds before idle layer is actually added to an empty stack) and `idlePriority` (which allows to add a single idle layer when multiple idle layers are waiting to be added).  

New properties are supported in `useLayer` hook and `Popover` component. These properties can easily be added to other components on top of `useLayer`. 

Interaction with `useLayer` is reworked in a way that `useLayer` now returns a piece of state (`allowedOpen`). This piece of state is connected to LayerManager and for idle layers `allowedOpen` updates when the layer is finally added to the stack.

I've also provided a small story that shows the desired behavior of idle popovers, as well as a set of tests for LayerManager itself.


ValeraS (changes requested):

src/components/utils/LayerManager.ts
```diff
@@ -17,27 +18,145 @@ export type ContentElement =
...
+        }
+    }
+
+    reset() {
+        this.stack = [];
+        this.preStack = [];
+        this.unscheduleCandidate();
+        this.mouseDownTarget = null;
+        this.removeListeners();
+    }
```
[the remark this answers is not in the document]
yndx-madfriend: fixed

src/components/utils/LayerManager.ts
```diff
@@ -17,27 +18,145 @@ export type ContentElement =
...
+            }
         }
+
+        return Promise.resolve();
     }
 
     remove(config: LayerConfig) {
         const index = this.stack.indexOf(config);
+
+        if (index < 0) {
```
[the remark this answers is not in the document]
yndx-madfriend: fixed

src/components/utils/useLayer.ts
```diff
@@ -55,7 +63,12 @@ export function useLayer({
     React.useEffect(() => {
         if (open && enabled) {
             const layerConfig = layerConfigRef.current;
-            layerManager.add(layerConfig);
+            const tryAdd = async () => {
+                await layerManager.add(layerConfig);
+                setAllowedOpen(true);
+            };
+
+            tryAdd();
```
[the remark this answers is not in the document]
yndx-madfriend: fixed

src/components/utils/useLayer.ts
```diff
@@ -28,8 +31,13 @@ export function useLayer({
         onOutsideClick,
         onClose,
         contentRefs,
+        idle,
```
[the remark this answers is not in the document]
yndx-madfriend: If I understood you correctly, changes of `idle` are not handled right now (and personally I wouldn't like `idle` to have a dynamic value), changes of `open` and `enabled` are handled since they are dependencies of `useEffect`: https://github.com/gravity-ui/uikit/pull/492/files/db588767a7cb9d6b1668a51c743c99edb8f914b7#diff-d28d4a02c53c486a096624c2ebcb7af8be0a1fecc256edaf8f226cd2aadaa016L66

yndx-madfriend:
@ValeraS I've added the logic to reset `allowedOpen` once the layer is removed from the stack. Now I struggle with incorrect placement of Popover which seems to be unrelated to my branch.. Have you encountered this behavior? 

Screen Shot 2023-03-02 at 20 03 45

yndx-madfriend:
^ I've fixed the problem with wrong position, it was a bug in our storybook (https://github.com/gravity-ui/uikit/pull/562)

I've also froze the `idle` property, meaning that components can not change the value of `idle` in the runtime – IMO popups are either idle at the start or non-idle.

src/components/utils/LayerManager.ts
```diff
@@ -17,27 +18,163 @@ export type ContentElement =
...
+            config.onRemove?.();
+        }
+
+        let preStackIndex = -1;
+        for (let i = 0; i < this.preStack.length; i++) {
+            if (this.preStack[i].config === config) {
+                preStackIndex = i;
+                break;
+            }
+        }
```
ValeraS, line 138:
```suggestion
        let preStackIndex = this.preStack.findIndex((item) => item.config === config);
```

src/components/utils/LayerManager.ts
```diff
@@ -17,27 +18,163 @@ export type ContentElement =
...
+    }
+
+    return Math.sign(priorityB - priorityA);
+};
+
+export class LayerManager {
     private stack: LayerConfig[] = [];
+    private preStack: LayerCandidate[] = [];
     private mouseDownLayerTarget?: {layer: LayerConfig; target: HTMLElement};
+    private scheduledAdd: ReturnType<typeof setTimeout> | undefined;
```
ValeraS, line 74: let's rename to `scheduleAddTimeoutId`

src/components/utils/LayerManager.ts
```diff
@@ -17,27 +18,163 @@ export type ContentElement =
...
+            this.addToStack(scheduledCandidate.config);
+            scheduledCandidate.resolve();
+            this.unscheduleCandidate();
+        };
+
+        const timeout = Number.isInteger(candidate.config.idleTimeout)
+            ? candidate.config.idleTimeout
+            : defaultIdleTimeout;
+
+        this.scheduledAdd = setTimeout(scheduledAdd.bind(this, candidate), timeout);
```
ValeraS, line 104:
```suggestion
        this.scheduledAdd = setTimeout(scheduledAdd.bind(null, candidate), timeout);
```

src/components/utils/LayerManager.ts
```diff
@@ -17,27 +18,163 @@ export type ContentElement =
...
+                if (entry.config === config) {
+                    return entry.promise;
+                }
+            }
+
+            return this.addToPreStack(config);
+        } else {
+            this.addToStack(config);
         }
+
```
ValeraS, line 120:
```suggestion
        } 
        
        this.addToStack(config);
```

src/components/utils/LayerManager.ts
```diff
@@ -17,27 +18,163 @@ export type ContentElement =
...
+    }
+
+    private unscheduleCandidate() {
+        if (this.scheduledCandidate) {
+            this.scheduledCandidate = undefined;
+        }
+        if (this.scheduledAdd) {
+            clearTimeout(this.scheduledAdd);
+            this.scheduledAdd = undefined;
         }
```
ValeraS, line 178:
```suggestion
        clearTimeout(this.scheduledAdd);
        this.scheduledCandidate = undefined;
        this.scheduledAdd = undefined;
```

src/components/utils/useLayer.ts
```diff
@@ -50,12 +63,20 @@ export function useLayer({
...
+                .add(layerConfig)
+                .then(() => {
+                    setAllowedOpen(true);
+                })
+                .catch(() => {
+                    setAllowedOpen(false);
+                });
 
             return () => {
                 layerManager.remove(layerConfig);
```
ValeraS, line 82: Do we really need `layerConfig.onRemove` callback?

```suggestion
                layerManager.remove(layerConfig);
                setAllowedOpen(false);
```

src/components/utils/useLayer.ts
```diff
@@ -28,8 +35,13 @@ export function useLayer({
         onOutsideClick,
         onClose,
         contentRefs,
+        idle: frozenIdle,
+        idlePriority: frozenIdlePriority,
+        idleTimeout: frozenIdleTimeout,
```
ValeraS, line 40: Sorry, looked at it again and I think it was ok at first. Let's restore it as it was.
```suggestion
        idle,
        idlePriority,
        idleTimeout,
```

ValeraS (changes requested):

src/components/utils/LayerManager.ts
yndx-madfriend: Wow, I didn't know it works like that – this `bind(...)` seems to completely ignore `null` because scheduledAdd is an arrow function (already bound).

src/components/utils/useLayer.ts
yndx-madfriend: I thought about this, but current config can be removed from layerManager independently from `useLayer` (e.g. click outside popup). This way discussed effect cleanup will not be called and `allowedOpen` will be stuck at `true`

yndx-madfriend: You are right. Fixed

yndx-madfriend:
@amje could you also take a look?

yndx-madfriend:
We decided to move idle logic to HelpPopup component. The only thing required from LayerManager is to notify us of stack changes, so that once stack is empty, HelpPopup can open.

PR #509, opened 2023-01-27
fix(TextInput): no unexpected autocomplete when label prop present

### LINKED ISSUE #507: Do not show autocomplete for TextInput with inner label if no id is set
Default behaviour for TextInput with inner labels should be not to show autocomplete if no `id` prop is specified.

This will help users to avoid explicit mention of autocomplete="off" every time when using TextInput with inner label and there is not need for autocomplete suggestions.

refactor256:
fixes #507

- fixed unexpected autocomplete behavior for case with label prop as well
- split tests for plain input and textarea
- add some tests for autocomplete


Marginy605 (approved):

refactor256:
@amje @ValeraS waiting for your review

amje:
@refactor256  `pointer-events: none` should resolve all the problems without using id attribute

refactor256:
> @refactor256 `pointer-events: none` on `<label>` should resolve all the problems without using id attribute

@amje the drawback is that for very long label, which are ending with ellipsis (...), there shall be no tooltips on hover

Screenshot 2023-02-03 at 21 35 17

Screenshot 2023-02-03 at 21 35 32

amje (approved):

PR #519, opened 2023-01-31
feat: introduce InfiniteScroll component

isqua:
I added a component for infinite scroll. The component renders child elements. If you scroll it to the bottom, it will show a loader and call an `onActivate` callback. When all the data is loaded and there is no need to load more, pass the `disabled={true}` property.


isqua:
It looks like the preview build is broken #520

isqua:
I’ve just rebased the branch after the fix of ci issues.

PR #545, opened 2023-02-20
fix: base-generic-ultralight color in dark-hc theme

Nil2000:
Updated color value for base-generic-ultralight in dark-hc theme


amje:
Thanks for resolving the issue!

amje (approved):

PR #551, opened 2023-02-27
feat: unstable layout components set

IsaevAlexandr:
continuation of https://github.com/gravity-ui/uikit/pull/430

About changes:

Briefly:

introduces the concept of spacing and the ability to work with media query expressions;
new LayoutContext provided props for base layout and other components;
mobile-first grids with 12-th column layout;
More details:

new css variables with spacing. Can be used and redefined at the project level;
basic media query expressions are introduced. The values can be redefined in the scss file at the project level;
new LayoutProvider component, where you can set default properties for new layout components;
useLayoutContext hook to access the current active media expression and the calculated props from the topic for this expression;
new Container component for centering content. Gets default props from LayoutContext depending on the current media screen value. Knows how to set the "rhythm" between the child components of Row;
new Row component. Gets default props from LayoutContext depending on the current media screen value. It is a container for Col components;
new Col component. Provides an opportunity to declaratively describe how many columns of a 12-column grid the component will occupy. If there are no props, then by default the component takes up all the available space in the row. Used only inside the Row component;
new Flex component. Declarative description of the flexbox model in the jsx world. Retrieves active media expressions from contexts, so you can specify property values as functions of media expressions. It is also possible to set margins between child elements via the spacing prop (analogous to the Stack component);
a new utility for generating classes to work with spacing indents. Motivation - no need to write a separate file with indents, you can use pre-generated classes;
warnings:

negative margins for offsets between elements. You need to be careful with the background-color properties of such elements;


src/components/Layout/Col/Col.tsx
```diff
@@ -0,0 +1,59 @@
+/* eslint-disable valid-jsdoc */
```
korvin89, line 1: @amje maybe we should disable this rule in config?

src/components/Layout/Flex/Flex.tsx
```diff
@@ -0,0 +1,232 @@
+/* eslint-disable react/display-name */
```
korvin89, line 1: You may add name to Flex
```
Flex.displayName = 'LayoutFlex'
```

src/components/Layout/demo/ColPresenter/ColPresenter.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import {Box} from '../Box/Box';
+import {Col, ColProps} from '../../Col/Col';
+
+// @ts-ignore-error
```
korvin89, line 5: What do this directive?

src/components/Layout/hooks/useCurrentActiveMediaQuery.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    removeListeners(fn: () => void) {
+        this.queryListsDecl.forEach(([_, queryList]) =>
+            queryList.removeEventListener('change', fn),
+        );
+    }
+}
+
+/**
+ * @private - use `useLayoutContext` hook instead
```
korvin89, line 74: Maybe we should not give this hook in api at all?

src/components/Layout/hooks/useCurrentActiveMediaQuery.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+        queries.addListeners(onChange);
+
+        setState();
+
+        return () => {
+            mounted = false;
+            queries.removeListeners(onChange);
+        };
+        // don't support runtime breakpoint redefinition. Breakpoints defined only one at LayoutTheme
```
korvin89, line 104: Is it by design?

korvin89:
Could you add a little bit documentation to each component?

src/components/Layout/demo/ColPresenter/ColPresenter.tsx
IsaevAlexandr: ignore type error in demo component. 
I omit nullable values here

src/components/Layout/hooks/useCurrentActiveMediaQuery.tsx
IsaevAlexandr: runtime breakpoint redefinition is a weird thing i guess)

IsaevAlexandr:
> Could you add a little bit documentation to each component?
All components already includes `jsdoc` on hover feature now. And yes, there is work in progress with general documentation for all components. It will be added soon
image

korvin89 (approved):

src/components/Layout/Col/Col.tsx
amje: Or not using jsdoc comments when it's not jsdoc)

src/components/Layout/Flex/Flex.tsx
amje: or simply `Flex`

src/components/Layout/Col/Col.tsx
IsaevAlexandr: If there is alternatives, pls tell me)

This this comment is developers best friends. Life will not be the same as before if you start using all the benefits of jsdoc on hover. Vscode has build in support of this feature

IsaevAlexandr: like this

image

amje (changes requested):

src/components/Layout/Col/Col.tsx
amje: What exactly those eslint errors are?

IsaevAlexandr: This is deprecated rule https://eslint.org/docs/latest/rules/valid-jsdoc

image

src/components/Layout/Flex/Flex.tsx
```diff
@@ -0,0 +1,232 @@
...
+            ),
+            style: {
+                width,
+                alignSelf,
+                flexDirection:
+                    typeof direction === 'function' ? direction(isMediaActive) : direction,
+                flexGrow: grow === true ? 1 : grow,
+                flexWrap: wrap,
+                flexBasis: basis,
+                flexShrink: shrink,
```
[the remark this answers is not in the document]
IsaevAlexandr: for example, you want your component to take up all the space inside the flex container:
```tsx
<Flex>
  <Flex grow={1} shrink={0} basis={200}>...<Flex />
</Flex>
```

In this example component will take all space inside container if container width more than 200px but not less than 200px

IsaevAlexandr:
/ping
added documentation yesterday

amje (approved):

PR #564, opened 2023-03-10
feat(Tabs): add AdjustableTabs for Tabs component

Flunt1k:
Introduce new way to render Tabs when space is not enough for full bar render. 

AdjustableTabs renders only when respective props (**adjustable**) provided. It will calc the width of container and determine how to hide tabs. 

There are 2 way to do it: 
1. Hide behind the Select
2. Hide behind the More button

Also new optional property was added for tabs - **breakpointsConfig**. It is relevant only for AdjustableTabs and will affect how the free space for tabs will be calculated

### DIFF
--- a/src/components/Tabs/AdjustableTabs.tsx
+++ b/src/components/Tabs/AdjustableTabs.tsx
@@ -870,6 +870,7 @@ class AdjustableTabs extends React.Component<AdjustableTabsProps, AdjustableTabs
 
         return (
             <Select
+                className={b('tabs-as-select-control')}
                 onUpdate={this.handleTabsAsSelectUpdate}
                 options={itemsForSelect}
                 value={activeTabID === undefined ? [] : [activeTabID]}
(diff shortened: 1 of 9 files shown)


src/components/Tabs/AdjustableTabs.tsx
```diff
@@ -0,0 +1,890 @@
...
+                return true;
+            }
+        }
+
+        return false;
+    };
+
+    handleResize = () => {
+        if (this.state.isSelectOpened) {
+            // https://github.com/gravity-ui/uikit/issues/553
```
korvin89, line 589: We have already closed this issue, you can do this staff a little bit better

amje:
I expected AdjustableTabs will be using Tabs as a base. But there I see completely separate implementation, different layout, different styles. Is it possible to use Tabs as a base here @Flunt1k ?

Flunt1k:
I will transfer it to another repository:  @gravity-ui/components

PR #567, opened 2023-03-14
fix(DropdownMenu): remove required path from DropdownMenuItemProps, fix navigation

kseniya57:
- removed required path from DropdownMenuItemProps
- fixed navigation of no items available

### DIFF
--- a/src/components/utils/useListNavigation/useListNavigation.ts
+++ b/src/components/utils/useListNavigation/useListNavigation.ts
@@ -43,6 +43,11 @@ export function useListNavigation<ItemType, AnchorType extends HTMLElement>({
             return;
         }
 
+        const canNavigate = items.some((item) => !skip?.(item));
+        if (!canNavigate) {
+            return;
+        }
+
         const anchor = anchorRef?.current;
         if (!anchor) {
             return;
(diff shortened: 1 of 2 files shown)


src/components/utils/useListNavigation/useListNavigation.ts
```diff
@@ -38,8 +38,16 @@ export function useListNavigation<ItemType, AnchorType extends HTMLElement>({
         }
     }, [items, reset]);
 
-    useLayoutEffect(() => {
+    const canNavigate = useMemo(() => {
```
ogonkov, line 41: I guess we can avoid memoization for boolean value

kseniya57: I thought we don't need to calculate skip for every element on every render, since there can be a lot of items, but you are right, it is a rare case that the available element will not be at the beginning

ogonkov: Yeah, but useMemo is not free to spend it for booleans

ogonkov (approved):

axtk (approved):

PR #568, opened 2023-03-15
refactor: extract `<ControlLabel/>` component

ogonkov:


src/components/CheckedLabel/CheckedLabel.tsx
```diff
@@ -0,0 +1,49 @@
...
+import {DOMProps, QAProps} from '../types';
+import {block} from '../utils/cn';
+
+import './CheckedLabel.scss';
+
+type Props = PropsWithChildren<{
+    labelClassName?: string;
+    title?: string;
+    disabled?: boolean;
+    size?: 'm' | 'l';
```
korvin89, line 11: Maybe we should add one type for this sizes collection?

src/components/CheckedLabel/CheckedLabel.tsx
```diff
@@ -0,0 +1,49 @@
+import React, {forwardRef, PropsWithChildren, ReactNode} from 'react';
+import {DOMProps, QAProps} from '../types';
+import {block} from '../utils/cn';
+
+import './CheckedLabel.scss';
+
+type Props = PropsWithChildren<{
```
korvin89, line 7: Maybe text should added by property label/text

ogonkov: I think passing through `children` is more robust

src/components/CheckedLabel/types.ts
```diff
@@ -0,0 +1,14 @@
+import {PropsWithChildren, ReactNode} from 'react';
+import {DOMProps, QAProps} from '../types';
+
+export type Sizes = 'm' | 'l';
+
+export type Props = PropsWithChildren<{
```
amje, line 6: Let's use a full name for Sizes and Props, and `Sizes` should be `Size`

src/components/CheckedLabel/types.ts
```diff
@@ -0,0 +1,14 @@
...
+    labelClassName?: string;
```
amje, line 7: simply `className`, there is no more elements to add class to

src/components/CheckedLabel/CheckedLabel.tsx
```diff
@@ -0,0 +1,39 @@
+import React, {forwardRef} from 'react';
+import {block} from '../utils/cn';
+import {Props} from './types';
+
+import './CheckedLabel.scss';
+
+const b = block('checked-label');
+
+export const CheckedLabel = forwardRef<HTMLLabelElement, Props>(
```
amje, line 9: I would name the component as `ControlLabel`, checked word is too specific

amje (changes requested):

src/components/CheckedLabel/types.ts
ogonkov: There is also `className`, inherited from `DOMProps`

src/components/CheckedLabel/CheckedLabel.tsx
ogonkov: Not sure that this component will fit regular TextInput

amje (approved):

trabemz (approved):

korvin89 (approved):

PR #586, opened 2023-03-28
feat: export LayoutDirection

niktverd:


src/components/SharePopover/index.ts
```diff
@@ -1,4 +1,4 @@
-export {ShareOptions} from './constants';
+export {ShareOptions, LayoutDirection} from './constants';
```
amje, line 1: It should be `ShareLayoutDirection`

amje (changes requested):

src/components/SharePopover/index.ts
niktverd: Ok

amje (approved):

PR #597, opened 2023-03-30
feat(Tabs): add support for tabs via `children`

ogonkov:


src/components/Tabs/Tabs.tsx
```diff
@@ -28,70 +29,84 @@ export interface TabsProps extends QAProps {
...
+const getActiveTabId = (
+    activeTab: TabsProps['activeTab'],
+    allowNotSelected: TabsProps['allowNotSelected'],
+    items: TabsProps['items'],
+) => {
+    if (activeTab) {
+        return activeTab;
+    }
+
+    if (allowNotSelected || items?.length === 0) {
```
sofiushko, line 53: We need a the second condition here?

src/components/Tabs/Tabs.tsx
```diff
@@ -28,70 +29,84 @@ export interface TabsProps extends QAProps {
...
+    /** Allows to wrap `TabItem` into another component or render custom tab.
+     * Ignored when tabs rendered via `children` */
     wrapTo?(item: TabsItemProps, node: React.ReactNode, index: number): React.ReactNode;
 }
 
-export const Tabs = forwardRef<HTMLDivElement, TabsProps>(
+const getActiveTabId = (
+    activeTab: TabsProps['activeTab'],
+    allowNotSelected: TabsProps['allowNotSelected'],
+    items: TabsProps['items'],
```
sofiushko, line 47: Let's do this prop not optional?

src/components/Tabs/Tabs.tsx
ogonkov: I have moved this code as is, since we don't need to memoize string

sofiushko (approved):

PR #599, opened 2023-03-31
feat(Persona): add component

ogonkov:


src/components/Persona/__stories__/Persona.stories.tsx
```diff
@@ -0,0 +1,50 @@
...
+export default {
+    title: 'Components/Persona',
+    component: Persona,
+} as ComponentMeta<typeof Persona>;
+
+const person = 'Charles Darwin';
+const email = faker.internet.email(...person.split(' '));
+const personImg = faker.internet.avatar();
+
+const Template: ComponentStory<typeof Persona> = (args) => <Persona {...args} />;
```
DaffPunks, line 15: Looks like we missed borders here:

image

Seems prop "theme" doesn't work

ogonkov: It's strange, i have borders here

ogonkov: Oh, i see it, bug is reproduced in chrome-based browsers, but not firefox

DaffPunks (approved):

korvin89 (approved):

PR #601, opened 2023-04-03
fix(Select): fix render empty group label

Flunt1k:


src/components/Select/components/SelectList/GroupLabel.tsx
```diff
@@ -8,8 +8,9 @@ type GroupLabelProps = {
 };
 
 export const GroupLabel = ({label}: GroupLabelProps) => {
+    const isEmpty = label === '';
```
korvin89, line 11: Looks a little bit redundant

korvin89 (approved):

PR #609, opened 2023-04-06
feat(SharePopover): export LayoutDirection

niktverd:


src/components/SharePopover/index.ts
```diff
@@ -1,4 +1,4 @@
-export {ShareOptions} from './constants';
+export * from './constants';
```
amje, line 1: If exporting all then members should be named with `Share*`, LayoutDirection is too common name

amje (changes requested):

PR #615, opened 2023-04-11
feat(Layout): changed letter designation of the spacings to a digital

IsaevAlexandr:
changes:
- switch numeric spacings to digits:
   - nano -> 0,5;
   - micro -> 1;
   - sm -> 2;
   - s -> 3;
   - m -> 4;
   - l -> 5;
   - ...
- Row component now has required prop `space` to exclude implicit usage of theme. This is done in order to eliminate bugs during the change of the project theme with the time. Looking at the component, you need to immediately understand what spacings it sets to child elements, implicit use is unacceptable
- on server side now activeMediaQuery has `s` value. You can override this value by optional `initialMediaQuery` prop at `LayoutProvider`


src/components/layout/constants.ts
```diff
@@ -1,16 +1,9 @@
...
-    xs: 8,
-    s: 12,
-    m: 16,
-    l: 20,
-    xl: 24,
-    xxl: 32,
-};
+// css custom properties doesn't support decimal numbers in name
+export const CSS_SIZE_EXCEPTION = {
+    '0.5': 'micro',
```
amje, line 5: `'half'` or `'05'`?

styles/spacing.scss
```diff
@@ -1,11 +1,16 @@
...
-    --yc-spacing-none: 0px;
-    --yc-spacing-nano: 2px;
-    --yc-spacing-micro: 4px;
-    --yc-spacing-xs: 8px;
-    --yc-spacing-s: 12px;
-    --yc-spacing-m: 16px;
-    --yc-spacing-l: 20px;
-    --yc-spacing-xl: 24px;
-    --yc-spacing-xxl: 32px;
+    --base-space-unit: 4px;
```
amje, line 2: `yc-space-base`, unit is px in this case

amje (changes requested):

amje (approved):

PR #632, opened 2023-04-18
feat(Persona): add `renderButton` prop

ogonkov:
Added new `renderButton` prop, that allows to override default button. Default button is now exposed as `<Persona.Button/>` for convenience.


src/components/PersonaWrap/PersonaWrap.tsx
```diff
@@ -1,46 +1,33 @@
...
     onClick,
     onClose,
+    renderButton,
     className,
     avatar,
     children,
 }) => {
     const clickable = Boolean(onClick);
-    const closeable = Boolean(onClose);
+    const closeable = Boolean(onClose || renderButton);
```
DaffPunks, line 22: Мне кажется теперь у этой переменной другой смысл. Не `closeable`, а какой-нибудь `withAction`

DaffPunks (approved):

src/components/PersonaWrap/PersonaWrap.tsx
ogonkov: It's alvays was like that)

src/components/Persona/Persona.tsx
```diff
@@ -47,13 +30,18 @@ export const Persona: FC<PersonaProps> = ({
             break;
     }
 
+    const renderCloseButton = useCallback(
```
amje, line 33: This button should always be close button. Let's not add a custom renderer here

amje (changes requested):

ogonkov:
I guess not

PR #637, opened 2023-04-19
chore: deprecate theme change via context methods

ogonkov:


resure:
I would suggest making this change during next major release.
Technically deprecation will not break something, but it can be pretty annoying.
cc @amje

ogonkov:
@resure but this is the whole point, first you deprecate something, then you remove it next major release.

I didn't get how deprecation comment could annoy anyone.

resure:
I don't hold strong opinion on this topic, but since this change is not urgent for us it would be nice to give a little more time to properly migrate — for example, deprecate it in the next major and completely remove in the one after that.

ogonkov:
@resure is such mechanics to help users to migrate

amje (approved):

resure (approved):

PR #640, opened 2023-04-19
fix(Table): withTableSorting can get sort value using column's ID written as path

chervyakovru:
Just duplicate get logic from main Table code: https://github.com/gravity-ui/uikit/blob/main/src/components/Table/Table.tsx#L171

### DIFF
--- a/src/components/Table/hoc/withTableSorting/withTableSorting.tsx
+++ b/src/components/Table/hoc/withTableSorting/withTableSorting.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 
+import _get from 'lodash/get';
 import _memoize from 'lodash/memoize';
 
 import {block} from '../../../utils/cn';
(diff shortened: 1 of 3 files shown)


src/components/Table/hoc/withTableSorting/withTableSorting.tsx
```diff
@@ -40,7 +41,7 @@ export function withTableSorting<I extends TableDataItem, E extends {} = {}>(
     const displayName = `withTableSorting(${componentName})`;
 
     function defaultCompareFunction(itemA: I, itemB: I, columnId: string) {
-        if (itemA[columnId] === itemB[columnId]) {
+        if (_get(itemA, columnId) === _get(itemB, columnId)) {
```
ogonkov, line 44: What was the problem here?

chervyakovru: for example: 
in the table we use Item with type: `{foo: {bar: string}, foo2: string, foo3: string}`
for columns Ids we use `['foo.bar', 'foo2', 'foo3']`

and right now, if we want sort table by first column, in the if we will get `undefined` from both items. Adding `_get` solve this issue

ogonkov: Then it looks like new feature, not a fix

```
feat(Table): add support for sort by nested properties
```

chervyakovru:
@ogonkov @krozhkov
Hi! Can we merge it?)

ogonkov:
Please rebase PR

Raubzeug:
Please check linter status, it fail now.

src/components/Table/hoc/withTableSorting/withTableSorting.tsx
Raubzeug: It seem it will be useful to add nested properties to stories and tests.

Raubzeug (changes requested):

Raubzeug (approved):

Raubzeug (approved):

amje (approved):

PR #641, opened 2023-04-19
feat(UserAvatar): add `srcset` support

ogonkov:


src/components/index.ts
```diff
@@ -67,3 +67,5 @@ export * from './utils/setRef';
 export {useOnFocusOutside} from './utils/useOnFocusOutside';
 export * from './utils/xpath';
 export * from './utils/useFileInput/useFileInput';
+export * from './utils/getSrcSet';
+export * from './utils/getClosestNumber';
```
amje, line 71: Why we export these two?

src/components/UserAvatar/UserAvatar.tsx
```diff
@@ -1,28 +1,58 @@
...
+                    srcSet={srcSet}
+                    sizes={sizes}
+                    alt={''}
+                />
+            </div>
+        );
+    }
+
     return (
         <div
```
amje, line 51: I think we should always use  in the component

src/components/index.ts
ogonkov: I guess they could be omited, will remove this exports

src/components/utils/getClosestNumber/__tests__/getClosestNumber.test.ts
```diff
@@ -0,0 +1,6 @@
+import {getClosestNumber} from '../getClosestNumber';
+
+test('should find appropriate number', () => {
+    expect(getClosestNumber(1, [1, 2, 3, 50])).toBe(1);
+    expect(getClosestNumber(20, [1, 2, 15, 30, 50])).toBe(30);
```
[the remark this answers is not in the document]
ogonkov: Yeah, just next number that is bigger than given or equal to it

amje (approved):

PR #651, opened 2023-04-27
Rounded style should be applied for all sizes and fix icon on s size

teleginzhenya:
* Dunno about sizes in specifications, but I guess should be height -4px, will ask design team about it
* extra: fix icon on s size


teleginzhenya:
@amje @goshander

PR #661, opened 2023-05-11
fix(Select): align items by baseline for label and values

artemmufazalov:
Due to SelectControl label font-weight, label and value don't align correctly

Before
Screen Shot 2023-05-11 at 17 16 06

After
Screen Shot 2023-05-11 at 17 16 37


korvin89:
I cannot reproduce this bug

korvin89:
I reproduced the bug. Lets try this solution:
```scss
& #{$buttonBlock}__text {
    ...
    align-items: baseline;
}
```

korvin89 (approved):

PR #665, opened 2023-05-16
feat(Label): added value prop for "key:value" labels

sunduckcow:
in this pr
fixed: bad display of icon in 's' size labels
refactored: styles for different sizes
added: support for value prop + story for display all combinations of labels

### DIFF
--- a/src/components/Label/Label.scss
+++ b/src/components/Label/Label.scss
@@ -2,16 +2,30 @@
 
 $block: '.#{variables.$ns}label';
 $disabled: #{$block}_disabled;
-$transitionDuration: 0.3s;
+$transitionDuration: 0.15s;
 $transitionTimingFunction: ease-in-out;
 
-@mixin themeState($bgColor, $bgHoverColor, $textColor) {
+@mixin themeState($bgColor, $bgHoverColor, $textColor, $borderColor: none) {
     color: #{$textColor};
     background-color: #{$bgColor};
 
-    // hover on interactive label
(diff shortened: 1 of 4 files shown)


amje (changes requested):

src/components/Label/Label.scss
```diff
@@ -250,5 +204,14 @@ $transitionTimingFunction: ease-in-out;
                 var(--yc-color-text-complementary)
             );
         }
+
+        &_clear {
+            @include themeState(
+                none,
```
amje, line 210: `none` is not valid for `background-color`. Either use `background` property or change `none` to `transparent`

amje (changes requested):

src/components/Label/Label.scss
```diff
@@ -2,18 +2,32 @@
...
 $transitionTimingFunction: ease-in-out;
 
-@mixin themeState($bgColor, $bgHoverColor, $textColor) {
+@mixin themeState($bgColor, $bgHoverColor, $textColor, $BorderColor: none) {
     color: #{$textColor};
     background-color: #{$bgColor};
 
-    // hover on interactive label
-    &:not(#{$disabled})#{$block}_is-interactive:hover {
+    @if $BorderColor != none {
```
amje, line 12: Why is var name capitalized? It's a typo?

amje (changes requested):

goshander (approved):

PR #667, opened 2023-05-16
feat(ClipboardButton): added tooltip to clipboard button

catakot:


SeqviriouM:
@damask can you take a look

src/components/ClipboardButton/ClipboardButton.tsx
```diff
@@ -1,51 +1,85 @@
...
+    /** Text shown after copy */
+    endCopyTip?: string;
 }
 
 const b = block('clipboard-button');
 
 const DEFAULT_ICON_SIZE = 24;
 const DEFAULT_TIMEOUT = 1000;
+const DEFAULT_START_COPY_TIP = 'Copy';
+const DEFAULT_END_COPY_TIP = 'Copied!';
```
amje, line 43: Texts should be defined via our i18n

src/components/ClipboardButton/ClipboardButton.tsx
```diff
@@ -1,51 +1,85 @@
...
+    onClick?: ButtonProps['onClick'];
+    /** tooltip component pops */
+    tooltipProps?: ClipboardButtonTooltipProps;
+}
+
+interface ClipboardButtonTooltipProps extends Omit<TooltipProps, 'content' | 'children'> {
+    /** Text shown before copy */
+    startCopyTip?: string;
+    /** Text shown after copy */
+    endCopyTip?: string;
```
amje, line 35: Do we actually need customization here? I mean all `tooltipProps`. We can add some in the future if it would be requested.

amje (changes requested):

src/components/ClipboardButton/ClipboardButton.tsx
catakot: done

src/components/ClipboardButton/ClipboardButton.tsx
catakot: It will save our time in future. Almost all future changes in Tooltip will be automatically available here. 

Why is it better to restrict users of using extra props? Let's give them more opportunity to customise  tooltip.

amje (approved):

PR #677, opened 2023-05-26
[hooks] useInfinityFetch

imechoim:


ogonkov (commented):
@amje should we move it to `@gravity-ui/components` maybe?

PR #696, opened 2023-05-30
fix(Select): add filterable props in storybook

dividead:


src/components/Select/__stories__/SelectShowcase.tsx
```diff
@@ -301,6 +301,7 @@ export const SelectShowcase = (props: SelectProps) => {
                 code={[EXAMPLE_CUSTOM_FILTER_SECTION]}
                 selectProps={{
                     ...props,
+                    filterable,
```
korvin89, line 304: There is a syntax error here. U don't declare `filterable` variable

korvin89 (changes requested):

korvin89:
I think that setting this prop makes the example less transparent. The user needs to see that the default filter will not be shown even if the function for rendering it is defined. Probably not the best solution, but this is the current behavior.

PR #697, opened 2023-05-30
feat(Select): add SelectAsync

imechoim:
**Features**

Select:
- `SelectProps.options` (extended to accept async function to fetch options. uses `SelectAsync` under the hood)
- `SelectAsync` (add - on over `Select` that use `useSelectLoading` and `useInfinityFetch` to bring async option fetch abilities).

Select utils:
- `useSelectLoading` (mix loading indicator to option list and invoke `onIntersect` callback)
- `useInfinityFetch` (hook that facilitates infinite scrolling by fetching data in a paginated manner)

Utils:
- `useIntersection` ([Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API))
- `useMountedState` (a primitive for tracking the mounted state)

### DIFF
--- a/src/components/utils/Select/useSelectLoading/README.md
+++ b/src/components/utils/Select/useSelectLoading/README.md
@@ -34,4 +34,3 @@ const ExampleSelect = () => {
 
 - `options`: an array of options to be displayed in the Select component. May include an additional loading option, which is depends on the presence of the `onFetch` prop..
 - `renderOption`: a function to render options for the Select component. Wraps the `renderOption` prop to render the loading component.
-- `filterOption` a function that render persistant loading indicator while `loading`
--- a/src/components/utils/Select/useSelectLoading/useSelectLoading.tsx
+++ b/src/components/utils/Select/useSelectLoading/useSelectLoading.tsx
@@ -12,11 +12,12 @@ import {SelectLoadingIndicator} from './SelectLoadingIndicator';
 
 const LOADER = 'LOADER';
 const LOADER_PERSISTENT = 'LOADER_PERSISTENT';
-const loaderOption = {value: LOADER, content: LOADER, disabled: true};
+const loaderOption = {value: LOADER, content: LOADER, disabled: true, notFilterable: true};
 const loaderPersistentOption = {
     value: LOADER_PERSISTENT,
     content: LOADER_PERSISTENT,
     disabled: true,
+    notFilterable: true,
 };
 
 export type UseSelectLoading = {
@@ -29,7 +30,6 @@ export const useSelectLoading = ({
     renderOption,
     onFetch,
     loading,
-    filterOption,
 }: UseSelectLoading) => {
     const isMounted = useMountedState();
     const [isFetching, setIsFetching] = React.useState(false);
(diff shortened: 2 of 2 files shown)


src/components/Select/types.ts
```diff
@@ -70,6 +74,17 @@ export type SelectProps<T = any> = QAProps &
...
 
+export type SelectBasicProps<Option = any> = Omit<SelectProps<Option>, 'options'> & {
+    options?: SelectOptionType<Option>[];
+};
+
+export type SelectAsyncProps<Option = any, Pagination = any> = Omit<
+    SelectProps<Option>,
+    'options'
+> & {
+    options: SelectAsyncOptionType<SelectOptionType<Option>[], Pagination>;
```
korvin89, line 85: What about initial options?

src/components/utils/Select/useSelectInfinityFetch/useInfinityFetch.ts
```diff
@@ -0,0 +1,73 @@
...
+) => Promise<PaginationResponse<Response, Pagination>>;
+
+export const useInfinityFetch = <Response, Pagination>({
+    fetcher,
+}: {
+    fetcher: Fetcher<Response, Pagination>;
+}) => {
+    const mounted = React.useRef(false);
+    const [isLoadingInitial, setIsLoadingInitial] = React.useState(false);
+    const [isLoadingInfinity, setIsLoadingInfinity] = React.useState(false);
```
korvin89, line 19: Maybe Infinity -> Pagination?

src/components/Select/__stories__/SelectShowcase.tsx
```diff
@@ -149,6 +170,14 @@ export const SelectShowcase = (props: SelectProps) => {
 
     return (
         <div className={b()}>
+            <ExampleItem
```
korvin89, line 173: In should not be the first example

src/components/Select/wrappers/SelectAsync/SelectAsync.tsx
```diff
@@ -0,0 +1,32 @@
...
+        const {responses, onFetchInfinity, isLoadingInitial} = useInfinityFetch({
+            fetcher,
+        });
+
+        const mappedOptions = React.useMemo(
+            () => flatMap(responses, ({response}) => response),
+            [responses],
+        );
+
+        const asyncProps = useSelectLoading({
```
korvin89, line 24: what about using fetcher in case of using filter section?

korvin89 (changes requested):

src/components/Select/types.ts
```diff
@@ -27,7 +28,10 @@ export type SelectRenderOption<T> = (
     options: SelectRenderOptionViewParams,
 ) => React.ReactElement;
 
-export type SelectProps<T = any> = QAProps &
+export type SelectOptionType<T> = SelectOption<T> | SelectOptionGroup<T>;
+export type SelectAsyncOptionType<Option, Pagination> = Fetcher<Option, Pagination>;
```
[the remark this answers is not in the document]
imechoim: Due to huge amount of Select logic I think that such abstraction is not bad idea.
When you read the Select.d.ts file, this name will be nice indicator of hidden abilities.

src/components/Select/wrappers/SelectAsync/SelectAsync.tsx
imechoim: You should pass `onFilterChange` callback and recreate `fetcher` function while search pattern changes.
I can write ShowCase for this scenario.

src/components/Select/types.ts
imechoim: I don't think that this is good idea. It will just extends number of props and introduce ambiguities. You can return initialOptions in async function you pass to option.

We can highlight it in additional story.

PR #699, opened 2023-05-31
feat(Icon): add data uri support

3y3:


src/components/Icon/Icon.tsx
```diff
@@ -104,6 +109,10 @@ export function Icon({
         ...a11yHiddenSvgProps,
     };
 
+    if (isDataUriSvgData(data)) {
+        return <img {...props} src={data} />;
```
ogonkov, line 113: `alt` is mandatory

ogonkov (changes requested):

src/components/Icon/Icon.tsx
3y3: Any idea, what we can write in alt?

ogonkov: Empty string i guess

amje, line 113: We should remove svg-specific props from here

amje (changes requested):

amje:
@3y3 Is this still relevant?

PR #707, opened 2023-06-05
draft: refactor: transfer common constants to fixtures

elemeNtk0:


ogonkov (changes requested):
@elemeNtk0  DATAUI-1639 is not a library component and should not be used as susch

amje:
@elemeNtk0 We decided not to add these constants, it adds redundant abstractions

elemeNtk0:
> @elemeNtk0 We decided not to add these constants, it adds redundant abstractions

What is dead may never die (:

PR #727, opened 2023-06-09
feat(Select): design improvements (control & options)

korvin89:
DATAUI-1626


src/components/Select/components/SelectControl/SelectControl.scss
```diff
@@ -1,28 +1,98 @@
...
+    transition: transform 0.1s ease-out, color 0.15s linear, background-color 0.15s linear;
+
+    &_view_clear {
+        border-color: transparent;
+    }
+
+    &_size_s {
+        @include block_style(s);
+        --_--select-border-radius: var(--yc-border-radius-s);
+        --_--select-options-text-left-padding: #{select-css-variables.$s-hor-padding};
```
amje, line 49: Why it's outside the mixin?

src/components/Select/components/SelectControl/SelectControl.scss
```diff
@@ -34,6 +104,7 @@ $buttonBlock: '.#{variables.$ns}button';
     &__placeholder,
     &__option-text {
         @include mixins.overflow-ellipsis();
+        padding-right: var(--_--select-options-text-left-padding);
```
amje, line 107: Naming is a bit off, left used for right

src/components/Select/components/SelectList/SelectList.scss
```diff
@@ -42,29 +32,34 @@ $xl-right-padding: '12px';
...
+            height: #{variables.$s-height};
+            padding: #{select-css-variables.$s-hor-padding} #{select-css-variables.$s-hor-padding}
+                4px;
         }
 
         #{$block}_size_m &:not(#{&}_empty) {
-            height: #{$m-height};
-            padding: 8px #{$m-right-padding} 4px #{$m-left-padding};
+            height: #{variables.$m-height};
+            padding: #{select-css-variables.$m-hor-padding} #{select-css-variables.$m-hor-padding}
```
amje, line 42: Using same var for both top and side padding can be misleading for other contributors

src/components/Select/select-css-variables.scss
```diff
@@ -0,0 +1,4 @@
+$s-hor-padding: 8px;
```
amje, line 1: Why not call the file just `variables.scss` and use it with alias?

src/components/variables.scss
```diff
@@ -1,3 +1,7 @@
 $ns: 'yc-';
 $ns-new: 'g-';
 $root: '.#{$ns}root';
+$s-height: 24px;
+$m-height: 28px;
+$l-height: 36px;
+$xl-height: 44px;
```
amje, line 7: Do we actually need these as global?

amje (changes requested):

amje (approved):

PR #728, opened 2023-06-14
fix(Label): fix rounded style

goshander:


amje (approved):

ogonkov:
@amje this is breaking change for us

PR #730, opened 2023-06-15
feat: rtl

dividead:


ogonkov:
@dividead may be we could use postcss-env instead?

amje:
Closing this, because it has been implemented in next branch

PR #737, opened 2023-06-19
fix(changelogDialog): optional date, reactNode description

Feverqwe:


amje:
@Feverqwe Could you open PR into [components](https://github.com/gravity-ui/components/tree/main/src/components/ChangelogDialog) repo? `ChangelogDialog` is in transition right now, and will be deleted from uikit soon

PR #749, opened 2023-06-21
chore: remove changelog from safe editing

amje:


CODEOWNERS
```diff
@@ -37,4 +37,3 @@
 # Allow everyone to publish releases
```
ValeraS, line 37: Is this comment no longer valid?

amje: Changed the comment

PR #761, opened 2023-06-23
feat(Select): loading indicator

imechoim:


src/components/Select/components/SelectList/SelectList.tsx
```diff
@@ -21,8 +22,11 @@ type SelectListProps = {
     flattenOptions: FlattenOption[];
     multiple?: boolean;
     virtualized?: boolean;
+    loading?: boolean;
 };
 
+const loadingOption = {value: 'LOADING', disabled: true};
```
amje, line 28: `LOADING` can conflict with user values, need more specific, Symbols are perfect for this

src/components/Select/components/SelectLoadingIndicator/SelectLoadingIndicator.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+import {Loader} from '../../../Loader/Loader';
+import {selectBlock} from '../../constants';
+
+import './SelectLoadingIndicator.scss';
+
+export const SelectLoadingIndicator = (props: {onRendered?: () => void}) => {
+    React.useEffect(() => {
+        props.onRendered?.();
```
amje, line 10: What is this for? It's not used

amje (changes requested):

src/components/Select/components/SelectList/SelectList.tsx
imechoim: In case of using `Symbol` we should extend `option.value` type. Is this modification necessary?

amje: No, let's just use more unique string for this

korvin89 (approved):

PR #771, opened 2023-06-28
feat(Card): added accessibility via keyboard

KirillDyachkovskiy:
UXRFC-155


src/components/Card/Card.scss
```diff
@@ -9,14 +9,21 @@ $block: '.#{variables.$ns}card';
...
-            --yc-card-box-shadow: 0px 3px 10px var(--yc-color-sfx-shadow);
-            cursor: pointer;
+        &#{$block}_clickable {
+            &:hover {
+                --yc-card-box-shadow: 0px 3px 10px var(--yc-color-sfx-shadow);
+                cursor: pointer;
+            }
+
+            &:focus-within {
+                box-shadow: 0 0 0 2px var(--yc-color-line-misc);
```
amje, line 25: Let's use pseudo element for outline, box-shadow is conflicting with own shadow

src/components/Card/Card.scss
```diff
@@ -34,13 +41,19 @@ $block: '.#{variables.$ns}card';
...
-            &::before {
-                content: '';
-                border: 2px solid var(--yc-color-line-selection-hover);
+                &::before {
+                    content: '';
+                    border: 2px solid var(--yc-color-line-selection-hover);
+                }
+            }
+
+            &:focus-within {
```
amje, line 55: Let's use the same technic used in the Button component, mouse click shouldn't trigger outline

amje (changes requested):

src/components/Card/Card.tsx
```diff
@@ -77,7 +97,10 @@ export const Card = React.forwardRef<HTMLDivElement, CardProps>(function Card(pr
                 },
                 className,
             )}
-            onClick={isClickable ? onClick : undefined}
+            onClick={handleClick}
+            onKeyDown={handleKeyDown}
```
ValeraS, line 101: Let's use button tag for interactive card to get rid of this handler and eslint errors. And add `role="button"`.

ValeraS (changes requested):

src/components/Card/Card.scss
```diff
@@ -9,14 +9,41 @@ $block: '.#{variables.$ns}card';
...
+        &#{$block}_clickable {
             cursor: pointer;
+            position: relative;
+
+            &:hover {
+                --yc-card-box-shadow: 0px 3px 10px var(--yc-color-sfx-shadow);
+            }
+
+            &:focus,
+            &:focus-visible {
```
amje, line 37: `:focus-visible` selector is not needed here, we use only for hiding box-shadow

src/components/Card/Card.scss
```diff
@@ -34,13 +61,36 @@ $block: '.#{variables.$ns}card';
...
+        &#{$block}_clickable {
             cursor: pointer;
 
-            &::before {
-                content: '';
-                border: 2px solid var(--yc-color-line-selection-hover);
+            &:hover {
+                &:not(:focus-visible)::before {
+                    content: '';
+                    border: 2px solid var(--yc-color-line-selection-hover);
```
amje, line 80: This chunk of code handles selection hover and should not be affected by focus

src/components/Card/Card.scss
```diff
@@ -34,13 +61,36 @@ $block: '.#{variables.$ns}card';
             pointer-events: none;
         }
 
-        &#{$block}_clickable:hover {
-            border-color: transparent;
```
amje, line 38: Seems like deleted by accident

src/components/Card/Card.scss
```diff
@@ -34,13 +61,36 @@ $block: '.#{variables.$ns}card';
...
-                border: 2px solid var(--yc-color-line-selection-hover);
+            &:hover {
+                &:not(:focus-visible)::before {
+                    content: '';
+                    border: 2px solid var(--yc-color-line-selection-hover);
+                }
+            }
+
+            &:focus,
+            &:focus-visible {
```
amje, line 85: Same here, no need in this line

amje (changes requested):

amje (approved):

src/components/Card/Card.tsx
```diff
@@ -60,11 +66,25 @@ export const Card = React.forwardRef<HTMLDivElement, CardProps>(function Card(pr
     /* View only with type 'container' and 'selection' */
     const defaultView = isTypeContainer || isTypeSelection ? 'outlined' : undefined;
 
+    const handleClick = isClickable ? onClick : undefined;
+
+    const handleKeyDown: React.KeyboardEventHandler<HTMLDivElement> = (event) => {
+        switch (event.key) {
+            case KeyCode.SPACEBAR:
```
ogonkov, line 73: We used to have hook for this, let's restore it like this, and use it instead
```shell
git co 907cadcd2a6750ccfc38f8a2642f954b0b759851 -- src/components/utils/useActionHandlers.ts
```

src/components/Card/Card.tsx
```diff
@@ -60,11 +66,25 @@ export const Card = React.forwardRef<HTMLDivElement, CardProps>(function Card(pr
...
+                event.preventDefault();
+
+                if (isClickable) {
+                    cardRef.current?.click();
```
ogonkov, line 77: why not invoke onClick handler?

amje: For correct event object I suppose

ValeraS (approved):

Lunory (approved):

PR #796, opened 2023-07-05
feat: add Disclosure component

Raubzeug:


src/components/Disclosure/Disclosure.scss
```diff
@@ -0,0 +1,81 @@
+@use '../variables';
+@use '../../../styles/mixins';
+
+$block: '.#{variables.$ns}disclosure';
```
amje, line 4: `$new-ns`

src/components/Disclosure/index.ts
```diff
@@ -0,0 +1,6 @@
+import {Disclosure, DisclosureComposition, DisclosureProps} from './Disclosure';
+
+export const ExportedDisclosure =
```
amje, line 3: What this type for?

src/components/Disclosure/Disclosure.tsx
```diff
@@ -0,0 +1,110 @@
...
+    Summary: typeof DisclosureSummary;
+}
+
+export interface DisclosureProps {
+    /** Disclosure size */
+    size: DisclosureSize;
+    /** Disabled state */
+    disabled?: boolean;
+    /** Default opening state */
+    defaultExpanded?: boolean;
```
amje, line 27: Component has only uncontrolled mode?

src/components/Disclosure/__stories__/Disclosure.stories.scss
```diff
@@ -0,0 +1,5 @@
+.label-stories {
```
amje, line 1: label?

amje (changes requested):

src/components/Disclosure/index.ts
Raubzeug: Thanks, it is really redundant by now.

src/components/Disclosure/Disclosure.tsx
Raubzeug: For now, yes. It takes initial condition and after that handles state on its own.

korvin89: It is a widely used functional I suppose. Could u add this props?

src/components/Disclosure/Disclosure.tsx
```diff
@@ -0,0 +1,110 @@
...
+        const {
+            size = 'xl',
+            disabled = false,
+            defaultExpanded = false,
+            arrowPosition = 'left',
+            summary = '',
+            className,
+            keepMounted = true,
+            children,
+            onChange = () => {},
```
korvin89, line 56: Why this mocked function? We can write smth like `onChange?.(...)`

src/components/Disclosure/DisclosureContext.tsx
```diff
@@ -0,0 +1,73 @@
...
+            </DisclosureToggleContext.Provider>
+        </DisclosureAttributesContext.Provider>
+    );
+}
+
+export function useDisclosureAttributes() {
+    const state = React.useContext(DisclosureAttributesContext);
+
+    if (state === undefined) {
+        throw new Error('useIsDisclosureExpanded must be used within DisclosureProvider');
```
korvin89, line 59: Maybe _useDisclosureAttributes_ ?

src/components/Disclosure/DisclosureDetails/DisclosureDetails.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+
+import {CSSTransition} from 'react-transition-group';
+
+import {getCSSTransitionClassNames} from '../../utils/transition';
+import {b} from '../Disclosure';
```
korvin89, line 6: It is not good that we have bidirectional relation here. Children of `Disclosure` should not import anything from it parent except types. It can generate circular dependency

src/components/Disclosure/DisclosureSummary/DisclosureSummary.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+
+import {ArrowToggle} from '../../ArrowToggle';
+import {DisclosureSize, b} from '../Disclosure';
```
korvin89, line 4: - Here the same about circular denepdency
- It would be better to import types with keyword `type`

korvin89 (changes requested):
I think we should add tests for basic functionality

src/components/Disclosure/Disclosure.tsx
```diff
@@ -0,0 +1,110 @@
...
+    /** Content summary */
+    summary?: string;
+    /** Class name */
+    className?: string;
+    /** Content */
+    children?: React.ReactNode;
+    /** Keep content in DOM */
+    keepMounted?: boolean;
+    /** Callback fired when the expand/collapse state is changed  */
+    onChange?: (event: React.SyntheticEvent, expanded: boolean) => void;
```
korvin89, line 39: I think we should use object as an argument. It will be more convenient and expandable

src/components/Disclosure/Disclosure.tsx
```diff
@@ -0,0 +1,110 @@
...
+}
+
+const isDisclosureSummaryComponent = isOfType(DisclosureSummary);
+
+// @ts-ignore this ts-error is appears when forwarding ref. It complains that DisclosureComposition props is not provided initially
+export const Disclosure: React.FunctionComponent<DisclosureProps> & DisclosureComposition =
+    React.forwardRef<HTMLDivElement, DisclosureProps>(function Disclosure(props, ref) {
+        const {
+            size = 'xl',
```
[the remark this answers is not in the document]
Raubzeug: I did it because in previous version of the component XL was default. But here is a new story, and M is better!

src/components/Disclosure/Disclosure.tsx
Raubzeug: I did it wilfully, because I wanted all properties in `DisclosureProvider` to be required. This will ensure that if new high-level props is added, developer will not forget to pass it to provider. And also I did't want to bloat `Omit` part in `DisclosureProviderProps`, so this mock has been added.

korvin89 (approved):

src/components/Disclosure/DisclosureContext.tsx
```diff
@@ -21,12 +24,22 @@ export const DisclosureToggleContext = React.createContext<
...
+        onChange,
+        expanded: controlledExpanded,
+    } = props;
     const [expanded, setExpanded] = React.useState(() => Boolean(defaultExpanded));
+    const controlledMode = controlledExpanded !== undefined;
 
     const handleToggle = (event: React.SyntheticEvent) => {
         setExpanded((prev) => !prev);
-        onChange(event, !expanded);
+        onChange({event, expanded: controlledMode ? undefined : !expanded});
```
amje, line 42: 1) Why in controlled mode `expanded` is undefined? It should always be defined
2) Object argument is not convenient in use, I would leave 2 arguments but swap them. Then basic usage will be like this:
```jsx
const [expanded, setExpanded] = React.useState(true);

<Disclosure expanded={expanded} onChange={setExpanded}/>
```

What do you think?

src/components/Disclosure/Disclosure.scss
```diff
@@ -6,59 +6,50 @@ $trigger: #{$block}__trigger;
...
         align-items: center;
-        &_disabled {
-            color: var(--yc-color-text-secondary);
-        }
+        font-size: inherit;
+        line-height: inherit;
+        font-weight: inherit;
+    }
+    &__trigger_disabled {
+        color: var(--yc-color-text-secondary);
```
amje, line 39: We also should show the default cursor instead of "pointer"

src/components/Disclosure/Disclosure.scss
```diff
@@ -6,59 +6,50 @@ $trigger: #{$block}__trigger;
...
 #{$block} {
-    #{$trigger} {
+    &_enter_active {
+        animation-name: #{variables.$ns}disclosure-expanded;
+        animation-duration: 0.5s;
+    }
+    &_exit_active {
+        display: block;
+        animation-name: #{variables.$ns}disclosure-collapsed;
+        animation-duration: 0.1s;
```
amje, line 16: Current animation looks like a FOUT (Flash Of Unstyled Text). IMO better without it completely

amje (changes requested):

src/components/Disclosure/DisclosureContext.tsx
Raubzeug: in `onChange` callback new value of `expanded` should be passed. But in controlled mode we don't know this new value. And that's why I passed `undefined` to it.

amje: In controlled mode we pass next value. That’s how in works in form controls

Raubzeug: But how can we find out the next value?..

amje: It is the opposite of current value, we pass the value we want to be next in onChange handle

src/components/Disclosure/Disclosure.scss
Raubzeug: Discussed with Yuriy and made some changes. Could you please take a look at the last storybook?

amje (approved):

PR #802, opened 2023-07-07
Stories media-block heating

GermanVor:


GermanVor:
I suggest add heating for medias to avoid blinking while changing news slides

PR #821, opened 2023-07-17
feat(TextInput): add an option to display error via tooltip

ilya-878:
1. Add 3 new props for TextInput component (errorMessage, errorPlacement, validationState) and 2 new props for TextArea component (errorMessage, validationState).
2. Create errorPropsMapper function that updates new props based on the deprecated prop (error).
3. Add error-in-tooltip feature for TextInput component.
4. Write few tests for new props for both components.
5. Add few error examples to the TextInputShowcase.
6. Mark 'error' prop as deprecated.


src/components/controls/TextArea/__tests__/TextArea.test.tsx
```diff
@@ -84,6 +69,65 @@ describe('TextArea', () => {
         });
     });
 
+    describe('error', () => {
+        test('render error message with error prop', () => {
+            const {container} = render(<TextArea error="Some Error" />);
+
+            // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
```
korvin89, line 76: Let`s rewrite this with [this](https://testing-library.com/docs/queries/bytext/) helper

src/components/controls/TextArea/__tests__/TextArea.test.tsx
```diff
@@ -84,6 +69,65 @@ describe('TextArea', () => {
...
+            expect(container.querySelector('.g-text-area__error')).toBeInTheDocument();
+            expect(screen.getByText('Some Error with errorMessage prop')).toBeVisible();
+        });
+
+        test('do not show error without error/errorMessage prop', () => {
+            const {container} = render(<TextArea />);
+
+            // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
+            expect(container.querySelector('.g-text-area__error')).not.toBeInTheDocument();
```
korvin89, line 95: U can use `data-qa` attributes for such purposes.
[How to set](https://github.com/gravity-ui/uikit/blob/main/src/components/Select/components/SelectControl/SelectControl.tsx#L139)
[How to use](https://github.com/gravity-ui/uikit/blob/main/src/components/Select/__tests__/Select.base-actions.test.tsx#L35)

src/components/controls/TextInput/TextInput.scss
```diff
@@ -129,6 +133,10 @@ $block: '.#{variables.$ns}text-input';
                 padding-right: 1px;
             }
 
+            #{$block}__error-icon {
```
korvin89, line 136: Let's use inner variable smth like `--_--text-input-error-icon-padding` instead of this styles

src/components/controls/TextInput/__tests__/TextInput.input.test.tsx
```diff
@@ -85,6 +70,94 @@ describe('TextInput input', () => {
             });
         });
 
+        describe('error', () => {
+            test('render error message with error prop', () => {
+                const {container} = render(<TextInput error="Some Error" />);
+
+                // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
```
korvin89, line 77: Here the same issues as above in `TextArea`

src/components/controls/utils.ts
```diff
@@ -11,7 +11,36 @@ export const prepareAutoComplete = (
...
+        'error' | 'errorMessage' | 'errorPlacement' | 'validationState'
+    >,
+) => {
+    const {error, errorMessage, errorPlacement, validationState} = errorProps;
+
+    let errorMessageProp: typeof errorMessage;
+    if (typeof error === 'string') {
+        errorMessageProp = error;
+    }
+    if (errorMessage || errorMessage === '') {
```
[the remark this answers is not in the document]
ilya-878: It definitely is, but what is wrong with this if statement?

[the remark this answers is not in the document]
korvin89: U can use `typeof` here

src/components/controls/TextArea/TextArea.tsx
```diff
@@ -57,16 +64,23 @@ export const TextArea = React.forwardRef<HTMLSpanElement, TextAreaProps>(functio
         onUpdate,
         onChange,
     } = props;
+
+    const {errorMessageProp, validationStateProp} = errorPropsMapper({
```
amje, line 68: It make sense to invert names here, function will accept all props and will return `errorMessage` and `validationState` without *Prop

src/components/controls/TextArea/TextArea.tsx
```diff
@@ -57,16 +64,23 @@ export const TextArea = React.forwardRef<HTMLSpanElement, TextAreaProps>(functio
...
     const [hasVerticalScrollbar, setHasVerticalScrollbar] = React.useState(false);
-    const state = getInputControlState({error});
+    const state = getInputControlState(validationStateProp);
     const handleRef = useForkRef(props.controlRef, innerControlRef);
     const innerId = useUniqId();
 
     const isControlled = value !== undefined;
     const inputValue = isControlled ? value : uncontrolledValue;
-    const isErrorMsgVisible = typeof error === 'string';
+    const isErrorMsgVisible = Boolean(errorMessageProp);
```
amje, line 83: If validationState is not `invalid` we shouldn't show the error

src/components/controls/types.ts
```diff
@@ -29,8 +31,16 @@ export type BaseInputControlProps<T = Element> = DOMProps &
...
         disabled?: boolean;
-        /** Shows error state and optional message if property identified as a string */
+        /** Shows error state and optional message if property identified as a string
+         * @deprecated Prop `error` has a lower priority than `errorMessage`. Use `errorMessage` instead
+         */
         error?: string | boolean;
+        /** Determines content of the error message */
+        errorMessage?: React.ReactNode;
+        /** Determines whether the error message will be placed under the input field as text or in the tooltip */
+        errorPlacement?: 'text' | 'tooltip';
```
amje, line 41: Let's change the values to `outside | inside`

amje (changes requested):

src/components/controls/TextArea/TextArea.tsx
amje: Looks like it's still an issue

saxumcordis: my bad
fixed

mfyodorov (approved):

src/components/controls/TextArea/TextArea.tsx
amje: Couldn't reproduce in the storybook, but according to code it's not fixed, if `errorMessage` is passed then `validationState` will be `invalid`.

How it should work:
First we map old prop to new ones
* `error = true` => `errorMessage = undefined` and `validationState = 'invalid'`
* `error = 'Some error'` => `errorMessage = 'Some error'` and `validationState = 'invalid'`

Then we check if both 1) `errorMessage` has a value and 2) `validationState = invalid` then we display an error, otherwise we won't.

amje: @saxumcordis  I forgot one statement, on [this](https://github.com/gravity-ui/uikit/pull/821/files#diff-e837ab9b77ac6a096b85a7932f340c6ff8419adb82f50c615ca6c52ef2e480a2R47) line you set `validationState = invalid` if `errorMessage` is not empty. New `errorMessage` prop should not set `validattionState`, only old `error` prop should. Also why empty strings in errors are considered as values? Empty string should be equal to undefined

saxumcordis: okay, got u

saxumcordis: done

amje (approved):

PR #825, opened 2023-07-17
fix(Card): prevent keyboard handler form abort

### LINKED ISSUE #822: Preventing of any events with no reason from Card children
it seems like this line should be fired conditionally:
https://github.com/gravity-ui/uikit/pull/771/files#diff-0dd621c2c92f4f42ea8884560f43696fd4e03204717f28a1a144aeb249de808dR19

because right now it intercept and prevent any propagated event from Card children

ogonkov:
Fixes #822

### DIFF
--- a/src/components/Card/__tests__/Card.test.tsx
+++ b/src/components/Card/__tests__/Card.test.tsx
@@ -210,16 +210,22 @@ describe('Card', () => {
     test('ignore onClick if type!="container"', async () => {
         const user = userEvent.setup();
         const handleOnClick = jest.fn();
+        const outerHandler = jest.fn();
         render(
-            <Card type="container" onClick={handleOnClick} qa={qaId}>
-                {cardText}
-            </Card>,
+            <div onClick={outerHandler} onKeyDown={outerHandler}>
+                <Card type="container" onClick={handleOnClick} qa={qaId}>
+                    {cardText}
+                </Card>
+            </div>,
         );
 
         const card = screen.getByText(cardText);
 
         await user.click(card);
-        expect(handleOnClick).toHaveBeenCalledTimes(0);
+        await user.keyboard('[Space]');
+        await user.keyboard('[Enter]');
+        expect(handleOnClick).not.toHaveBeenCalled();
+        expect(outerHandler).toHaveBeenCalled();
     });
 
     describe('should respond to clicks correctly', () => {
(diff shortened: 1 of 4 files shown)


src/components/Card/__tests__/Card.test.tsx
```diff
@@ -210,16 +210,22 @@ describe('Card', () => {
...
         );
 
         const card = screen.getByText(cardText);
 
         await user.click(card);
-        expect(handleOnClick).toHaveBeenCalledTimes(0);
+        await user.keyboard('[Space]');
+        await user.keyboard('[Enter]');
+        expect(handleOnClick).not.toHaveBeenCalled();
+        expect(outerHandler).toHaveBeenCalled();
```
ValeraS, line 228: What does it test? Is active element inside `Сard` after click?

ValeraS: 1. onCLick handler is not called for Card with type container
2. onKeyDown is not fired on div element, because active element is not inside it.

> Test that we didnt prevent outside handlers

then check should be this `expect(outerHandler).toHaveBeenCalledTimes(3);`, right?

ogonkov: Yes, we can count invocations too

ogonkov: I've changed test to case that failed for us

src/components/utils/useActionHandlers.ts
```diff
@@ -15,10 +15,8 @@ interface UseActionHandlersResult<T> {
 export function useActionHandlers<T>(callback?: AnyFunction): UseActionHandlersResult<T> {
     const onKeyDown = React.useCallback(
         (event: React.KeyboardEvent<T>) => {
-            if (['Enter', ' ', 'Spacebar'].includes(event.key)) {
-                event.preventDefault();
-
-                return callback?.(event);
+            if (callback && ['Enter', ' ', 'Spacebar'].includes(event.key)) {
+                return callback(event);
```
Lunory, line 19: If callback exists should we use `event.preventDefault`?

ogonkov: Don't think so, this hook emulates handling click in controls like `<button>`, they didn't prevent events

ValeraS (approved):

Lunory (approved):

PR #830, opened 2023-07-19
feat(LayerManager): export getLayersCount

yndx-madfriend:
In order to create idle Popups in common, I need a way to determine in render-time whether current layer stack is empty (so that idle popup can start rendering). Even though I can subscribe to `layerschange` with event broker, there is no way to know if stack is initially empty.


yndx-madfriend:
An example of where this function is needed can be found [here](https://a.yandex-team.ru/review/4192907/details).

korvin89 (approved):

Issue #684, opened 2023-05-30
[a11y] get rid of onclicks on div in the Popup component


KirillDyachkovskiy:
It would be great to increase the accessibility of components from the keyboard and screen readers, for this you can get rid of onclicks on non-interactive elements

PR #834, opened 2023-07-20
feat(Popover): add props to improve accessibility

zamkovskaya:
https://github.com/gravity-ui/uikit/issues/691
https://github.com/gravity-ui/uikit/issues/684

### DIFF
--- a/src/components/utils/useRestoreFocus.tsx
+++ b/src/components/utils/useRestoreFocus.tsx
@@ -17,6 +17,7 @@ export function useRestoreFocus({enabled, restoreFocusRef, focusTrapped}: UseRes
         if (enabled && initialActiveElementRef.current === null) {
             initialActiveElementRef.current = event.relatedTarget as HTMLElement | null;
             lastActiveElementRef.current = initialActiveElementRef.current;
+            ref.current = (restoreFocusRef?.current || initialActiveElementRef.current) ?? null;
         }
     };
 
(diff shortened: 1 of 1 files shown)


src/components/Popover/README.md
```diff
@@ -38,6 +38,11 @@ Extends [`QAProps`](../types.ts).
...
+| focusTrap               | `Boolean`          |          |                                          | `false`             | Prevent focus from leaving the popover while open                                                                                                                              |
+| autofocus               | `Boolean`          |          |                                          | `false`             | If true, focus will be transferred to the first element when the popover opens                                                                                                 |
+| restoreFocusRef         | `React.RefObject`  |          |                                          |                     | Focused element when the popover closes open                                                                                                                                   |
```
kseniya57, line 45: There seems to be a typo in the description

kseniya57 (approved):

amje (changes requested):

src/components/Popup/Popup.tsx
```diff
@@ -139,23 +150,29 @@ export function Popup({
...
-                            />
-                        )}
-                        {children}
-                    </div>
+                    <FocusTrap enabled={focusTrap && open} disableAutoFocus={!autoFocus}>
+                        {/* The event handlers should only be used to capture bubbled events */}
+                        {/* eslint-disable-next-line jsx-a11y/no-static-element-interactions */}
+                        <div
+                            onClick={onClick}
+                            onKeyDown={onKeyDown}
```
amje, line 158: Didn't notice this at first. Why we're listening keydown?

zamkovskaya: The onClick event handler is currently used to execute an action within a popup when it's clicked. It returns true or false, which determines whether to close the popup. This behavior should also work for keyboard and screen reader users. I can't think of a specific use case for onClick at the moment, but in any case, if there's an action involved, it should be accessible to users who don't use a mouse.

amje: I think onClick here is by mistake. Popup should not handle clicks or keyboard events, it's utility component.

zamkovskaya: I marked onClick as deprecated and removed onKeyDown

PR #838, opened 2023-07-24
fix: add menu-item-content-children class name

GermanVor:


src/components/DropdownMenu/DropdownMenuItem.tsx
```diff
@@ -100,7 +100,9 @@ export const DropdownMenuItem = <T,>({
                 onClick={handleMenuItemClick}
             >
                 <div className={cnDropdownMenu('menu-item-content')}>
-                    <div>{text || children}</div>
+                    <div className={cnDropdownMenu('menu-item-content-children')}>
```
GermanVor, line 103: without this class name `div` will have width equal `text || children` width.
this is unacceptable if we want all items in DropDown have same width.

for exemple 

image

axtk (approved):

PR #856, opened 2023-08-02
feat: migrate from Popper to Floating UI

Handaz:


korvin89:
@Handaz Do you want to continue work in this PR?

PR #863, opened 2023-08-04
feat(Breadcrumbs): add `title` prop to item

polikashina:
Make available setting custom title for breadcrumb item

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.tsx
+++ b/src/components/Breadcrumbs/Breadcrumbs.tsx
@@ -16,6 +16,7 @@ export interface BreadcrumbsItem {
     action: (event: React.MouseEvent<HTMLElement, MouseEvent> | KeyboardEvent) => void;
     href?: string;
     items?: BreadcrumbsItem[];
+    htmlTitle?: string
 }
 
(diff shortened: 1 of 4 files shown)


src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -16,6 +16,7 @@ export interface BreadcrumbsItem {
     action: (event: React.MouseEvent<HTMLElement, MouseEvent> | KeyboardEvent) => void;
     href?: string;
     items?: BreadcrumbsItem[];
+    htmlTitle?: string
```
amje, line 19: Lets name it just `title`, we don't use `html` prefix for such attributes

amje (changes requested):

src/components/Breadcrumbs/Breadcrumbs.tsx
polikashina: [done](https://github.com/gravity-ui/uikit/pull/863/commits/7f9b8dc7d9e82500764e2a01ccbc1d85e6b2b8c9#diff-d13783796f2d03ab1bf02433a6bb4b36d1b7a20e7b8d9a8f6f4ef07bc7ac33e7L19)

amje (approved):

PR #872, opened 2023-08-08
feat(Link): add visitable prop

helenjer:
Adds views for visited primary and secondary links
[UXRFC-163]


amje:
@helenjer What about instead of adding new views we add a new prop `visitable`?

LakeVostok (approved):

helenjer:
@amje changed to new `visitable` prop

amje (approved):

PR #879, opened 2023-08-09
feat(Toaster): make title optional

mournfulCoroner:


src/components/Toaster/Toast/Toast.tsx
```diff
@@ -118,7 +118,9 @@ export const Toast = React.forwardRef<HTMLDivElement, ToastUnitedProps>(function
                         <Icon data={Xmark} />
                     </Button>
                 )}
-                {Boolean(content) && <div className={b('content')}>{content}</div>}
+                {Boolean(content) && (
+                    <div className={b('content', {'without-title': !title})}>{content}</div>
```
ogonkov, line 122: Move title check to single var `hasTitle` please

```ts
const hasTitle = Boolean(title);
```

src/components/Toaster/Toast/Toast.scss
```diff
@@ -110,6 +110,10 @@ $block: '.#{variables.$ns}toast';
         line-height: 24px;
     }
 
+    &__content_without-title {
+        padding-right: 32px;
```
ogonkov, line 114: I guess that this value should be calculated, like we do it for `title` above

ogonkov (changes requested):

src/components/Toaster/Toast/Toast.tsx
mournfulCoroner: Should I do the same with `Boolean(content)`?

ogonkov: What alternative you thinking of?

ogonkov: This is used only once, this is not required i think, but ok

src/components/Toaster/Toast/Toast.tsx
```diff
@@ -118,7 +121,9 @@ export const Toast = React.forwardRef<HTMLDivElement, ToastUnitedProps>(function
                         <Icon data={Xmark} />
                     </Button>
                 )}
-                {Boolean(content) && <div className={b('content')}>{content}</div>}
+                {hasContent && (
+                    <div className={b('content', {'without-title': !title})}>{content}</div>
```
ogonkov, line 125: `{'without-title': !hasTitle}`

src/components/Toaster/__stories__/ToasterShowcase.tsx
```diff
@@ -292,6 +303,12 @@ export const ToasterDemo = ({
         </Button>
     );
 
+    const withoutTitleToastBtn = (
+        <Button view="outlined" size="l" onClick={createWithoutTitleToast} style={btnStyle}>
+            Create toast without title
+        </Button>
+    );
```
ogonkov, line 310: We have option to add content for Toast, it is better to add same option for title, instead of another button

ogonkov (approved):

PR #882, opened 2023-08-10
fix(Toaster): add min-height

mournfulCoroner:


src/components/Toaster/Toast/Toast.scss
```diff
@@ -81,6 +81,7 @@ $block: '.#{variables.$ns}toast';
         row-gap: 8px;
         width: 100%;
         height: 100%;
+        min-height: 18px;
```
ogonkov, line 84: It is line-height?

ogonkov (approved):

PR #895, opened 2023-08-14
feat: add note prop into TextArea and withCounter hoc

Zveroboev:


src/components/controls/TextArea/TextArea.scss
```diff
@@ -48,10 +48,28 @@ $block: '.#{variables.$ns-new}text-area';
...
+        margin-left: auto;
+    }
+
     &__error {
         @include mixins.text-body-1();
 
         color: var(--g-color-text-danger);
+
+        &:not(:last-child) {
+            margin-right: 8px;
```
ogonkov, line 67: May be we can use `var(--g-spacing-2)`?

src/components/controls/TextArea/TextArea.tsx
```diff
@@ -156,7 +159,10 @@ export const TextArea = React.forwardRef<HTMLSpanElement, TextAreaProps>(functio
                     />
                 )}
             </span>
-            {isErrorMsgVisible && <div className={b('error')}>{error}</div>}
+            <div className={b('additional')}>
```
ogonkov, line 162: Should it be rendered if both error and note are empty?

src/components/controls/TextArea/__tests__/TextArea.test.tsx
```diff
@@ -23,6 +23,16 @@ describe('TextArea', () => {
             expect(screen.getByText('Some Error')).toBeVisible();
         });
 
+        test('render note container with note prop', () => {
+            const {container} = render(
+                <TextArea error="Some Error" note={<div>Additional</div>} />,
+            );
+
+            // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
+            expect(container.querySelector('.g-text-area__note')).toBeInTheDocument();
```
ogonkov, line 32: Why check for element existence? Isn't it the same element that you check next?

src/components/controls/TextArea/hoc/withCounter.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+
+import {Text} from '../../../Text';
+import {getComponentName} from '../../../utils/getComponentName';
+import {TextArea, TextAreaProps} from '../TextArea';
+
+export interface WithCounterProps {
+    maxLength: number;
```
ogonkov, line 8: This name is looks like something related to input width

I guess something like `counterMax` would be better

src/components/controls/TextArea/__tests__/TextArea.test.tsx
Zveroboev: Did the same test with the error container

image

korvin89: I think getByText is quite enough

src/components/controls/TextInput/TextInput.scss
```diff
@@ -91,10 +91,28 @@ $block: '.#{variables.$ns}text-input';
         flex-shrink: 0;
     }
 
+    &__additional {
```
korvin89, line 94: We already have `additional-content`. It is a little bit confusing. Maybe smth like `outer-additional-content` ?

src/components/controls/TextArea/__tests__/TextArea.test.tsx
```diff
@@ -16,13 +16,19 @@ describe('TextArea', () => {
         });
 
         test('render error message with error prop', () => {
-            const {container} = render(<TextArea error="Some Error" />);
+            render(<TextArea error="Some Error" />);
 
             // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
-            expect(container.querySelector('.g-text-area__error')).toBeInTheDocument();
```
korvin89, line 22: Here and below: there is no need in this comment more

amje (approved):

PR #900, opened 2023-08-16
feat(Popover): accept function as a children

Kyzyl-ool:
- Now Popover can accept function as children.
- Fixed bug in the [Popover](https://preview.gravity-ui.com/uikit/?path=/story/components-popover--accessible) (when focused on Popover button, unable to open tooltip with Enter)

### DIFF
--- a/src/components/Link/Link.scss
+++ b/src/components/Link/Link.scss
@@ -13,10 +13,6 @@ $block: '.#{variables.$ns}link';
         outline: 2px solid var(--g-color-line-focus);
     }
 
-    &:focus:not(:focus-visible) {
-        outline: 0;
-    }
-
     &_view_normal,
     &_view_normal-visitable {
         color: var(--g-color-text-link);
--- a/src/components/Link/Link.scss
+++ b/src/components/Link/Link.scss
@@ -12,6 +12,9 @@ $block: '.#{variables.$ns}link';
     &:focus {
         outline: 2px solid var(--g-color-line-focus);
     }
+    &:focus:not(:focus-visible) {
+        outline: 0;
+    }
 
     &_view_normal,
     &_view_normal-visitable {
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -57,6 +57,7 @@ export const Popover = React.forwardRef<PopoverInstanceProps, PopoverProps & QAP
         focusTrap,
         autoFocus,
         restoreFocusRef,
+        enableClickEmulationByKeyboard = true,
     },
     ref,
 ) {
@@ -209,6 +210,7 @@ export const Popover = React.forwardRef<PopoverInstanceProps, PopoverProps & QAP
                 disabled={disabled}
                 onClick={onClick}
                 closedManually={closedManually}
+                enableClickEmulationByKeyboard={enableClickEmulationByKeyboard}
             >
                 {children}
             </Trigger>
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -57,7 +57,7 @@ export const Popover = React.forwardRef<PopoverInstanceProps, PopoverProps & QAP
         focusTrap,
         autoFocus,
         restoreFocusRef,
-        clickEmulationKeys = [],
+        enableButtonEmulation = true,
     },
     ref,
 ) {
@@ -210,7 +210,7 @@ export const Popover = React.forwardRef<PopoverInstanceProps, PopoverProps & QAP
                 disabled={disabled}
                 onClick={onClick}
                 closedManually={closedManually}
-                clickEmulationKeys={clickEmulationKeys}
+                enableButtonEmulation={enableButtonEmulation}
             >
                 {children}
             </Trigger>
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -57,7 +57,7 @@ export const Popover = React.forwardRef<PopoverInstanceProps, PopoverProps & QAP
         focusTrap,
         autoFocus,
         restoreFocusRef,
-        enableClickEmulationByKeyboard = false,
+        clickEmulationKeys = [],
     },
     ref,
 ) {
@@ -210,7 +210,7 @@ export const Popover = React.forwardRef<PopoverInstanceProps, PopoverProps & QAP
                 disabled={disabled}
                 onClick={onClick}
                 closedManually={closedManually}
-                enableClickEmulationByKeyboard={enableClickEmulationByKeyboard}
+                clickEmulationKeys={clickEmulationKeys}
             >
                 {children}
             </Trigger>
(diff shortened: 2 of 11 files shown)


src/components/utils/useActionHandlers.ts
```diff
@@ -10,16 +12,22 @@ interface UseActionHandlersResult<T> {
 /**
  * Emulates behaviour of system controls, that respond to Enter and Spacebar
  * @param callback
+ * @param keyCodes
  * @return {onKeyDown}
  */
-export function useActionHandlers<T>(callback?: AnyFunction): UseActionHandlersResult<T> {
+export function useActionHandlers<T>(
+    callback?: AnyFunction,
+    keyCodes: EKeyCode[] = [EKeyCode.ENTER, EKeyCode.SPACEBAR, EKeyCode.SPACEBAR_OLD],
```
amje, line 20: This hook is only for handling "action" (alternative for click on keyboard), therefore we should handle only Enter and Spacebar. No need in such option

src/components/constants.ts
```diff
@@ -1,8 +1,17 @@
-export const KeyCode = {
+export enum EKeyCode {
```
amje, line 1: Why introduce this enum?

amje (changes requested):

src/components/utils/useActionHandlers.ts
Kyzyl-ool: The reason why I added `keyCodes` arg – <a> tag reacts to the Enter by default and browser triggers click event. Thus, one should call this hook with `EKeyCode.SPACEBAR`, `EKeyCode.SPACEBAR_OLD`. For all other elements (except for <button>) one should `useActionHandlers` for Spacebar and Enter

src/components/constants.ts
Kyzyl-ool: Because I needed to use KeyCode keys as type. Something like `keyof typeof KeyCode` seemed complex

src/components/utils/useActionHandlers.ts
Kyzyl-ool: And I need this function in another project with `<a>` tags)

Kyzyl-ool: Maybe one should restrict this parameter only for `EKeyCode.ENTER`, `EKeyCode.SPACEBAR` and `EKeyCode.SPACEBAR_OLD`?

amje: Why do you need this hook for `<a>` tags in the first place? Seems incorrect usage. The only purpose of this hook is to add keyboard functionality of the button to another element.

Kyzyl-ool: To make `<a>` tags behave like buttons. But fine, really looks like incorrect usage as browsers consider Spacebar as page down action.

Kyzyl-ool: Reverted the changes for the hook

src/components/constants.ts
Kyzyl-ool: Reverted

src/components/Link/Link.scss
```diff
@@ -8,9 +9,7 @@ $block: '.#{variables.$ns}link';
     -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
     cursor: pointer;
 
-    &:focus {
-        outline: 2px solid var(--g-color-line-misc);
-    }
+    @include mixins.focusable();
```
amje, line 12: This mixin should also include the next piece of code, to be visible only for keyboard events

src/components/Popover/Popover.tsx
```diff
@@ -57,6 +57,7 @@ export const Popover = React.forwardRef<PopoverInstanceProps, PopoverProps & QAP
         focusTrap,
         autoFocus,
         restoreFocusRef,
+        enableButtonEmulation = true,
```
amje, line 60: Shouldn't be it always enabled?

src/components/Popover/components/Trigger/Trigger.tsx
```diff
@@ -81,7 +83,7 @@ export const Trigger = ({
         <div
```
amje, line 83: This div is not accessible from the keyboard, we need to add `tabIndex={0}` and focus styles

amje (changes requested):

src/components/Link/Link.scss
Kyzyl-ool: right, thanks

src/components/Popover/Popover.tsx
Kyzyl-ool: No, as Popover can wrap `<button>` element, which emulates click event on Enter pressing. After that the browser emulates click event, so no need to emulate click event twice

src/components/Popover/components/Trigger/Trigger.tsx
Kyzyl-ool: I think it is applicable only for cases when Trigger wraps non-interactive elements, not accessible from keyboard by default. Otherwise, if `<button>` is the child node, the Trigger itself will be focused first, then `<button>` element. Seems there is unnecessary focusable element.

Maybe one should pass `tabIndex` from props, at the developer's discretion?

amje: The problem is we have no control in trigger node. We will lose a11y if user won't provide proper element to it

src/components/Popover/Popover.tsx
Kyzyl-ool: reverted changes. There is no longer `enableButtonEmulation` prop

src/components/Popover/components/Trigger/Trigger.tsx
Kyzyl-ool: Accessible element should be inside the trigger element, but the trigger itself must not be, to avoid focus duplication

src/components/Link/Link.scss
```diff
@@ -1,4 +1,5 @@
 @use '../variables';
+@use '../../../styles/mixins';
```
amje, line 2: Not used

src/components/Popover/components/Trigger/Trigger.tsx
```diff
@@ -75,7 +81,9 @@ export const Trigger = ({
 
     const {onKeyDown} = useActionHandlers(handleClick);
 
-    return (
+    return typeof children === 'function' ? (
+        <React.Fragment>{children({onClick: handleClick, className, onKeyDown})}</React.Fragment>
```
amje, line 85: No need to wrap into fragment

amje (changes requested):

src/components/Link/Link.scss
Kyzyl-ool: Thanks

src/components/Popover/components/Trigger/Trigger.tsx
Kyzyl-ool: Sorry, there will be an error in Popover `undefined is not assignable to type Element | null`

amje (approved):

kseniya57 (approved):

amje:
@Kyzyl-ool Resolve the conflicts please

PR #915, opened 2023-08-18
feat(Button): move transformation to content

ogonkov:


amje:
Currently it looks a bit odd

amje:
@ogonkov Seems we're not planning to change the transformation. Can we close this PR?

PR #922, opened 2023-08-21
fix(Select): enable FocusTrap for filterable Select

ma-efremoff:


korvin89 (changes requested):
It is `bydesign` feature, we do not want to have focus trap in case of Select. it breaks user experience in case of native focus navigation by `tab` from control to control

ma-efremoff:
> It is `bydesign` feature, we do not want to have focus trap in case of Select. it breaks user experience in case of native focus navigation by `tab` from control to control

Without the changes it is impossible to filter items of a Select that embedded to a Modal. TextInput-filter control that placed above items cannot be focused because of enabled FocusTrap of underlying Modal component.

PR #923, opened 2023-08-21
fix(Table): fix copying the array from getRowClassNames

B0nBun:
Mutating `classNames` array with `push` can cause problems in cases where the `getRowClassNames` function isn't pure and gets it's return value from the outer scope.
e.g.
```ts
const classNames = ["foo-row"];
const getRowClassNames = () => classNames;

const TableWithSelection = withTableSelection(Table);

const Component = () => {
  // ...
  return <TableWithSelection {...props} getRowClassNames={getRowClassNames} />;
}
```
The example above results in rows having `row_selected` class even after being removed from `selectedIds`


amje:
Thx!

Issue #616, opened 2023-04-11
Table: content jumping in case of adaptive sizes


helenjer:
When table column sizes are set, width of columns is visually jumping: first render - arbitrary sizes, then right size from props. 
Reproduces on storybook "Adaptive" story: https://preview.gravity-ui.com/uikit/?path=/story/components-table--adaptive

helenjer:
CC codeowners: @krozhkov

PR #926, opened 2023-08-21
fix(Table): calculate column styles on mount

lunaticenslaved:
[issue](https://github.com/gravity-ui/uikit/issues/616)


src/components/Table/Table.tsx
```diff
@@ -259,6 +259,7 @@ export class Table<I extends TableDataItem = Record<string, string>> extends Rea
         });
 
         if (this.tableRef.current) {
+            this.updateColumnStyles();
```
ogonkov, line 262: why not in `constructor`?

ogonkov, line 262: I guess state could be updated without any conditions

PR #933, opened 2023-08-23
feat: create useViewportSize hook

NikitaCG:


amje (changes requested):

src/hooks/UseViewportSize/index.ts
```diff
@@ -0,0 +1,2 @@
+export {useViewportSize} from './UseViewportSize';
```
amje, line 1: Let's place this hook with others, inside utils directory, we will refactor directory structure later. Also the file should be named in lowercase

amje (changes requested):

src/hooks/UseViewportSize/index.ts
NikitaCG: fix

amje (approved):

PR #960, opened 2023-08-28
feat(List): improve a11y

### LINKED ISSUE #686: get rid of onclicks on div in the List component
It would be great to increase the accessibility of components from the keyboard and screen readers, for this you can get rid of onclicks on non-interactive elements

zamkovskaya:
Closes #686


src/components/List/components/ListItem.tsx
```diff
@@ -17,13 +17,24 @@ export class ListItem<T = unknown> extends React.Component<ListItemProps<T>> {
...
+            role = 'listitem',
+        } = this.props;
 
         return (
-            // eslint-disable-next-line jsx-a11y/click-events-have-key-events
             <div
-                role="listitem"
+                role={role}
+                tabIndex={
+                    (this.props.onClick || role === 'option') && !item.disabled ? 0 : undefined
```
amje, line 35: Item should never be tabbable

amje (changes requested):

src/components/List/components/ListItem.tsx
```diff
@@ -74,6 +88,12 @@ export class ListItem<T = unknown> extends React.Component<ListItemProps<T>> {
         });
     };
 
+    private onKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
+        if (this.props.onClick && [' ', 'Spacebar'].includes(event.key)) {
```
korvin89, line 92: These are breaking changes. Users handle this actions by themselves. And now for example `Select` fires double select action in case of using keyboard

src/components/List/List.tsx
```diff
@@ -201,11 +210,6 @@ export class List<T = unknown> extends React.Component<ListProps<T>, ListState<T
                 }
                 break;
             }
-            default: {
```
korvin89, line 204: These are breaking changes because after them the behavior that users expect will change

src/components/List/components/SimpleContainer.tsx
```diff
@@ -51,6 +51,7 @@ export class SimpleContainer extends React.Component<SimpleContainerProps, Simpl
                 ref.current.scrollIntoView?.({
                     block: 'nearest',
                 });
+                ref.current.focus();
```
korvin89, line 54: It as a breaking change too

src/components/Select/components/SelectList/SelectList.tsx
```diff
@@ -127,6 +138,8 @@ export const SelectList = React.forwardRef<List<FlattenOption>, SelectListProps>
             virtualized={virtualized}
             renderItem={renderItem}
             onItemClick={onOptionClick}
+            selectedItemIndex={selectedIndexes}
```
korvin89, line 141: It brakes styles for selected options in case of multiple choice

korvin89 (changes requested):

src/components/List/components/ListItem.tsx
ogonkov: I guess it should follow roving tabindex pattern

src/components/List/README.md
```diff
@@ -251,6 +251,9 @@ Likewise, you can forward `onFocus` and `onBlur` if you need to repeat the behav
...
 | itemsClassName    | Custom class name to be added to an item list                                                                                                                                                          | `String`          |         |
+| role              | HTML `role` attribute                                                                                                                                                                                  | `String`          | list    |
+| id                | HTML `id` attribute                                                                                                                                                                                    | `string`          |         |
+| onChangeActive    | Fires when the index of an option in the listbox, visually indicated as having keyboard focus, is changed. `(index?: number) => void`                                                                  | `Function`        |         |
```
korvin89, line 259: 1. Why we need this?
2. There is no availability to use focus for navigation among list items

src/components/List/List.tsx
```diff
@@ -98,20 +100,32 @@ export class List<T = unknown> extends React.Component<ListProps<T>, ListState<T
         if (this.props.activeItemIndex !== prevProps.activeItemIndex) {
             this.activateItem(this.props.activeItemIndex);
         }
+
+        if (this.props.onChangeActive && this.state.activeItem !== prevState.activeItem) {
```
korvin89, line 104: There is a better place for such logic - https://github.com/gravity-ui/uikit/blob/d51039bb0fd4129ac511681e7fa23f051d1c3397/src/components/List/List.tsx#L157

src/components/Select/Select.tsx
```diff
@@ -96,6 +106,8 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function
         onClose,
         onOpenChange,
     });
+    const uniqId = useUniqId();
+    const selectId = id ?? uniqId;
```
korvin89, line 110: Maybe smth like this?
```js
const selectId = id || useUniqId();
```
Remove extra line and change operator case of empty string does not a valid id

src/components/Select/Select.tsx
```diff
@@ -242,6 +254,9 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function
                 onKeyDown={handleControlKeyDown}
                 renderControl={renderControl}
                 value={value}
+                popupId={`select-popup-${selectId}`}
+                selectId={`select-${selectId}`}
+                activeIndex={activeIndex}
```
korvin89, line 259: I think we should add this properties to custom controls [here](https://github.com/gravity-ui/uikit/blob/d51039bb0fd4129ac511681e7fa23f051d1c3397/src/components/Select/components/SelectControl/SelectControl.tsx#L114)

src/components/utils/useSelect/useSelect.ts
```diff
@@ -60,13 +61,15 @@ export const useSelect = <T extends unknown>(props: UseSelectProps) => {
 
     return {
         value,
+        activeIndex,
```
korvin89, line 64: I think we need add description for new properties

korvin89 (commented):
Do we actually need [this logic](https://github.com/gravity-ui/uikit/blob/d51039bb0fd4129ac511681e7fa23f051d1c3397/src/components/Select/components/SelectList/OptionWrap.tsx#L29) if we use List selection api?

src/components/List/README.md
zamkovskaya: In this improvement, I am trying to follow the Select-Only Combobox pattern at https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/. When visual focus is in the the listbox, the DOM focus remains on the combobox, and the value of aria-activedescendant on the combobox is set to a value that refers to the listbox option that is visually indicated as focused. Therefore, we need a function to track the visually focused element in order to pass its id to aria-activedescendant. As a result, when navigating through the elements, a screen reader will announce the currently active element.

src/components/List/List.tsx
zamkovskaya: The function activateItem is not the only place where this.state.activeItem is modified. It is also modified, for example, in the [deactivate](https://github.com/gravity-ui/uikit/blob/d51039bb0fd4129ac511681e7fa23f051d1c3397/src/components/List/List.tsx#L384) function  and [onSortEnd](https://github.com/gravity-ui/uikit/blob/d51039bb0fd4129ac511681e7fa23f051d1c3397/src/components/List/List.tsx#L470). I don't want to duplicate the logic in each of these places, so I have moved it to componentDidUpdate.

src/components/Select/Select.tsx
zamkovskaya: useUniqId is a hook and cannot be called conditionally.

zamkovskaya:
In the List component, any type of selection (single or multiple) is represented with orange highlighting. In the Select component, single selection is also displayed as orange highlighting, while multiple selection is indicated by a checkbox icon. As a result, in the Select component, we disable the styles used for selection in the List and apply our custom styles. All the logic related to the selected option is also contained within the Select component. We pass the indexes of the selected options to the List to correctly set the aria-selected attribute. So, [this logic](https://github.com/gravity-ui/uikit/pull/960#:~:text=we%20actually%20need-,this%20logic,-if%20we%20use) is still required.

korvin89 (approved):

amje (approved):

Issue #861, opened 2023-08-04
Gravity UI landing – components documentation


imsitnikov:
# Our goals

We have to show documentation about our components with interactive examples of usage on the landing – https://gravity-ui.com/components/uikit

At the same time we want to maintain the ability of reading docs in the GitHub UI.

# The solution

We decided to store these docs into `uikit` repo for saving the one source of truth. And we've already prepared docs for two components:
1) `Button`
- https://github.com/gravity-ui/uikit/blob/main/src/components/Button/README.md
- https://gravity-ui.com/components/uikit/button
2) `Label`
- https://github.com/gravity-ui/uikit/blob/main/src/components/Label/README.md
- https://gravity-ui.com/components/uikit/label

Specific syntax in `README.md` files:
```
<!--LANDING_BLOCK

This string will be shown only on the landing

LANDING_BLOCK-->
```
```
<!--GITHUB_BLOCK-->

This string will be shown only on the GitHub UI

<!--/GITHUB_BLOCK-->
```

Interactive block for the landing:
```
<!--LANDING_BLOCK
<ExampleBlock
    code={`
<Label theme="normal" size="s">Normal</Label>
<Label theme="info" size="s">Info</Label>
<Label theme="warning" size="s">Warning</Label>
`}
>
    <UIKit.Label theme="normal" size="s">Normal</UIKit.Label>
    <UIKit.Label theme="info" size="s">Info</UIKit.Label>
    <UIKit.Label theme="warning" size="s">Warning</UIKit.Label>
</ExampleBlock>
LANDING_BLOCK-->
```

The same block with code example for the GitHub UI:
```
<!--GITHUB_BLOCK-->
`` `tsx
<Label theme="normal" size="s">Normal</Label>
<Label theme="info" size="s">Info</Label>
<Label theme="warning" size="s">Warning</Label>
`` `
<!--/GITHUB_BLOCK-->
```

# How can I see my changes on the landing before merge?

1. Prepare `README.md` in your branch, push it to the remote `uikit` repo.
2. Clone `landing` repo – https://github.com/gravity-ui/landing
3. `cd landing && npm ci && npm run start`
4. Find or add your component here – https://github.com/gravity-ui/landing/blob/main/src/content/components/uikit/index.ts
5. If component already exists, don't forget to remove the `isComingSoon` field.
6. Add this block to your component object:
```
content: {
[...]

PR #970, opened 2023-08-29
docs(Checkbox, Radio, RadioGroup, RadioButton): update Readme

zamkovskaya:
https://github.com/gravity-ui/uikit/issues/861

### DIFF
--- a/src/components/Switch/README.md
+++ b/src/components/Switch/README.md
@@ -79,94 +79,28 @@ You can set a label for a `Switch` component using the `content` property or pas
 <ExampleBlock
     code={`
 <div>
-  <Switch
-    content="content m"
-    size="m"
-  />
-  <span
-    style={{
-      margin: '16px'
-    }}
-   />
-  <Switch
-    content="content l"
-    size="l"
-  />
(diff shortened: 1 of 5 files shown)


src/components/Switch/README.md
```diff
@@ -83,7 +83,7 @@ LANDING_BLOCK-->
 | onChange       | Event handler for when the switch's value changes.              |                  `Function`                   |         |
 | onFocus        | Event handler for when the switch input element receives focus. |                  `Function`                   |         |
```
korvin89, line 84: The same issue about handlers

korvin89:
About the onChange and onUpdate handlers. I suggest pointing out that they are almost the same, called for the same user actions, but have differences in the signature of the arguments.

zamkovskaya:
> About the `onChange` and `onUpdate` handlers. I suggest pointing out that they are almost the same, called for the same user actions, but have differences in the signature of the arguments.

That's exactly what I specified in the last commit (#6ab43cf)

korvin89 (approved):

PR #973, opened 2023-08-30
feat: add CSS variables to control focus outline

Kyzyl-ool:
- Fixed focus styles for the components
- Introduced new css-variable `--g-color-line-focus`
- Introduced new css-variable `--g-focus-border-radius` to add border radius to backgroundless components

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.scss
+++ b/src/components/Breadcrumbs/Breadcrumbs.scss
@@ -22,7 +22,9 @@ $block: '.#{variables.$ns}breadcrumbs';
         &:focus {
             outline: 2px solid var(--g-color-line-focus);
             outline-offset: -4px; // due to padding
-            border-radius: calc(var(--g-focus-border-radius) + 4px); // +4px as outline-offset is -4px
+            border-radius: calc(
+                var(--g-focus-border-radius) + 4px
+            ); // +4px as outline-offset is -4px
         }
         &:focus:not(:focus-visible) {
             outline: 0;
--- a/src/components/Breadcrumbs/Breadcrumbs.scss
+++ b/src/components/Breadcrumbs/Breadcrumbs.scss
@@ -13,11 +13,13 @@ $block: '.#{variables.$ns}breadcrumbs';
 
     &__item {
         flex-shrink: 1;
-        margin: 4px 8px;
+        padding: 4px 8px;
         display: inline-block;
         overflow: hidden;
         white-space: nowrap;
         text-overflow: ellipsis;
+        outline-offset: -4px;
+        border-radius: calc(var(--g-focus-border-radius) + 4px); // +4px as outline-offset is -4px
 
         &_current {
             padding: 0 8px;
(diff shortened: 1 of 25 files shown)


src/components/Button/Button.scss
```diff
@@ -151,6 +151,10 @@ $iconWidth: 16px;
             --yc-button-background-color: var(--g-color-base-brand);
             --yc-button-background-color-hover: var(--g-color-base-brand-hover);
 
+            &:focus::before {
+                outline-offset: 1px;
```
amje, line 155: This setting should be global for all components

src/components/Breadcrumbs/Breadcrumbs.scss
```diff
@@ -13,7 +13,7 @@ $block: '.#{variables.$ns}breadcrumbs';
 
     &__item {
         flex-shrink: 1;
-        padding: 4px 8px;
+        margin: 4px 8px;
```
amje, line 16: This will probably break [AdaptiveTabs](https://github.com/gravity-ui/components/blob/main/src/components/AdaptiveTabs/AdaptiveTabs.tsx#L323)

src/components/Disclosure/Disclosure.scss
```diff
@@ -24,11 +24,14 @@ $gap: 8px;
         font-size: inherit;
         line-height: inherit;
         font-weight: inherit;
+        border-radius: 2px;
```
amje, line 27: This should not be a fixed value. And I doubt focus should have a radius at all by default. It component that determines the radius

amje (changes requested):

src/components/Button/Button.scss
Kyzyl-ool: Decided to make changes only for action button

src/components/Breadcrumbs/Breadcrumbs.scss
Kyzyl-ool: Returned padding, used outline-offset and adjusted border-radius

src/components/Disclosure/Disclosure.scss
Kyzyl-ool: Introduced --g-focus-border-radius

src/components/Button/Button.scss
```diff
@@ -150,6 +150,11 @@ $iconWidth: 16px;
         &_action {
             --yc-button-background-color: var(--g-color-base-brand);
             --yc-button-background-color-hover: var(--g-color-base-brand-hover);
+            --yc-button-outline-color: var(--g-color-base-brand);
+
+            &:focus::before {
+                outline-offset: 1px;
```
amje, line 156: Should be reset back to 0 when `selected` state is active

amje (changes requested):

src/components/Button/Button.scss
amje: Changed my mind. Let's add offset and color for selected state too

Kyzyl-ool: I think one shouldn't because the focus will "jump" when selected state will change

amje (approved):

PR #997, opened 2023-09-11
fix(Disclosure): exporting DisclosureSummary components

B0nBun:
In my case I only needed `DisclosureSummary` components, but maybe `DisclosureDetails` should be exported too?

### DIFF
--- a/src/components/Disclosure/index.ts
+++ b/src/components/Disclosure/index.ts
@@ -1,2 +1,3 @@
 export {Disclosure} from './Disclosure';
+export {DisclosureSummary, DefaultDisclosureSummary} from './DisclosureSummary/DisclosureSummary';
 export type {DisclosureSize, DisclosureArrowPosition, DisclosureProps} from './Disclosure';


Raubzeug:
Could you please provide a use case when `DefaultDisclosureSummary` is needed outside?
BTW, if you need `DisclosureSummary`, just use `Disclosure.DisclosureSummary`.

B0nBun:
@Raubzeug Yes, sorry for the hold up

In my case, I just needed to add some icons and styling in the summary, but `summary` prop of Disclosure only accepts strings, which is why I did a wrapper like this:

```js
<Disclosure>
  <DisclosureSummary>
    {({ onClick, ...summaryProps }) => (
      <div className={"summary-class"} onClick={onClick}>
        {renderSummary()}
        <DefaultDisclosureSummary onClick={noop} {...summaryProps} />
      </div>
    )}
  </DisclosureSummary>
  {renderContent()}
</Disclosure>
```

`DefaultDisclosureSummary` here is used just to recreate the functionality without maintaining the open/close state and events myself

PR #1006, opened 2023-09-13
feat: add hook useOnClickOutside

NikitaCG:


src/components/utils/useOnClickOutside/useOnClickOutside.tsx
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Hook for observing clicks outside a given target
+ *
+ * @param observedRef - purpose of observation
+ * @param enabled - enabled/disable flag
+ * @param handleCallback - callback when a click is triggered outside the observation target
+ *
+ * @return - nothing
+ */
+export const useOnClickOutside: UseOnClickOutsideType = ({
```
amje, line 21: Let's name the hook `useOutsideClick`

src/components/utils/useOnClickOutside/useOnClickOutside.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+
+export type UseOnClickOutsideType = <T extends HTMLElement>({
```
amje, line 3: We should export hook props type, not the hook function type

src/components/utils/useOnClickOutside/useOnClickOutside.tsx
```diff
@@ -0,0 +1,45 @@
...
+    observedRef,
+    enabled,
+}: {
+    observedRef: React.RefObject<T>;
+    enabled?: boolean;
```
amje, line 8: Seems this prop is needless and can be removed.

src/components/utils/useOnClickOutside/useOnClickOutside.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+
+export type UseOnClickOutsideType = <T extends HTMLElement>({
+    observedRef,
+    enabled,
+}: {
+    observedRef: React.RefObject<T>;
```
amje, line 7: Just `ref`

src/components/utils/useOnClickOutside/useOnClickOutside.tsx
```diff
@@ -0,0 +1,45 @@
...
+    enabled?: boolean;
+    handleCallback?: () => void;
```
amje, line 9: `handler` is more common name used in codebase

src/components/utils/useOnClickOutside/useOnClickOutside.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+                if (elem && !elem.contains(e.target as Node) && handleCallback) {
+                    handleCallback();
+                }
+            };
+
+            window.addEventListener('click', handler, {capture: true});
+
+            return () => {
+                window.removeEventListener('click', handler, {capture: true});
```
amje, line 39: There is no touch devices support. Let's add it

src/components/utils/useOnClickOutside/useOnClickOutside.tsx
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @return - nothing
+ */
+export const useOnClickOutside: UseOnClickOutsideType = ({
+    observedRef,
+    enabled = true,
+    handleCallback,
+}) => {
+    React.useEffect(() => {
+        if (enabled && observedRef) {
```
amje, line 27: ref is required in props, why check it?

amje (changes requested):

src/components/utils/useOnClickOutside/useOnClickOutside.tsx
NikitaCG: yes, but it can be null

src/components/utils/useOutsideClick/useOutsideClick.tsx
```diff
@@ -0,0 +1,40 @@
...
+            const callback = (e: MouseEvent | TouchEvent) => {
+                const elem = ref?.current;
+
+                if (elem && !elem.contains(e.target as Node) && handler) {
+                    handler();
+                }
+            };
+
+            window.addEventListener('click', callback, {capture: true});
+            window.addEventListener('touchstart', callback, {capture: true});
```
amje, line 30: Shouldn't it be `touchend` event?

amje (changes requested):

src/components/utils/useOnClickOutside/useOnClickOutside.tsx
NikitaCG: ok, I'm agree

src/components/utils/useOutsideClick/useOutsideClick.tsx
NikitaCG: It depends on how we decide. I assumed that we need a callback to be triggered as soon as the user touches another place. If you set `touchend`, it will work only when the user releases his finger, it may take some time before that

amje: But we use click event not mousedown (click happens after mouseup), there should be some consistency

amje: I did some research, most popular UI libraries either use click/touchend event or let specify event. For example, [MUI](https://github.com/mui/material-ui/blob/master/packages/mui-base/src/ClickAwayListener/ClickAwayListener.tsx#L237), [ChakraUI](https://github.com/chakra-ui/chakra-ui/blob/main/packages/hooks/use-outside-click/src/index.ts#L66)

src/components/utils/useOutsideClick/index.ts
```diff
@@ -0,0 +1,2 @@
+export {useOutsideClick} from './useOutsideClick';
+export type {UseOutsideClickProps} from './useOutsideClick';
```
amje, line 2: Let's also reexport this file in main index

amje (changes requested):

src/components/utils/useOutsideClick/index.ts
NikitaCG: done

amje (changes requested):

amje (approved):

PR #1027, opened 2023-09-22
fix(DocsDecorator): add new css prefix support

chelentos:


src/demo/DocsDecorator/DocsDecorator.scss
```diff
@@ -93,4 +44,55 @@ $root: '.docs-decorator';
     .docs-example {
         margin: $offset-m 0;
     }
+
+    :not([class*='yc-']):not([class*='g-']) > {
```
chelentos, line 48: Added to prevent styles from affecting elements without classes inside components

korvin89 (approved):

PR #1032, opened 2023-10-02
feat(withTableSorting): add control the use of data sorting

zzzorgo:
The method `isControlledState()` in this HOC is for determining whether the `sortState` is provided externally or not. It shouldn't turn off the whole sorting as it does in this case making the whole HOC useless.

### DIFF
--- a/src/components/Table/hoc/withTableSorting/withTableSorting.tsx
+++ b/src/components/Table/hoc/withTableSorting/withTableSorting.tsx
@@ -28,6 +28,7 @@ export interface WithTableSortingProps {
     defaultSortState?: SortState;
     sortState?: SortState;
     onSortStateChange?: (sortState: SortState) => void;
+    disableDataSorting?: boolean;
 }
 
 interface WithTableSortingState {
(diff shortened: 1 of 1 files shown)


src/components/Table/hoc/withTableSorting/withTableSorting.tsx
```diff
@@ -76,7 +76,7 @@ export function withTableSorting<I extends TableDataItem, E extends {} = {}>(
             const {data, columns} = this.props;
             const sortState = this.getSortState();
 
-            if (this.isControlledState() || sortState.length === 0) {
+            if (sortState.length === 0) {
```
ValeraS, line 79: It's breaking change. I think it's better add prop that controls a use of inner/external sorting,  and `onSortStateChange` should be always called, not only in controlled state.

```ts
const {useExternalSortings = this.isControlledState()} = this.props
if (useExternalSorting || sortState.length === 0) {
```

ValeraS (changes requested):

zzzorgo:
@ValeraS Pushed the change in accord with your comment

src/components/Table/hoc/withTableSorting/withTableSorting.tsx
```diff
@@ -28,6 +28,7 @@ export interface WithTableSortingProps {
     defaultSortState?: SortState;
     sortState?: SortState;
     onSortStateChange?: (sortState: SortState) => void;
+    useExternalSorting?: boolean;
```
ValeraS, line 31: Let's rename to
```suggestion
    disableDataSorting?: boolean;
```

ValeraS (changes requested):

ValeraS (approved):

PR #600, opened 2023-04-02
chore(Link)!: "href" property is required

LakeVostok:
UXRFC-1

PR #1045, opened 2023-10-10
fix: improve a11y on some components

KirillDyachkovskiy:
UXRFC-202


src/components/Link/Link.tsx
```diff
@@ -75,23 +75,24 @@ export const Link = React.forwardRef<HTMLElement, LinkProps>(function Link(
...
+            </a>
         );
     } else {
         return (
             <span
                 {...(extraProps as React.HTMLAttributes<HTMLSpanElement>)}
                 {...commonProps}
                 ref={ref as React.Ref<HTMLSpanElement>}
-                // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+                role="link"
```
ogonkov, line 95: I'm not sure that it is true

src/components/Table/hoc/withTableSettings/TableColumnSetup/TableColumnSetup.tsx
```diff
@@ -201,6 +201,7 @@ export const TableColumnSetup = (props: TableColumnSetupProps) => {
 
     return (
         <div className={b(null, className)}>
+            {/* It is used to handle clicks on the switcher control */}
```
ogonkov, line 204: that should not work this way

we should pass handler via some context i guess to easily pass it to rendered component

src/components/UserAvatar/UserAvatar.scss
```diff
@@ -14,6 +14,16 @@ $block: '.#{variables.$ns}user-avatar';
     background-position: center;
     background-size: cover;
 
+    outline: none;
+
+    &:focus {
```
ogonkov, line 19: we probably have SCSS mixin for it

src/components/UserAvatar/UserAvatar.tsx
```diff
@@ -34,9 +35,18 @@ export const UserAvatar = React.forwardRef<HTMLDivElement, UserAvatarProps>(
...
         return (
-            // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
-            <div className={b({size}, className)} title={title} onClick={onClick} ref={ref}>
+            <div
+                role="button"
+                tabIndex={0}
+                onKeyDown={onKeyDown}
+                className={b({size}, className)}
+                title={title}
+                onClick={onClick}
```
ogonkov, line 47: onClick supposed to be removed from this component

src/components/Link/Link.tsx
amje: argeed

amje: And I think we should remove this behaviour in the next major (#600)

src/components/Label/Label.tsx
```diff
@@ -154,10 +171,12 @@ export const Label = React.forwardRef<HTMLDivElement, LabelProps>(function Label
         }
 
         return (
-            // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
             <div
                 ref={ref}
-                onClick={hasOnClick ? onClick : undefined}
+                role={hasOnClick ? 'role' : undefined}
```
amje, line 176: role="button"?

src/components/Label/Label.scss
```diff
@@ -14,6 +14,20 @@ $transitionTimingFunction: ease-in-out;
         border: var(--border-size) solid #{$borderColor};
     }
 
+    // focus on interactive label (excluding focus on addon)
+    &:not(#{$disabled})#{$block}_is-interactive:focus:not(
```
amje, line 18: What about not setting is-interactive flag if disabled is true?

src/components/Label/Label.scss
```diff
@@ -14,6 +14,20 @@ $transitionTimingFunction: ease-in-out;
...
+            :has(#{$block}__addon_interactive:focus)
```
amje, line 19: If Label is clickable and has close button this wouldn't allow to focus the Label itself

src/components/Label/Label.scss
```diff
@@ -14,6 +14,20 @@ $transitionTimingFunction: ease-in-out;
...
+        ) {
+        box-shadow: 0 0 0 2px var(--g-color-line-focus);
```
amje, line 21: Let's use outline here

src/components/Table/hoc/withTableSettings/TableColumnSetup/TableColumnSetup.tsx
```diff
@@ -201,9 +210,10 @@ export const TableColumnSetup = (props: TableColumnSetupProps) => {
 
     return (
         <div className={b(null, className)}>
+            {/* FIXME remove switcher prop and this wrapper */}
             {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
             <div className={b('control')} ref={refControl} onClick={handleControlClick}>
-                {switcher || (
+                {renderSwitcher?.({onClick: handleControlClick}) || switcher || (
```
amje, line 216: We should create onKeyDown via `useActionHandlers` and pass it along with the onClick

amje (changes requested):

LakeVostok (approved):

amje (approved):

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -135,26 +147,34 @@ const DropdownMenu = <T,>({
         togglePopup();
     };
 
+    const {onKeyDown: handleSwitcherKeyDown} = useActionHandlers(handleSwitcherClick);
+
     return (
         <DropdownMenuContext.Provider value={contextValue}>
-            {/*as this div has Button component as child, clicking on it one fires onClick of this div on bubbling*/}
+            {/* FIXME remove switcher prop and this wrapper */}
```
kseniya57, line 154: When is this planned to be done?

kseniya57 (approved):

src/components/DropdownMenu/DropdownMenu.tsx
KirillDyachkovskiy: After a major library update

korvin89 (approved):

NikitaCG (approved):

DaffPunks (approved):

goshander (approved):

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -56,8 +62,13 @@ export type DropdownMenuProps<T> = {
     disabled?: boolean;
     /**
      * Menu toggle control.
+     * @deprecated Use renderSwitcher instead
```
axtk, line 65: With this change, the component's README also requires an update

src/components/DropdownMenu/README.md
```diff
@@ -395,24 +395,31 @@ LANDING_BLOCK-->
 
 <!--/GITHUB_BLOCK-->
 
+## Switcher properties
```
axtk, line 398: This section breaks in the current section, it breaks the following paragraph away from it

Let's move this section to the very bottom of the file, under a third-level `SwitcherProps` heading after the `DropdownMenuItem` section. Here, we could add a reference to the `SwitcherProps` section:

```md
It can be any function that returns a React component (or any `(props: SwitcherProps) => React.ReactNode` in the TypeScript terms, see [`SwitcherProps`](#switcherprops) below).
```

axtk (approved):

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -56,8 +62,13 @@ export type DropdownMenuProps<T> = {
...
      */
     switcher?: React.ReactNode;
+    /**
+     * Menu toggle control.
+     */
+    renderSwitcher?: (props: SwitcherProps) => React.ReactNode;
```
ogonkov, line 71: Why we need this prop? We can keep current api, but use cloneElement to add handlers to `switcher`

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -135,26 +147,34 @@ const DropdownMenu = <T,>({
...
 
+    const {onKeyDown: handleSwitcherKeyDown} = useActionHandlers(handleSwitcherClick);
+
     return (
         <DropdownMenuContext.Provider value={contextValue}>
-            {/*as this div has Button component as child, clicking on it one fires onClick of this div on bubbling*/}
+            {/* FIXME remove switcher prop and this wrapper */}
             {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
             <div
                 ref={anchorRef}
```
KirillDyachkovskiy, line 157: The ref is passed twice, yes it will return the second element, but is it better to add a check like this `Boolean(render Switcher) ? undefined : controlRef`

PR #1049, opened 2023-10-11
chore: refactor hooks : moved few hooks in right folder

NikitaCG:
- useSelect
 - useTimeout
 - useViewportSize
 - useListNavigation
 - useIntersection


src/hooks/useViewportSize/useViewportSize.tsx
```diff
@@ -1,11 +1,11 @@
 import React from 'react';
 
-export interface ViewportSize {
+export interface ViewportSizeResult {
```
amje, line 3: We should keep this type as is, and create `UseViewportSizeResult` type as an alias to this type

amje (changes requested):

src/hooks/useViewportSize/useViewportSize.tsx
NikitaCG: Up to this point, we haven't exported this hook yet, so no one can use it. That's why I renamed without an alias. Are you thinking of making an alias after all?

amje: Then `Use` should be at the beginning

NikitaCG: done

src/hooks/useViewportSize/README.md
```diff
@@ -0,0 +1,18 @@
...
+```tsx
+import {useViewportSize} from '@gravity-ui/uikit';
+```
+
+The `useViewportSize` hook calculates the size of the viewport at the resize event
+
+## Result
+
+|  Name  |   Description   |   Type   |
+| :----: | :-------------: | :------: |
```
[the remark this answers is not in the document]
NikitaCG: fixed

amje (approved):

PR #1060, opened 2023-10-17
refactor: refactor hooks useEventBroker and useLayer

NikitaCG:
Hook `useEventBroker` and `useLayer` very connected with their classes, therefore, there was no point in taking them out separately. Instead, I refactored folder for these utilities


korvin89 (commented):
About naming: we are trying to use kebab-case for folders and files except react component  files naming that uses camel case. Could u rename `utils/eventBroker` -> `utils/event-broker`?

NikitaCG:
> About naming: we are trying to use kebab-case for folders and files except react component files naming that uses camel case. Could u rename `utils/eventBroker` -> `utils/event-broker`?

done

korvin89 (approved):

src/components/utils/layer-manager/index.ts
```diff
@@ -0,0 +1,9 @@
+export {layerManager, getLayersCount} from './LayerManager';
+export type {
+    LayerConfig,
+    LayerExtendableProps,
+    ContentElement,
```
amje, line 5: Not sure if we want LayerExtendableProps and ContentElement to be exposed

amje (changes requested):

src/components/utils/layer-manager/index.ts
NikitaCG: done

amje (approved):

PR #1069, opened 2023-10-24
fix(Button): added outline-offset for outlined buttons family 

### LINKED ISSUE #1044: fix: need to avoid double outline on buttons
Button with `view=outlined` has double outline on focus. Need to adjust outline-offset for it

arudenkoofficial:
Closes #1044

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -389,6 +389,18 @@ $iconWidth: 16px;
                 }
             }
         }
+
+        &_outlined,
+        &_outlined-info,
+        &_outlined-success,
+        &_outlined-warning,
+        &_outlined-danger,
+        &_outlined-action,
+        &_outlined-contrast {
+            &:focus::before {
+                outline-offset: -2px;
+            }
+        }
     }
 
     @include mixins.pin($block, ('::before', '::after'), var(--yc-button-border-radius));


amje:
@arudenkoofficial Could you please give an example of double outline mentioned in the issue? I have only one

arudenkoofficial:
@amje Sure!

`view=outlined` or `view=outlined-*` buttons are named for the border css property set. So each of these buttons has a contrasting brightly colored border. 

When the button gets focus from pressing Tab for example, the button element gets outline second layer above its border (see some screenshots before fixes from storybook)
Screenshot 2023-10-25 at 21 47 38
Screenshot 2023-10-25 at 21 47 06
Screenshot 2023-10-25 at 21 46 39

My fixes move outline 2 pix inside the element. In this case, the outline will not affect the visual size of the element. However, as I was writing this reply I've realised that the elements look a bit dirty - the semi-transparent outline is overlapping the contrasting border.

What do you think?
Screenshot 2023-10-25 at 21 57 32
Screenshot 2023-10-25 at 21 56 46
Screenshot 2023-10-25 at 21 56 19

amje:
@arudenkoofficial I think the current behaviour is OK. Focus line enlarging outside is better then enlarging inside, because it's more noticeable for the user. Besides, your fixes introduce visual glitches of 2 borders stacking.

arudenkoofficial:
@amje Lets close PR and isseu in that case?

arudenkoofficial:
@amje on the other hand, I can avoid visual glitches by manipulating the border property when focusing on the button. I mean – making border transparent on focus, but keep outline. To be honest I'm not sure if this is the best solution

arudenkoofficial:
@amje Gentle ping 🙂

amje:
@arudenkoofficial I asked designers opinion on that. Waiting for the response

amje:
@arudenkoofficial Designers are also not willing to apply this change. Let's keep the current styles. PR can be closed.

PR #1077, opened 2023-10-28
fix(Popup): position render

Antamansid:
When rendering popup to popup, the internal popup loses its position. 
The position of the browser autocomplete is also lost when placing forms in the popup.
Trying to solve this problem using the `onTransitionEnd` event.

### DIFF
--- a/src/components/Popup/Popup.tsx
+++ b/src/components/Popup/Popup.tsx
@@ -82,6 +82,7 @@ export function Popup({
     autoFocus = false,
 }: PopupProps) {
     const containerRef = React.useRef<HTMLDivElement>(null);
+    const [inTransition, setInTransition] = React.useState(true);
 
     useLayer({
         open,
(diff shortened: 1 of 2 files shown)


korvin89:
@Antamansid Hey! We really sorry for long response. You can use state for ref handling. It may look like:
```tsx

export const PopupInPopup: StoryFn<PopupProps> = (props: PopupProps) => {
    const anchorRef = React.useRef<HTMLButtonElement>(null);
    const [anotherAnchor, setAnotherAnchor]= React.useState<HTMLButtonElement>(null);
    const [open, setOpen] = React.useState(false);

    return (
        <React.Fragment>
            <Popup {...props} open={open} anchorRef={anchorRef} onClose={() => setOpen(false)}>
                <div style={{padding: 10}} ref={setAnotherAnchor}>
                    Popup content
                </div>
                <Popup {...props} open={open} anchorRef={{current: anotherAnchor}}>
                    <div style={{padding: 10}}>Popup content</div>
                </Popup>
            </Popup>
            <div
                style={{
                    width: '100%',
                    height: 200,
                    display: 'flex',
                    alignItems: 'center',
                    justifyContent: 'center',
                }}
            >
                <Button ref={anchorRef} onClick={() => setOpen(!open)}>
                    {open ? 'Hide' : 'Show'}
                </Button>
            </div>
        </React.Fragment>
    );
};
```

PR #1039, opened 2023-10-04
feat: new list proposal

IsaevAlexandr:
Only to familiarize yourself with the approach. Work still in progress

PR #1090, opened 2023-11-03
feat(TreeSelect): added TreeSelect unstable component and new list hooks

IsaevAlexandr:
Completely rewritten to version with hooks.

Prev pr: https://github.com/gravity-ui/uikit/pull/1039


Mishnya:
I see you're using the `react-window`.

This library has a problem. In rtl, the list for an item when dragging sets the style `right: 0` instead `right: undefined`, which causes the item to move to the right until the end of the screen.

To fix this you can manually set `right: undefined` to the item. I didn't find any more problems with with this bug and fix but maybe we can find an alternative to this library (which has not been updated since December 2018).

src/components/useList/components/ListContainerView/ListContainerView.tsx
```diff
@@ -0,0 +1,43 @@
...
+import type {QAProps} from 'src/components/types';
+
+import {Flex} from '../../../layout';
+import {block} from '../../../utils/cn';
+
+import './ListContainerView.scss';
+
+const b = block('list-container-view');
+
+export interface ListContainerViewProps extends QAProps, React.HTMLAttributes<'div'> {
```
amje, line 12: Let's apply rest props via separate prop, `extraProps`. The component should alse allow to override `div` element

src/components/useList/components/ListContainerView/ListContainerView.scss
```diff
@@ -0,0 +1,17 @@
...
+
+$block: '.#{variables.$ns}list-container-view';
+
+#{$block} {
+    box-sizing: border-box;
+    width: 100%;
+    outline: none;
+
+    &_fixed-height {
+        height: var(--g-list-height, 300px);
```
amje, line 11: Let's not mix stable and unstable feautures together, `--g-list-height` probably will be a part of CSS for old List

amje (changes requested):

src/components/useList/hooks/useListFilter.ts
```diff
@@ -0,0 +1,97 @@
...
+ * @returns -
+ */
+export function useListFilter<T>({
+    items: externalItems,
+    initialFilterValue = '',
+    filterItem,
+    filterItems,
+    debounceTimeout = 300,
+}: UseListFilterProps<T>) {
+    const filterRef = React.useRef<HTMLInputElement>(null);
```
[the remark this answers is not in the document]
IsaevAlexandr: for two reasons:
- over time, logic may appear where access is needed access to DOM element;
- it is convenient to have a ready-made ref to use right in the hook

src/components/useList/components/ListItemView/ListItemView.tsx
```diff
@@ -0,0 +1,183 @@
...
+                                    size={16}
+                                    className={colorText({color: 'info'})}
+                                />
+                            ) : null}
+                        </Slot>
+                    )}
+
+                    {renderSafeIndentation(indentation)}
+
+                    {startSlot ??
```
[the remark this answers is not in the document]
IsaevAlexandr: to override this behaviour. If you need both icons pass component in slot like:
```tsx
startSlot={
   <>
      <Icon />
      <Icon data={Check} size={16} className={colorText({color: 'info'})} />
  </>
}
```

src/components/useList/components/ListItemView/ListItemView.scss
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    &_selected,
+    &_selected:hover#{$block}_activeOnHover {
+        background: var(--g-color-base-selection);
+    }
+
+    &__slot {
+        &_indent_1 {
+            width: 16px;
```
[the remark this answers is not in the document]
IsaevAlexandr: it's a component with fixed width corresponding to space if list has nested structure

src/components/useList/components/ListItemView/ListItemView.scss
```diff
@@ -20,6 +20,19 @@ $block: '.#{variables.$ns}list-item-view';
...
+        border-radius: var(--g-list-item-border-radius-s, 3px);
+    }
+    &_radius_m {
+        border-radius: var(--g-list-item-border-radius-m, 6px);
+    }
+    &_radius_l {
+        border-radius: var(--g-list-item-border-radius-l, 8px);
+    }
+    &_radius_xl {
+        border-radius: var(--g-list-item-border-radius-xl, 8px);
```
amje, line 33: public var should be without any size, just `--g-list-item-border-radius`

src/components/useList/types.ts
```diff
@@ -1,6 +1,6 @@
 export type ListItemId = string;
 
-export type ListSizeTypes = 's' | 'm' | 'l' | 'xl';
+export type ListItemSizeType = 's' | 'm' | 'l' | 'xl';
```
amje, line 3: `ListItemSize`, "Type" suffix is redundant

src/unstable.ts
```diff
@@ -1,2 +1,8 @@
-export {useList, useListFilter, useListKeydown} from './components/useList';
+/* eslint-disable camelcase */
+export {
+    useList as unstable_useList,
+    useListState as unstable_useListState,
+    useListFilter as unstable_useListFilter,
+    useListKeydown as unstable_useListKeydown,
+} from './components/useList';
 export * from './components/TreeSelect';
```
amje, line 8: We keep TreeSelect stable?

amje (changes requested):

src/components/useList/components/ListItemView/ListItemView.tsx
```diff
@@ -0,0 +1,184 @@
...
+            hasSelectionIcon = true,
+            indentation,
+            startSlot,
+            subtitle,
+            endSlot,
+            title,
+            height,
+            expanded,
+            style,
+            role = 'listitem',
```
amje, line 100: role `option` more suitable here, it supports `aria-selected` attribute, while `listitem` doesn't

amje (changes requested):

korvin89 (approved):

src/components/useList/hooks/useListFilter.ts
```diff
@@ -0,0 +1,98 @@
...
+
+            return () => items;
+        },
+        [filterItem, filterItems],
+    );
+
+    if (externalItems !== prevItems) {
+        setItems(filterItemsFn(filter, externalItems));
+        setPrevItems(externalItems);
+    }
```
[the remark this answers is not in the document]
IsaevAlexandr: it's new recommended way to not use `useEffect` things in code - https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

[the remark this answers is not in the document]
GermanVor: Interesting. Why they do not use `React.useRef` for prev state

Issue #881, opened 2023-08-09
Hooks docs and typings


amje:
Each exported hook must have a README file and follow this template:
```ts
// hooks/useMyHookName.ts

export interface MyHookNameProps {}

export interface MyHookNameResult {}

export function useMyHookName(props: MyHookNameProps): MyHookNameResult {}
```

```ts
// index.ts

// ...
export * from './hooks/useMyHookName';
// ...
```

NikitaCG:
Сall me when we start

NikitaCG:
I suggest attaching PR in the comments, since there will be more than one of them before the issue closes
[useOutsideClick](https://github.com/gravity-ui/uikit/pull/1035)

NikitaCG:
[PR](https://github.com/gravity-ui/uikit/pull/1038) with next hooks: 

- useBodyScrollLock
- useVirtualElementRef
- usePortalContainer

NikitaCG:
PRs:
 - https://github.com/gravity-ui/uikit/pull/1038
 - https://github.com/gravity-ui/uikit/pull/1042
 - https://github.com/gravity-ui/uikit/pull/1049
 - https://github.com/gravity-ui/uikit/pull/1060
 - https://github.com/gravity-ui/uikit/pull/1084

PR #1095, opened 2023-11-04
feat(useAsyncActionHandler): add useAsyncActionHandler hook

kseniya57:


src/hooks/useAsyncActionHandler/useAsyncActionHandler.ts
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+export function useAsyncActionHandler<Result>(
```
korvin89, line 3: 1. What is `Result`? There is no such type in the current namespace
2. It is not convenient to have generic with а required parameter

amje, line 3: Let's follow this convention https://github.com/gravity-ui/uikit/issues/881

src/hooks/useAsyncActionHandler/useAsyncActionHandler.ts
```diff
@@ -0,0 +1,30 @@
...
+            return action(...args).finally(() => {
+                if (mounted.current) {
+                    setLoading(false);
+                }
+            });
+        },
+        [action],
+    );
+
+    return [isLoading, handleAction];
```
amje, line 29: I would change to object here: `{isLoading, handler}`. Tuples are primary used for `[value, setValue]`

src/hooks/useAsyncActionHandler/useAsyncActionHandler.ts
```diff
@@ -0,0 +1,30 @@
...
+
+        return () => {
+            mounted.current = false;
+        };
+    }, []);
+
+    const handleAction: typeof action = React.useCallback(
+        (...args) => {
+            setLoading(true);
+            return action(...args).finally(() => {
```
amje, line 20: Let's add check if returned value is a promise before calling finally

amje (changes requested):

korvin89 (approved):

amje (approved):

src/hooks/useAsyncActionHandler/useAsyncActionHandler.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    const handleAction: typeof handler = React.useCallback(
+        (...args) => {
+            setLoading(true);
+
+            const handlerResult = handler(...args);
+
+            if (handlerResult instanceof Promise) {
+                return handlerResult.finally(() => {
+                    if (mounted.current) {
```
ValeraS, line 34: Why do we need this check?

ValeraS (changes requested):

src/hooks/useAsyncActionHandler/useAsyncActionHandler.ts
amje: Not calling setState on unmount component?

ValeraS: In React 18  the warning about using setState on an unmounted component has been removed as misleading. And this wrong solution is described here https://github.com/reactwg/react-18/discussions/82

amje: Didn't know that. Ok then let's remove this code

amje:
@ValeraS ping

src/hooks/useAsyncActionHandler/useAsyncActionHandler.ts
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+
+export interface UseAsyncActionHandlerProps<ActionResult = void> {
+    handler: (...args: any[]) => Promise<ActionResult>;
```
ValeraS, line 4:
```suggestion
    handler: (...args: any[]) => PromiseLike<ActionResult>;
```

src/hooks/useAsyncActionHandler/useAsyncActionHandler.ts
```diff
@@ -0,0 +1,48 @@
...
+                // eslint-disable-next-line no-console
+                console.error(
+                    "@gravity-ui/uikit: (useAsyncActionHandler) the provided handler doesn't return a promise",
+                );
+            }
+
+            setLoading(false);
+
+            return handlerResult;
+        },
```
ValeraS, line 40:
```suggestion
        async (...args) => {
            setLoading(true);

            try {
                return await handler(...args);
            } finally {
                setLoading(false);
            }
        },
```

ValeraS (changes requested):

src/hooks/useAsyncActionHandler/useAsyncActionHandler.ts
kseniya57: Why? `Promise` can have only `catch` (https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_es5_d_.promise.html), but `PromiseLike` can't (https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_es5_d_.promiselike.html)

src/hooks/useAsyncActionHandler/useAsyncActionHandler.ts
kseniya57: 1) We should notify the user that the handler is not a promise, otherwise the use of the hook will not be justified 
2) The error should be caught by the user, not by us

src/hooks/useAsyncActionHandler/useAsyncActionHandler.ts
ValeraS: Why do you need `catch`?  The code does not call this method.

src/hooks/useAsyncActionHandler/useAsyncActionHandler.ts
ValeraS: 1. Why? The types say we are waiting for a PromiseLike and if the user passes anything else they won't see the loader.
2. We do not catch errors here, they will be propagated to the user.

src/hooks/useAsyncActionHandler/useAsyncActionHandler.ts
kseniya57: The user of the library

src/hooks/useAsyncActionHandler/useAsyncActionHandler.ts
kseniya57: The user of the library can use `as`

src/hooks/useAsyncActionHandler/useAsyncActionHandler.ts
kseniya57: See, we return typeof the provided handler and it's important, we don't want to change the type of it. If we do like you want we receive the following:


Снимок экрана 2023-11-14 в 08 13 02

ValeraS: Current types are not good enough. Now `useAsyncActionHandler` is returning type of `handler` as `(...args: any[]) => Promise<ReturnType<RealHandler>>`  and loosing type of arguments. I suggest next:

```ts
export interface UseAsyncActionHandlerProps<Action> {
    handler: Action;
}

type AnyAsyncAction = (...args: any[]) => PromiseLike<any>;

export interface UseAsyncActionHandlerResult<Action extends AnyAsyncAction> {
    isLoading: boolean;
    // Here we return a handler with the same parameters and the same result as the user handler, but the result is always wrapped in a Promise.
    handler: (...args: Parameters<Action>) => Promise<Awaited<ReturnType<Action>>>;
}

export function useAsyncActionHandler<Action extends AnyAsyncAction>({
    handler,
}: UseAsyncActionHandlerProps<Action>): UseAsyncActionHandlerResult<Action> {
    const [isLoading, setLoading] = React.useState(false);

    const handleAction = React.useCallback<<UseAsyncActionHandlerResult<Action>['handler']>>(
        async (...args) => {
            setLoading(true);

            try {
                return await handler(...args);
            } finally {
                setLoading(false);
            }
        },
        [handler],
    );

    return {
        isLoading,
        handler: handleAction,
    };
}
```

ValeraS (changes requested):

ValeraS (approved):

PR #1096, opened 2023-11-07
fix(Icon): correct prepareStringData

DarkGenius:


src/components/Icon/__tests__/utils.test.ts
```diff
@@ -0,0 +1,11 @@
+import {prepareStringData} from '../utils';
+
+describe('Icon utils', () => {
+    it('should remove width and height only inside svg tag', () => {
+        const data =
+            '<svg width="100" viewBox="0 0 200 200" height="200">\n<rect x="50" y="50" width="100" height="100" fill="red" />\n</svg>';
+        expect(prepareStringData(data)).toEqual(
```
ogonkov, line 7: `.toBe` is more straightforward

DarkGenius: `.toBe` is specifically useful for reference equality, for string comparison there is no difference here

korvin89 (approved):

src/components/Icon/utils.ts
```diff
@@ -22,7 +22,12 @@ export function isStringSvgData(data: SVGIconData): data is SVGIconStringData {
     return typeof data === 'string';
 }
 export function prepareStringData(data: SVGIconStringData) {
-    return data.replace(/(width|height)=(["']?)\d+\2/g, '');
```
amje, line 25: We should keep the same logic for quotes

amje (changes requested):

amje (approved):

PR #1107, opened 2023-11-08
fix: pass ref to renderSwitcher props

KirillDyachkovskiy:


src/components/Table/hoc/withTableSettings/TableColumnSetup/TableColumnSetup.tsx
```diff
@@ -216,8 +201,16 @@ export const TableColumnSetup = (props: TableColumnSetupProps) => {
         <div className={b(null, className)}>
             {/* FIXME remove switcher prop and this wrapper */}
             {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions */}
-            <div className={b('control')} ref={refControl} onClick={handleControlClick}>
-                {renderSwitcher?.({onClick: handleControlClick, onKeyDown: handleControlKeyDown}) ||
+            <div
+                //@ts-ignore
+                ref={controlRef}
```
KirillDyachkovskiy, line 206: The ref is passed twice, yes it will return the second element, but is it better to add a check like this `Boolean(render Switcher) ? undefined : controlRef`

PR #1111, opened 2023-11-08
feat!: support RTL direction

amje:


NikitaCG (approved):

zamkovskaya (approved):

korvin89 (approved):

.stylelintrc
```diff
@@ -1,6 +1,8 @@
 {
     "extends": ["@gravity-ui/stylelint-config", "@gravity-ui/stylelint-config/prettier"],
+    "plugins": ["stylelint-use-logical"],
     "rules": {
+        "csstools/use-logical": ["always", {"except": ["float", "width", "min-width", "max-width", "height", "min-height", "max-height"]}],
```
ogonkov, line 5: why this props are exception?

src/components/theme/updateBodyElement.ts
ogonkov, line 1: that would be better to add new function for `dir` toggle instead of putting new logic here

.stylelintrc
amje: float has weak browsers support, widht and height needed for different `writing-mode` which we are not planning to support now

src/components/theme/updateBodyElement.ts
amje: done

mournfulCoroner (approved):

PR #1122, opened 2023-11-14
feat(Text): support multiline and fixed width ellipsis

IsaevAlexandr:


src/components/Text/text/text.ts
```diff
@@ -71,6 +71,13 @@ export interface TextBaseProps {
      * - text-overflow: ellipsis;
      */
     ellipsis?: boolean;
+    /**
+     * The number of whole lines of text after which the content will be cut off.
+     *
+     * !Note: supports only modern browsers
+     * https://caniuse.com/?search=display%3A%20-webkit-box%3B
+     */
+    ellipsisLines?: 2 | 3 | 4 | 5;
```
amje, line 80: What if I want more than 5? Maybe let this to be set via JS, in style attribute?

amje (changes requested):

src/components/Text/Text.tsx
```diff
@@ -57,19 +61,33 @@ export const Text = React.forwardRef(
...
+
         return (
             <Tag
                 ref={ref}
                 className={text(
-                    {variant, ellipsis, whiteSpace, wordBreak},
+                    {
+                        variant,
+                        ellipsis: Boolean(ellipsis),
+                        ellipsisLines,
```
amje, line 84: Maybe `ellipsisLines`  is better to be the `ellipsis` prop? Boolean is for normal ellipsis, Number is for line-clamp. Why we need this shortcut: `ellipsis: 200` === `style: {maxWidth: 200}`? It seems weird for me

IsaevAlexandr: Very often it is necessary to limit the length of the text to no more than a fixed value. Every time add a container for this - redundant

src/components/Text/text/text.ts
IsaevAlexandr: added limit by 10 in text utility and unlimited lines count in Text compoennt

src/components/Text/Text.tsx
amje: We can use `style: {maxWidth: <value>}`. Why put it into prop `ellipsis`, there is no relation to it

src/components/Text/text/text.scss
```diff
@@ -68,8 +70,28 @@ $block: '.#{variables.$ns}text';
         @include mixins.text-subheader-3();
     }
     &_ellipsis {
+        display: inline-block;
         @include mixins.overflow-ellipsis();
     }
+
+    // ellipsis lines
+    &_el {
```
amje, line 78: Let's name it `_ellipsis-lines`

src/components/Text/text/text.scss
```diff
@@ -68,8 +70,28 @@ $block: '.#{variables.$ns}text';
...
+        display: -webkit-box; // stylelint-disable-line value-no-vendor-prefix
+        -webkit-box-orient: vertical;
+        overflow: hidden;
+        align-self: center;
+        white-space: normal;
+        // by default clamp 2 lines
+        -webkit-line-clamp: 2;
+    }
+
+    @each $size in $supportedLineCountSizes {
```
amje, line 88: We generate ton of CSS for no reason. Why don't we just let `ellipsisLines` to be a `number` type instead? There is no need in second prop `ellipsisLinesCount`.

amje (changes requested):

src/components/Text/text/text.scss
IsaevAlexandr: it's only for inline text util usage, a remove this

src/components/Text/Text.tsx
IsaevAlexandr: it's not obvious to use max width with this component to work. So now it's build in

amje (changes requested):

amje (approved):

src/components/Text/Text.tsx
```diff
@@ -48,39 +54,54 @@ type TextPropsWithoutRef<C extends React.ElementType> = {as?: C} & Omit<TextProp
...
-            </Tag>
-        );
-    },
-) as (<C extends React.ElementType = 'span'>({
+    const style: React.CSSProperties = {
+        ...outerStyle,
+    };
+
+    if (typeof ellipsisLines === 'number') {
+        style.WebkitLineClamp = ellipsisLines;
```
ogonkov, line 81: What the point to use inline style, and not modifier in stylesheet?

PR #1125, opened 2023-11-14
feat: support RTL in components

amje:


src/components/Button/ButtonIcon.tsx
```diff
@@ -6,10 +6,17 @@ const b = block('button');
 
 type Props = React.PropsWithChildren<{
     className?: string;
-    side?: 'start' | 'end';
+    side?: 'left' | 'right' | 'start' | 'end';
 }>;
 
 export const ButtonIcon = ({side, className, children}: Props) => {
+    if (side === 'left') {
```
korvin89, line 13: Lets add a new var here?

src/components/Popup/Popup.scss
```diff
@@ -75,15 +75,17 @@ $transition-distance: 10px;
     }
 
     &[data-popper-placement*='left'] #{$block}__arrow {
-        inset-inline-end: -$arrow-offset;
+        // stylelint-disable-next-line csstools/use-logical
```
korvin89, line 78: Should we add todo here?

amje: No, popper.js uses physical properties, it doesn't support RTL, we should migrate to Floating UI, which does

src/components/Button/ButtonIcon.tsx
amje: done

korvin89 (approved):

PR #1130, opened 2023-11-16
Lack of control over clear icon in Select

richvoronov:
I would like to have a click event on the clear state icon for the Select component

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -74,6 +74,7 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function 
         disablePortal,
         hasClear = false,
         onClose,
+        onClearClick,
         id,
     } = props;
     const [mobile] = useMobile();
@@ -102,6 +103,7 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function 
         open: propsOpen,
         onClose,
         onOpenChange,
+        onClearClick,
     });
     const uniqId = useUniqId();
     const selectId = id ?? uniqId;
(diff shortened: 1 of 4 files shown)


richvoronov:
hi

1. Select component does not have an input and text selections
2. but there is a way to clear state with _hasClear_ prop. It's useful for Select with _multiple_ prop
3. and when you click on the clear button you don't have a callback, but for my UI case it's needed (I hide my select after clear click)

ValeraS:
> 3. and when you click on the clear button you don't have a callback, but for my UI case it's needed (I hide my select after clear click)

What happens when the user deselects all options in the popup?

richvoronov:
> > 3. and when you click on the clear button you don't have a callback, but for my UI case it's needed (I hide my select after clear click)
> 
> What happens when the user deselects all options in the popup?

onUpdate fires

ValeraS:
> > > 3. and when you click on the clear button you don't have a callback, but for my UI case it's needed (I hide my select after clear click)
> > 
> > 
> > What happens when the user deselects all options in the popup?
> 
> onUpdate fires

Yes, and what happens next? Why if user clears select by the clear button then select disappears but if user deselect all options select stays? How in this scenario can the user hide the select after deselecting all options?

richvoronov:
I did not understand the question. Nothing changed in the UI. PR about adding additional control over the cleaning function

ValeraS:
> I did not understand the question. Nothing changed in the UI. PR about adding additional control over the cleaning function

I'm trying to understand your scenario and don't get it. User can hide select only by clear button, right? And he can do it only after selecting at least one option?

richvoronov:
user can hide select with the clear button and with other cases. Check the video here https://github.com/gravity-ui/uikit/issues/1134 (20 sec) to see what I mean.

Anyway, I would not like to discuss how this is in my design, in my opinion it is not bad if there is more control over clicks in the component

PR #1140, opened 2023-11-17
feat(Disclosure): use logical properties

Raubzeug:


src/components/Disclosure/Disclosure.tsx
```diff
@@ -11,7 +11,7 @@ import {b} from './cn';
 import './Disclosure.scss';
 
 export type DisclosureSize = 'm' | 'l' | 'xl';
-export type DisclosureArrowPosition = 'left' | 'right';
+export type DisclosureArrowPosition = 'start' | 'end';
```
amje, line 14: Let's keep old values, see [example](https://github.com/gravity-ui/uikit/blob/next/src/components/Button/ButtonIcon.tsx#L15)

amje (changes requested):

src/components/Disclosure/README.md
```diff
@@ -11,7 +11,7 @@ Disclosure component that shows and hides enclosed content.
 | disabled        | `Boolean`                     |          | `false` | Disabled state                                           |
 | defaultExpanded | `Boolean`                     |          | `false` | Default opening state                                    |
 | expanded        | `Boolean`                     |          |         | Controlled opening state                                 |
-| arrowPosition   | `left` `right`                |          | `left`  | Control position                                         |
+| arrowPosition   | `left` `right` `start` `end`  |          | `start` | Control position                                         |
```
amje, line 14: We suppose left and right are being deprecated, so they don’t need to be documented

src/components/Disclosure/DisclosureSummary/DisclosureSummary.tsx
```diff
@@ -38,11 +38,19 @@ export function DefaultDisclosureSummary({
     disabled,
 }: DisclosureSummaryRenderFunctionProps) {
     const {size, summary, arrowPosition} = useDisclosureAttributes();
+    let sideMod = arrowPosition;
```
amje, line 41: Isn’t it `arrowMod`?

amje (changes requested):

src/components/Disclosure/Disclosure.tsx
```diff
@@ -11,7 +11,7 @@ import {b} from './cn';
 import './Disclosure.scss';
 
 export type DisclosureSize = 'm' | 'l' | 'xl';
-export type DisclosureArrowPosition = 'left' | 'right' | 'start' | 'end';
```
amje: It would break typings for those who passes this type to the props. Let's keep them here. I will add warning later, for all such types

amje (changes requested):

amje (approved):

PR #1141, opened 2023-11-19
feat(Progress): support rtl

eyoue:


src/components/Progress/Progress.tsx
```diff
@@ -144,7 +144,7 @@ export class Progress extends React.Component<ProgressProps> {
         const className = b('item', {theme: this.getTheme(), loading: this.props.loading});
 
         const offset = Progress.getOffset(value);
-        const style = {transform: `translateX(${offset}%)`};
+        const style = {insetInlineStart: `${offset}%`};
```
amje, line 147: Animating margin instead of transform will cause a reflow on update. Let's keep transform

amje (changes requested):

src/components/Progress/Progress.tsx
eyoue: Yes, tnx. I changed to transform using css variables

src/components/Progress/Progress.scss
```diff
@@ -7,6 +7,14 @@ $progress-m-height: 20px;
 $progress-s-height: 10px;
 $progress-xs-height: 4px;
 
+:root {
+    --flow-direction: 1;
+}
+
+[dir='rtl'] {
+    --flow-direction: -1;
+}
```
amje, line 16: Please you local variable here, with "private" naming:
```scss
.progress {
  --_--flow-direction: 1;
  
  [dir='rtl'] & {
    --_--flow-direction: -1;
  }
}
```

amje (changes requested):

amje:
Also Progress with stack still ignores rtl

src/components/Progress/Progress.scss
eyoue: Tnx, i changed it.

eyoue:
> Also Progress with stack still ignores rtl

Ops :C
Yes, really.. Comonent use float.
Ok, i changed it for rtl. Tnx.

amje (approved):

amje:
@Lunory Numbers have own semantics here, 60/100 is the correct one for both LTR and RTL

Lunory (approved):

PR #1145, opened 2023-11-20
feat(List): support RTL

Mishnya:


korvin89 (approved):

src/components/Select/__stories__/SelectShowcase.tsx
```diff
@@ -95,7 +95,7 @@ const ExampleItem = (props: {
                 code.map((codeItem, i) => {
                     return (
                         <React.Fragment key={`${title}-${i}`}>
-                            <pre>
+                            <pre dir="ltr">
```
amje, line 98: Why this needed?

Mishnya: Without this code for Select is wrong
Снимок экрана 2023-11-20 в 17 12 07

amje (approved):

PR #1149, opened 2023-11-21
fix: align checkbox when there is more than 1 line of text

Bluepuper:


EgorKluch:
Please, add to storybook example of rows with multiline text.

Bluepuper:
> Please, add to storybook example of rows with multiline text.

[added](https://github.com/gravity-ui/uikit/pull/1149/commits/9dd042e12f8747c36c3c1bdc642eb1ef68726881)

Bluepuper:
Ended up aligning the content to the top of the cell
Screenshot 2023-11-22 at 12 18 53

src/components/Table/Table.scss
```diff
@@ -55,6 +55,7 @@
         border-bottom: 1px solid var(--g-color-line-generic);
 
         line-height: variables.$tableLineHeight;
+        vertical-align: top;
```
amje, line 58: That is a breaking change. We use default alignment for column (which is center). This behaviour can be changed via `verticalAlign` prop

src/components/Table/__stories__/utils.tsx
```diff
@@ -54,6 +56,17 @@ export const columns: TableColumnConfig<DataItem>[] = [
...
         id: 'name',
         name: 'Name',
+        template(item, index) {
+            if (index % 2 === 0) {
+                return item.name;
+            }
+            return (
+                <div>
+                    {item.name} <br />
+                    Lorem
```
amje, line 66: Better to split name by space instead of adding word

amje (changes requested):

src/components/Table/__stories__/utils.tsx
Bluepuper: [change](https://github.com/gravity-ui/uikit/pull/1149/files#diff-a2fac8f14d415f502e927594f887e36463ee190ef7b3a0bc0230b5a0bc19f0c4R63)

src/components/Table/Table.scss
Bluepuper: [fixed](https://github.com/gravity-ui/uikit/pull/1149/files#diff-e550fc4a251be96df10c117d97a0b9525d3ed80d4177be7d2169fa7518e4fa0cR106)

src/components/Table/hoc/withTableSelection/withTableSelection.scss
```diff
@@ -7,13 +7,19 @@
     }
 
     &__selection-checkbox {
+        display: flex;
+        align-items: center;
```
amje, line 11: Let's add an extra container for Checkbox and not put layout properties to the Checkbox itself, we can move positioning properties there as well

src/components/Table/hoc/withTableSelection/withTableSelection.scss
```diff
@@ -7,13 +7,19 @@
...
         padding: inherit;
         border-bottom: none;
         position: absolute;
         top: 0;
         bottom: 0;
         left: 0;
         right: 0;
+
+        &_verticalAlign_top {
```
amje, line 20: We prefer kebab-case in CSS

amje (changes requested):

src/components/Table/hoc/withTableSelection/withTableSelection.scss
Bluepuper: But we have to expand the checkbox label tag over the entire cell to have an ability clicking in any cell place to toggle selection. So it seems like if i will wrap that Checkbox component and center it using this wrapper - this label tag(that we need to being extended) will not take all cell place because of "padding" that centers that checkbox component

amje (approved):

PR #1152, opened 2023-11-22
feat(Table): support rtl

Raubzeug:


src/components/Table/hoc/withTableSelection/withTableSelection.scss
```diff
@@ -16,15 +16,15 @@
     &__row_selected {
         background: var(--g-color-base-selection);
 
-        #{variables.$block}__cell_sticky_left,
-        #{variables.$block}__cell_sticky_right {
+        #{variables.$block}__cell_sticky_start,
+        #{variables.$block}__cell_sticky_end {
             background: linear-gradient(
-                    to right,
+                    to end,
```
amje, line 22: This won't work, there is no support for logical values in here. And actually we don't need to specify direction at all, it doesn't matter.

amje (changes requested):

amje (approved):

PR #1160, opened 2023-11-24
feat: introduce new tooltip component

DenisVershkov:
changes compared to the previous version:
- no animation (should be approved by design team)
- smaller padding + max-width 360px
- background color changes according to the system theme (not only black)
- only String values are allowed as content
- text should cut off when height is somewhere around 360px

Previous version is moved to the ActionTooltip folder


DenisVershkov:
some strange merge happend, i will create a new PR

PR #1177, opened 2023-11-30
feat(RadioButton): allow an extended value other than a string

GermanVor:


GermanVor:
may be even `extends string` is extra

src/components/RadioButton/RadioButton.tsx
```diff
@@ -10,33 +10,37 @@ import './RadioButton.scss';
...
 
-interface RadioButtonComponent
-    extends React.ForwardRefExoticComponent<
-        RadioButtonProps & React.RefAttributes<HTMLDivElement>
-    > {
-    Option: React.ComponentType<ControlGroupOption>;
-}
+type RadioButtonComponentType = (<T extends string>(
+    props: RadioButtonProps<T> & {ref?: React.ForwardedRef<HTMLDivElement>},
+) => React.JSX.Element) & {
```
GermanVor, line 30: I do not know how to write like this

```
<T extends string>React.ForwardRefExoticComponent<
        RadioButtonProps<T> & React.RefAttributes<HTMLDivElement>
>
```

Ts does not allow

src/hooks/private/useRadioGroup/useRadioGroup.ts
```diff
@@ -67,7 +69,7 @@ export function useRadioGroup(props: UseRadioGroupProps): UseRadioGroupResult {
 
     const optionsProps = options.map((option) => ({
         name: name || controlId,
-        value: String(option.value),
```
GermanVor, line 70: I did not understand why - `type option.value === 'string'` already

amje (approved):

korvin89 (approved):

PR #1180, opened 2023-12-03
fix(Select): proper filter updates when closing the popup

DenisVershkov:


src/components/Select/Select.tsx
```diff
@@ -201,11 +201,15 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function
             if (filterable) {
                 filterRef.current?.focus();
             }
-        } else {
```
korvin89, line 204: Lets add `handleFilterChange` invocation here

korvin89 (changes requested):

korvin89 (approved):

PR #1186, opened 2023-12-05
fix: Pagination controls in Storybook

ingerman20:


jhoncool:
@ingerman20 It works for the storybook controls but you can't navigate now

PR #1193, opened 2023-12-11
feat(Progress): support customization

Lunory:


src/components/Progress/Progress.scss
```diff
@@ -63,28 +71,35 @@ $progress-xs-height: 4px;
...
+        background-color: var(
+            --g-progress-filled-background-color,
+            var(--_--filled-background-color)
+        );
 
-        &_theme_success {
-            background-color: var(--g-color-base-positive-medium);
-        }
+        &_theme {
+            &_default {
```
korvin89, line 80: Lets write this styles in more flatten way. For example `&_theme_default` and so on

src/components/CopyToClipboard/index.ts
```diff
@@ -1,2 +1,2 @@
 export * from './CopyToClipboard';
-export {CopyToClipboardStatus} from './types';
+export {type CopyToClipboardStatus} from './types';
```
korvin89, line 2: Lets rewrite this export in this way:
```typescript
export type {CopyToClipboardStatus} from './types'
```
This syntax [have released in version 4.5](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#type-modifiers-on-import-names). This export will crushed code that uses typescript below version

korvin89 (approved):

PR #1194, opened 2023-12-11
feat: add FocusTrap export

gorgeousvlad:


korvin89:
Hi, @gorgeouslady Vlad! Unfortunately, we cannot provide our focus trap realisation as a public api. U can copy-paste this code and use where u want.

PR #1199, opened 2023-12-14
feat: added useSelectAllFilter hook

Radomir-Drukh:
Added useSelectAllFilter hook which allows to quickly modify filter input to include "Select all" button


korvin89:
This functionality have added here https://github.com/gravity-ui/uikit/pull/1356

PR #1204, opened 2023-12-14
feat(Select): add component api

korvin89:


korvin89:
We have discussed this PR with @amje and decided that `apiRef` does not a good idea. There is a new idea to provide required control for select options. More details [here](https://github.com/gravity-ui/uikit/issues/1274)

cc @Radomir-Drukh

PR #1206, opened 2023-12-15
fix(Label): missing text overflow

Vladeeg:


src/components/Label/Label.scss
```diff
@@ -47,13 +47,28 @@ $hover-opacity: 0.7;
...
 
+    &__content,
+    &__key,
+    &__separator,
+    &__value {
+        vertical-align: top;
+
+        display: inline-block;
+    }
+
```
amje, line 67: It does nothing, all these elements are flex's children. Can be removed without any difference

amje (changes requested):

src/components/Label/Label.scss
Vladeeg: You're right, I guess it came from my first attempts to fix this the other way so I assumed it did something important

amje (approved):

PR #1211, opened 2023-12-20
feat(Dialog): add prop `contentOverflow` to control overflow behaviour

DenisVershkov:


DenisVershkov:
[DATAUI-1254]

korvin89:
Could u provide some description?

DenisVershkov:
1) Added prop 'scroll' to the Modal and Dialog. The prop handles scroll appearance:  
- if 'inner' is selected, then scroll will be applied to the DialogBody
- if 'outer' is selected, then scroll will appear outside (the same behaviour as before)
- 'outer' is the default value

2) Added prop 'withDivider' to the DialogBody component. When this prop has 'true' value, then the DialogBody itself will have top and bottom border. This is made for the 'inner' scroll case and should be used something like: <DialogBody withDivider={scroll === 'inner'}>

src/components/Dialog/Dialog.tsx
```diff
@@ -38,6 +38,7 @@ interface DialogOwnProps {
     disableFocusTrap?: boolean;
     disableAutoFocus?: boolean;
     restoreFocusRef?: React.RefObject<HTMLElement>;
+    scroll?: 'inner' | 'outer';
```
amje, line 41: I suggest this naming instead: `contentOverflow?: 'visible' | 'auto'`, symmetrical to the CSS overflow property

src/components/Modal/Modal.scss
```diff
@@ -102,4 +121,5 @@ $block: '.#{variables.$ns}modal';
 :root {
     --yc-modal-border-radius: 5px;
     --yc-modal-margin: 20px;
+    --yc-modal-with-inner-scroll-margin: 24px;
```
amje, line 124: From v6 we introduce public CSS API, this looks like a private. Please use SCSS var or name it like `--_--inner-scroll-margin` on component's root, not `:root`

amje (changes requested):

src/components/Dialog/Dialog.scss
```diff
@@ -5,6 +5,13 @@ $block: '.#{variables.$ns}dialog';
 #{$block} {
     position: relative;
 
+    &_with-scroll {
+        display: flex;
+        flex-direction: column;
```
amje, line 10: Let's make container always flex, not with this mod only

amje (changes requested):

src/components/Modal/Modal.tsx
```diff
@@ -129,8 +131,8 @@ export function Modal({
                 }}
             >
                 <div ref={containerRef} style={style} className={b({open}, className)} data-qa={qa}>
-                    <div className={b('table')}>
-                        <div className={b('cell')}>
+                    <div className={b('content-aligner-wrapper')}>
+                        <div className={b('content-container')}>
```
amje, line 135: Let's call them `content-aligner` and `content-wrapper`

src/components/Dialog/DialogBody/DialogBody.tsx
```diff
@@ -9,10 +9,11 @@ const b = block('dialog-body');
 export interface DialogBodyProps {
     children: React.ReactNode;
     className?: string;
+    withDividers?: boolean;
 }
 
 export function DialogBody(props: DialogBodyProps) {
-    const {className} = props;
+    const {className, withDividers = false} = props;
```
amje, line 16: We use `has*` naming throughout the codebase (i.e. `hasClear`, `hasClose`, `hasArrow`)

src/components/Dialog/Dialog.scss
```diff
@@ -4,6 +4,13 @@ $block: '.#{variables.$ns}dialog';
 
 #{$block} {
     position: relative;
+    display: flex;
+    flex-direction: column;
+
+    &_with-scroll {
+        overflow-y: auto;
+        max-height: calc(100vh - var(--yc-modal-margin) * 2);
```
amje, line 12: Let's use SCSS var instead, this var will be removed in v6

amje (changes requested):

DenisVershkov:
in order not to deal with merge conflicts, i created a new one #1251

DenisVershkov:
in order not deal with merge conflicts, i created a new one #1251

PR #1219, opened 2023-12-22
fix(Table): columns change throws

ogonkov:


src/components/Table/Table.tsx
```diff
@@ -346,9 +346,15 @@ export class Table<I extends TableDataItem = Record<string, string>> extends Rea
...
         return (
             <colgroup>
-                {columnsStyles.map(({width}, index) => (
-                    <col style={{width}} key={columns[index].id} />
-                ))}
+                {columnsStyles.flatMap(({width}, index) => {
+                    const key = columns[index]?.id;
+
+                    if (!key) {
+                        return [];
```
korvin89, line 353: Why not null value?

korvin89 (approved):

Raubzeug (approved):

PR #1222, opened 2023-12-22
feat(Table)!: remove uppercase heading style

ArturAbdullin:
Issue: for now, there is no clean way to give the title to the column with lower case first letter, like `vCPU` for instance.

Solution: Removed the uppercase transform for the first letter of the header cell. The property is uncontrollable and when removed can be easily reproduced in more controllable way with the text itself.


ogonkov:
May be we can do it via option, like `capitalizeHeading`?

ogonkov:
Oh, this is for next major, ok

amje (approved):

Raubzeug (approved):

PR #1228, opened 2023-12-27
fix(Select): activate first clickable item

ArturAbdullin:
Issue: the `activateFirstClickableItem` is not working properly. It activates the item if it is in the first place and in the disabled state.

### DIFF
--- a/src/components/Select/utils.tsx
+++ b/src/components/Select/utils.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 
 import {KeyCode} from '../../constants';
-import type {List, ListItemData} from '../List';
+import {List, type ListItemData} from '../List';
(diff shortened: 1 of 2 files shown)


src/components/Select/utils.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 
 import {KeyCode} from '../../constants';
-import type {List, ListItemData} from '../List';
+import {List, type ListItemData} from '../List';
```
korvin89, line 4: Let's use types import separately. This import corrupts code in case of using typescript below version 4.5

korvin89 (changes requested):

korvin89 (approved):

PR #1230, opened 2023-12-27
fix(Table): use default arg

ogonkov:


src/components/Table/Table.tsx
```diff
@@ -560,9 +560,9 @@ export class Table<I extends TableDataItem = Record<string, string>> extends Rea
     }
 
     private getCellStyles(
-        columnStyles: React.CSSProperties | undefined,
+        columnStyles: React.CSSProperties | undefined = {},
```
amje, line 563: undefined type is not needed then

ogonkov:
False start

PR #1235, opened 2024-01-03
fix(ThemeProvider): inner provider should overwrite specified props and only them

ValeraS:


test-utils/utils.tsx
```diff
@@ -0,0 +1,41 @@
...
+    direction: 'ltr',
+    theme: 'light',
+    themeValue: 'light',
+};
+
+function Providers({children}: {children: React.ReactElement}) {
+    return <ThemeContext.Provider value={themeContextValue}>{children}</ThemeContext.Provider>;
+}
+
+const createWrapper = memoize(
```
ogonkov, line 20: Memoization is really necessary?

ValeraS: Taking a closer look, memoization is unnecessary here.

ogonkov:
We are just add wrapper with theme context to tests here?

styles/themes/common/_index.scss
```diff
@@ -14,4 +14,23 @@
     --g-border-radius-xl: 10px;
 
     --g-focus-border-radius: 2px;
+
+    @at-root body#{&} {
```
amje, line 18: The root can be on any element, not only the body. Also these styles should go into styles.scss, not here

amje (changes requested):

styles/themes/common/_index.scss
ValeraS: These defaults only for the class on the body, if present, any other element always has `dir` attribute if it set to `ThemeProvider`.

amje (approved):

PR #1238, opened 2024-01-09
feat(TextInput/TextArea): prevent zoom on iOS

ogonkov:


src/components/controls/variables.scss
```diff
@@ -1 +1,4 @@
 $border-width: 1px;
+// Safari on iOS zoom page if input font-size is smaller than 16px
+// https://css-tricks.com/16px-or-larger-text-prevents-ios-form-zoom/
+$mobile-safari-font-size: 16px;
```
korvin89, line 4: I suggest u to rename this to $mobile-font-size case of it uses not only for safari styles

ogonkov: May be `$mobile-min-font-size` then?

korvin89: But we do not have a max size or other size at all

ogonkov: But we do have minimal size, it's not just any mobile font size, it is a minimal possible size for Safari

korvin89: ok. It's not a public api

korvin89 (approved):

src/components/controls/TextArea/TextArea.scss
```diff
@@ -9,6 +10,10 @@ $block: '.#{variables.$ns-new}text-area';
     width: 100%;
     position: relative;
 
+    &_mobile {
+        font-size: control-variables.$mobile-safari-font-size;
```
amje, line 14: Not sure if it's a good idea to hardcode 16px for all the sizes in mobile devices

korvin89: It's really annoying bug. Why not?

amje (changes requested):
Setting 16px to all sizes is a bad workaround

ogonkov:
> Setting 16px to all sizes is a bad workaround

Do you know better?

amje:
> > Setting 16px to all sizes is a bad workaround
> 
> Do you know better?

Revise font sizes in all components for l/xl size to be at least 16px. But it's cannot be done right away, we need some time to think about it

PR #1240, opened 2024-01-10
feat(Alert): changed default Alert actions background

IsaevAlexandr:
UXRFC-354

### DIFF
--- a/src/components/Alert/Alert.tsx
+++ b/src/components/Alert/Alert.tsx
@@ -8,7 +8,9 @@ import {Icon} from '../Icon';
 import {colorText} from '../Text';
 import {Flex, spacing} from '../layout';
 
+import {AlertAction} from './AlertAction';
 import {AlertActions} from './AlertActions';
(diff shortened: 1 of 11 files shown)


src/components/Alert/Alert.tsx
```diff
@@ -28,48 +30,50 @@ export const Alert = (props: AlertProps) => {
...
-        </Card>
+            </Card>
+        </AlertContextProvider>
     );
 };
 
 Alert.Icon = AlertIcon;
 Alert.Title = AlertTitle;
 Alert.Actions = AlertActions;
+Alert.Action = AlertAction;
```
amje, line 79: If it's a public API, then we should export types for these

IsaevAlexandr: It is assumed that static component methods cannot be used in isolation from the component. So only `AlertProps` type is exported

amje (approved):

PR #1244, opened 2024-01-15
fix: revert #831

amje:
This change was breaking. Colgroup doesn't take into account `box-sizing` property. We use `content-box`, but colgroup sets the width for the entire column, not only the content.


ogonkov (changes requested):
Add motivation for revert please

ogonkov (changes requested):
Fix in docs, that we change content with, not the whole column width

ogonkov (approved):
Agreed to change docs in separate PR

Raubzeug (approved):

PR #1247, opened 2024-01-16
fix: remove extra allocations in List

GermanVor:
before fix each rerender triggers `n` times [renderItem](https://github.com/gravity-ui/uikit/blob/252229e6498effe9628d193103d9e5117b2bd79b/src/components/List/List.tsx#L253), `renderItem` will trigger [getItemsWithLoading](https://github.com/gravity-ui/uikit/blob/252229e6498effe9628d193103d9e5117b2bd79b/src/components/List/List.tsx#L266), `getItemsWithLoading` in case of `this.props.loading` allocates new array with size `n + 1`.

to summarize: each rerender in case of `this.props.loading` allocates `n*(n+1)` bytes (plus memory to support `n` arrays)


GermanVor:
I would like to get rid of `[...items, LOADING_ITEM]` and write something like `<>{items.map(this.renderItem)}{props.loading && <LoadingPlugin />}<>`, but these are big changes and I chose the simplest way.

GermanVor:
@korvin89  updated the description

korvin89:
Could u show memory calculations?

GermanVor:
@korvin89 

calculations here - https://github.com/gravity-ui/uikit/pull/1247#issue-2084599817

proofs:
telegram-cloud-document-4-5852615959889580921

telegram-cloud-document-4-5852615959889580920

korvin89:
I am talking about some numbers from browser. I want to understand why we need this changes.
There is almost a new List component and we wouldn't like to add any changes here except bug fixes.

GermanVor:
@korvin89 

In my opinion, the allocation of extra arrays - is a bug. 
Especially when this is done with every render.

`I am talking about some numbers from browser` - you can copy paste code from the screenshots and add [here](https://github.com/gravity-ui/uikit/blob/252229e6498effe9628d193103d9e5117b2bd79b/src/components/List/__stories__/ListShowcase.tsx#L89) `loading = {true}`, then if you open this Storybook of List.Showcase you will see the same console output. Each time when you will see `Аллокация массива размером` means allocation - `return [...items, loadingPlug]`.

I know about GB, I know about pointers - in any case this is potentially a lot of unnecessary allocations, memory purges and memory used for all arrays.

GermanVor:
I don't know anything about the new List, so I can't comment on it in any way.

korvin89:
Okay. I suggest u to add test cases for loader using and then fix this behaviour. Is it okay for u?

GermanVor:
@korvin89 added test

PR #1250, opened 2024-01-17
[WIP] feat(table): add getRowDescriptor property

GermanVor:
**Description of the problem**
The Table component now has several methods, each of which interacts with the display of the string -- getRowId, getRowClassNames, isRowDisabled.

It seems that it would be logical to make one method instead of 3, which will improve the user experience.

In the future, this method can be expanded, for example, you can add a description of whether the line should change the background on hover.

**The proposed solution**
Set up all 3 props, add a new one.

https://st.yandex-team.ru/UXRFC-393

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -53,8 +53,7 @@ export interface TableColumnConfig<I> {
 export interface DescriptorType {
     /**
      * Row ID.
-     * Used when selecting and sorting rows. If you pass a row,
-     * its ID will be the value of the field in the row data named the same as the column ID.
+     * Used when selecting and sorting rows.
      */
     id?: string;
 
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -50,6 +50,25 @@ export interface TableColumnConfig<I> {
     meta?: Record<string, any>;
 }
 
+export interface DescriptorType {
+    /**
+     * Row ID.
+     * Used when selecting and sorting rows. If you pass a row,
+     * its ID will be the value of the field in the row data named the same as the column ID.
+     */
+    id?: string;
+
+    /**
+     * Row CSS classes.
+     */
+    classNames?: string[];
+
+    /**
+     * Condition for disabling columns.
+     */
+    disabled?: boolean;
+}
+
 // TODO: Replace @default in props description with defaultProps in order to work with Storybook.
 export interface TableProps<I> extends QAProps {
     /** Data */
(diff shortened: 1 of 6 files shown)


src/components/Table/hoc/withTableSelection/withTableSelection.tsx
```diff
@@ -50,6 +51,7 @@ export function withTableSelection<I extends TableDataItem, E extends {} = {}>(
                     columns={this.enhanceColumns(columns)}
                     onRowClick={this.enhanceOnRowClick(onRowClick)}
                     getRowClassNames={this.enhanceGetRowClassNames(getRowClassNames)}
```
Raubzeug, line 53: If we pass `getRowDescriptor`, it is no need to pass `getRowClassNames`.

Raubzeug:
https://github.com/gravity-ui/uikit/blob/dc062af8f514a570dcbacaa7a61dfb05a21a4b96/src/components/Table/hoc/withTableSelection/withTableSelection.tsx#L241 - getRowDescriptor should be used here too

Raubzeug:
Also could you please add tests and stories with new property?

Raubzeug (changes requested):

Raubzeug:
https://github.com/gravity-ui/uikit/blob/1c551f72df06bde4b6fa7745e501ce1586f9eb69/src/components/Table/hoc/withTableActions/withTableActions.tsx#L118 - new property should also be used here.

src/components/Table/hoc/withTableSelection/withTableSelection.tsx
GermanVor: but we still want support old props for a while

src/components/Table/Table.tsx
```diff
@@ -117,7 +117,7 @@ export interface TableProps<I> extends QAProps {
      *
      * @returns {DescriptorType} {@link DescriptorType}
      */
-    getRowDescriptor?: (item: I, index: number) => DescriptorType;
+    getRowDescriptor?: (item: I, index: number) => DescriptorType | undefined;
```
Raubzeug, line 120: Why? If `getRowDescriptor` is passed, it should definitely return something.

src/components/Table/hoc/withTableSelection/withTableSelection.tsx
Raubzeug: Yeah, but it's internal <TableComponent> call. So we may take `getRowClassNames` into account inside `enhanceGetRowDescriptor`.

Raubzeug (changes requested):

src/components/Table/Table.tsx
GermanVor: `it should definitely return something` - it will be more convenient  in case

```
const getRowDescriptor = (item) => {
    if (isValidRow(item)) {
        return {...}
    }
    
    // otherwise i do not want even return empty object
}
```

Raubzeug (changes requested):
Please check story for table with selection. It fails with error `this.enhanceGetRowClassNames is not a function`

Raubzeug (approved):

korvin89 (approved):

PR #1251, opened 2024-01-17
feat(Dialog): add prop contentOverflow to control overflow behaviour

DenisVershkov:
1. Added prop 'contentOverflow' to the Modal and Dialog. The prop handles scroll appearance:

- if 'auto' is selected, then scroll will be applied to the DialogBody
- if 'visible' is selected, then scroll will appear outside (the same behaviour as before)
- 'visible' is the default value

2. Added prop 'hasDivider' to the DialogBody component. When this prop value is 'true', then DialogBody has top and bottom border. I've done this for the contentOverflow 'auto' case, so it should be used like: <DialogBody hasDivider={contentOverflow === 'auto'}>


src/components/Dialog/Dialog.scss
```diff
@@ -4,6 +4,15 @@ $block: '.#{variables.$ns}dialog';
 
 #{$block} {
     position: relative;
+    display: flex;
+    flex-direction: column;
+
+    &_has-scroll {
+        overflow-y: auto;
+        max-height: calc(
+            100vh - calc(var(--g-modal-margin, #{variables.$modal-default-margin})) * 2
```
amje, line 13: 2nd calc is just wraps var?

src/components/Dialog/DialogBody/DialogBody.tsx
```diff
@@ -9,10 +9,11 @@ const b = block('dialog-body');
 export interface DialogBodyProps {
     children: React.ReactNode;
     className?: string;
+    hasDividers?: boolean;
```
amje, line 12: Wouldn't be `hasBorders` more precise? Divider - smth that splits the content, border - is on the edge. While it's a `DialogBody` prop, borders suits better imo.

src/components/Modal/Modal.scss
```diff
@@ -16,24 +16,29 @@ $block: '.#{variables.$ns}modal';
...
     &__content {
-        display: inline-block;
-        margin: var(--g-modal-margin, 20px);
-        text-align: start;
         background-color: var(--g-color-base-modal);
-        border-radius: var(--g-modal-border-radius, 5px);
+        &_has-scroll {
+            overflow-y: auto;
+            max-height: calc(
+                100vh - calc(var(--g-modal-margin, #{variables.$modal-default-margin})) * 2
```
amje, line 39: Same comment as above

src/components/variables.scss
```diff
@@ -5,3 +5,5 @@ $s-height: 24px;
 $m-height: 28px;
 $l-height: 36px;
 $xl-height: 44px;
+$modal-default-border-radius: 5px;
```
amje, line 8: While we're only using it inside the Modal, we don't need to share it.

amje (changes requested):

amje (approved):

DenisVershkov:
closes [DATAUI-1254]

PR #1253, opened 2024-01-19
feat(Disclosure): added qa attribute for details and summary nodes

vrozaev:


Raubzeug (changes requested):
Should added attributes be hard-coded? May be we should be able to pass them via props?

vrozaev:
> Should added attributes be hard-coded? May be we should be able to pass them via props?

Hi @Raubzeug!  Good point, I used next code as reference here:

1. https://github.com/gravity-ui/uikit/blob/1c551f72df06bde4b6fa7745e501ce1586f9eb69/src/components/Table/Table.tsx#L490
2. https://github.com/gravity-ui/uikit/blob/1c551f72df06bde4b6fa7745e501ce1586f9eb69/src/components/Select/components/SelectClear/SelectClear.tsx#L26

The idea is: we need only one `qa` prop for the component itself, all other qa attributes can be hardcoded because we can build an unique selector anyway:

```
<Disclosure qa={'uniq-disclosure-id'} />

...

await page.locator('uniq-disclosure-id').locator('disclosure-details').getText();
```

Of course we can add more props, but I don't see much benefits from that, aside of guarantee that attributes values will not change in future.

Raubzeug:
Oh, I've got it! Agreed, thanks for the explanation!

Raubzeug (approved):

PR #1270, opened 2024-01-23
feat(Link)!: the "href" property is required

LakeVostok:
UXRFC-1


src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -12,13 +12,22 @@ import {BreadcrumbsSeparator} from './BreadcrumbsSeparator';
...
+
+type LinkBreadcrumbsItem = {
+    href: string;
+} & BaseBreadcrumbsItem;
+
+type ButtonBreadcrumbsItem = {
+    href?: undefined;
+} & BaseBreadcrumbsItem;
+
+export type BreadcrumbsItem = LinkBreadcrumbsItem | ButtonBreadcrumbsItem;
```
amje, line 30: Isn't the same as before?

src/components/Breadcrumbs/BreadcrumbsButton.tsx
```diff
@@ -0,0 +1,13 @@
...
+import {block} from '../utils/cn';
+
+const b = block('breadcrumbs');
+
+export function BreadcrumbsButton(props: {
+    title: string;
+    onClick: React.MouseEventHandler<HTMLElement>;
+    children: React.ReactNode;
+}) {
+    return <button {...props} className={b('switcher', {more: true})} />;
```
amje, line 12: button must have the `type` attribute with the `"button"` value

amje: And add please focus styling to the button

src/components/Link/README.md
```diff
@@ -159,32 +156,32 @@ LANDING_BLOCK-->
...
-| className  | HTML `class` attribute                     |                            `string \| undefined`                             |
-| qa         | HTML `data-qa` attribute, used for testing |                            `string \| undefined`                             |
-| ref        | React ref to Link DOM node                 |                `React.ForwardedRef<HTMLElement> \| undefined`                |
+| Name       | Description                                |                           Type                            |  Default   |
+| :--------- | :----------------------------------------- | :-------------------------------------------------------: | :--------: |
+| view       | Link appearance                            |          `"normal" \| "primary" \| "secondary"`           | `"normal"` |
+| visitable  | Display `:visitable` CSS state             |                  `boolean \| undefined`                   |
+| href       | HTML `href` attribute                      |                   `string \| undefined`                   |
```
amje, line 174: table formatting broke

src/components/Link/__stories__/Link.new.stories.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 
 import type {Meta, StoryFn} from '@storybook/react';
 
-import {Link} from '../Link';
+import {Link, type LinkProps} from '../Link';
```
amje, line 5: We're not using this TS syntax

amje (changes requested):

src/components/Breadcrumbs/Breadcrumbs.tsx
LakeVostok: It was types product, became types sum. Its more powerful way for union declaration.

src/components/Link/__stories__/Link.new.stories.tsx
LakeVostok: Well, then it should be regulated by the linter. The language provides me with tools, and I use them. In addition, this constuction is used in other components.
I can set up an appropriate rule to ban inline important. But at first glance it looks strange. Why did you decide not to use it?

src/components/Breadcrumbs/Breadcrumbs.tsx
LakeVostok: > I mean there is no difference, `href` is optional anyway. What's the point of this change?

Oh, I figure out, why it wasn't obvious to use this kind of approach. Lost it while rebasing, it should have been:
```
type BaseBreadcrumbsItem = {
    text: string;
    items?: BreadcrumbsItem[];
    title?: string;
};

type LinkBreadcrumbsItem = {
    href: string;
    action?: (event: React.MouseEvent<HTMLElement, MouseEvent> | KeyboardEvent) => void;
} & BaseBreadcrumbsItem;

type ButtonBreadcrumbsItem = {
    href?: undefined;
    action: (event: React.MouseEvent<HTMLElement, MouseEvent> | KeyboardEvent) => void;
} & BaseBreadcrumbsItem;
```
So, we describe 2 categories of breadcrumb items: the one is like regular anchor(and should have the href prop), the two is like button, that should have some action(action prop).
So, if user want to create Breadcrumb, that is not like regular link, he will be forced(by typescript) to specify the action prop.
And if the developer(of uikit) needs to add a new behavior for only one of the types of breadcrumb item(e.g. the target attribute for anchor), then he modifies only LinkBreadcrumbsItem and any time in the code he will know, that he works with certain type of item(due to the href prop check).

src/components/Link/__stories__/Link.new.stories.tsx
amje: I remember the main reason was that it requires to use TS 4.5+

amje (approved):

PR #1278, opened 2024-01-28
Add hook sync with local storage

eyoue:
Perform action in multiple tabs simultaneously using Local Storage key and 'storage' event listener.

How to use:
```
export const App = () => {
  const [open, setOpen] = useState(false);

  const close = useCallback(() => setOpen(true), []);
  const handleClose = useCallback(() => {
    close();
    closeWithLS();
  }, []);
  const { callback: closeWithLS } = useSyncWithLS({
    callback: close,
    uniqueKey: 'close-key',
  });
  return (
    <div>
      <button onClick={handleClose}>Close</button>
      <div>{open ? 'Open' : 'Close'}</div>
    </div>
  );
};
```

### DIFF
--- a/src/hooks/useSyncWithLS/README.md
+++ b/src/hooks/useSyncWithLS/README.md
@@ -1,0 +1,19 @@
+<!--GITHUB_BLOCK-->
+
+# useSyncWithLS
+
+<!--/GITHUB_BLOCK-->
+
+```tsx
+import {useSyncWithLS} from '@gravity-ui/uikit';
+```
+
+The `useSyncWithLS` hook executes callback when value changed in Local Storage
+
+## Properties
+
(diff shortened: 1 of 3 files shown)


ogonkov:
Why only localStorage? There is also [`BroadcastChannel`](https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel), that was designed for communication between tabs

korvin89:
I'm not sure if this should be part of the public api of the library with basic components

eyoue:
> Why only localStorage? There is also [`BroadcastChannel`](https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel) API, that was designed for communication between tabs

BroadcastChannel not supported in erly version IE, Safari

PR #1284, opened 2024-01-29
fix(List): calling onLoadMore function while using keyboard

### LINKED ISSUE #1105: List: Call OnLoadMore when navigating from the keyboard
Now if you navigate through the List using the arrows, OnLoadMore is not called.

kkirik:
Closes #1105


korvin89 (commented):
Could u add height to main container to avoid such situation?

https://github.com/gravity-ui/uikit/assets/26501073/8ba13dff-3f32-48c1-a7c1-d6ac93b55654

korvin89 (approved):

PR #1285, opened 2024-01-29
chore(a11y): allow `ul` role override

ogonkov:


amje:
@ogonkov What is the story behind this change? No code changed, only the rule

ogonkov:
@amje i added this rule in my other PR, decided to split it as separate PR

ogonkov:
We decide that without comment near `role=list` this attribute could be removed, so it is better to use inline ignore instead

PR #1291, opened 2024-01-30
feat(Select): added errorMessage, errorPlacement and validationType props

### LINKED ISSUE #1273: Select: add error view like in TextInput component
- deprecate property [error](https://github.com/gravity-ui/uikit/blob/main/src/components/Select/types.ts#L85)
- add property [errorMessage](https://github.com/gravity-ui/uikit/blob/main/src/components/controls/types.ts#L39)
- add property [errorPlacement](https://github.com/gravity-ui/uikit/blob/main/src/components/controls/types.ts#L39)
- add property [validationState](https://github.com/gravity-ui/uikit/blob/main/src/components/controls/types.ts#L43C9-L43C24)

kkirik:
Closes #1273

### DIFF
--- a/src/components/Select/components/SelectControl/SelectControl.tsx
+++ b/src/components/Select/components/SelectControl/SelectControl.tsx
@@ -5,9 +5,9 @@ import isEmpty from 'lodash/isEmpty';
 
 import {Icon} from '../../../Icon';
 import {Popover} from '../../../Popover';
-import {CONTROL_ERROR_ICON_QA} from '../../../controls/utils';
 import type {CnMods} from '../../../utils/cn';
 import {selectControlBlock, selectControlButtonBlock} from '../../constants';
+import i18n from '../../i18n';
 import type {
     SelectProps,
     SelectRenderClearArgs,
@@ -170,7 +170,7 @@ export const SelectControl = React.forwardRef<HTMLButtonElement, ControlProps>((
 
                 {errorMessage && (
                     <Popover content={errorMessage}>
-                        <span data-qa={CONTROL_ERROR_ICON_QA}>
+                        <span aria-label={i18n('label_show-error-info')}>
                             <Icon
                                 data={TriangleExclamation}
                                 className={selectControlBlock('error-icon')}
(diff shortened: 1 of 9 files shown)


src/components/Select/components/SelectControl/SelectControl.tsx
```diff
@@ -1,10 +1,12 @@
 import React from 'react';
 
-import {ChevronDown} from '@gravity-ui/icons';
+import {ChevronDown, TriangleExclamation} from '@gravity-ui/icons';
 import isEmpty from 'lodash/isEmpty';
+import type {CnMods} from 'src/components/utils/cn';
```
korvin89, line 5: We don't support aliases. Btw, CI tests failed because of it

kkirik: Thanks. I'll check

korvin89:
One more thing: could u add some tests for this features? U can get tests from [here](https://github.com/gravity-ui/uikit/blob/main/src/components/controls/TextInput/__tests__/TextInput.input.test.tsx#L91). And add QA constant [here](https://github.com/gravity-ui/uikit/blob/main/src/components/Select/constants.ts#L34) please. And there is some different behaviour regarding TextInput (in [this](https://github.com/gravity-ui/uikit/blob/main/src/components/controls/TextInput/__tests__/TextInput.input.test.tsx#L98) test case, but it has a confusing name :))

kkirik:
> One more thing: could u add some tests for this features? U can get tests from [here](https://github.com/gravity-ui/uikit/blob/main/src/components/controls/TextInput/__tests__/TextInput.input.test.tsx#L91). And add QA constant [here](https://github.com/gravity-ui/uikit/blob/main/src/components/Select/constants.ts#L34) please. And there is some different behaviour regarding TextInput (in [this](https://github.com/gravity-ui/uikit/blob/main/src/components/controls/TextInput/__tests__/TextInput.input.test.tsx#L98) test case, but it has a confusing name :))

So, I added some tests for error feature also I added an aria-label attribute for getting internal error icon by i11y attributes.

korvin89 (approved):

PR #1292, opened 2024-01-31
Feat: refactor Persona and rename it to UserLabel

DakEnviy:


src/components/UserLabel/README.md
```diff
@@ -0,0 +1,126 @@
...
+<!--/GITHUB_BLOCK-->
+
+## Properties
+
+| Name      | Description                                                 |                                    Type                                     |   Default    |
+| :-------- | :---------------------------------------------------------- | :-------------------------------------------------------------------------: | :----------: |
+| type      | Avatar appearance                                           |                       `'person'` `'email'` `'empty'`                        |  `'person'`  |
+| avatar    | User avatar                                                 | [AvatarProps](../Avatar/README.md#properties) `string` `React.ReactElement` |              |
+| text      | Visible text                                                |                                  `string`                                   |              |
+| size      | Text size                                                   |                                 `'s'` `'m'`                                 |    `'s'`     |
```
amje, line 120: Let's remove this prop, it confuses not scaling size, but font-size

src/components/UserLabel/UserLabel.scss
```diff
@@ -0,0 +1,138 @@
...
+        #{$block}_clickable & {
+            outline-offset: -1px;
+
+            &:focus {
+                outline: 2px solid var(--g-color-line-focus);
+            }
+
+            &:focus:not(:focus-visible) {
+                outline: 0;
+            }
```
amje, line 75: We can freely use only `focus-visible`

src/components/UserLabel/UserLabel.scss
```diff
@@ -0,0 +1,138 @@
...
+    &__close-icon {
+        border-radius: var(--g-focus-border-radius);
+
+        #{$block}__close:focus & {
+            outline: 2px solid var(--g-color-line-focus);
+        }
+
+        #{$block}__close:focus:not(:focus-visible) & {
+            outline: 0;
+        }
```
amje, line 136: Same as above, and why it's on the icon, not button itself?

src/components/UserLabel/UserLabel.tsx
```diff
@@ -0,0 +1,102 @@
...
+    const textValue = extractTextValue(text);
+    const textView = extractTextView(text);
+
+    const closeButtonAriaAttributes: React.AriaAttributes = {
+        'aria-label': i18n('label_remove-button', {text: textValue}),
+    };
+
+    const clickable = Boolean(onClick);
+    const closeable = Boolean(onClose);
+    const MainComponent = clickable ? 'button' : 'div';
```
amje, line 43: `'button'` should be with `type="button"` attribute in order to work correctly inside forms

src/components/UserLabel/UserLabel.tsx
```diff
@@ -0,0 +1,102 @@
...
+            )}
+            style={style}
+            data-qa={qa}
+        >
+            <MainComponent className={b('main')} onClick={handleClick}>
+                {avatarView ? <div className={b('avatar')}>{avatarView}</div> : null}
+                <div className={b('text')}>{textView}</div>
+            </MainComponent>
+            {onClose ? (
+                <button className={b('close')} onClick={handleClose} {...closeButtonAriaAttributes}>
```
amje, line 96: type="button"

src/components/UserLabel/types.ts
```diff
@@ -0,0 +1,23 @@
+import type React from 'react';
+
+import type {DistributiveOmit} from '../../types/utils';
+import type {AvatarProps} from '../Avatar';
+import type {DOMProps, QAProps} from '../types';
+
+export type UserLabelType = 'person' | 'email' | 'empty';
+export type UserLabelText = {value: string; content: React.ReactNode} | string;
```
amje, line 8: Can it be just `string | React.ReactNode`?

src/components/UserLabel/types.ts
```diff
@@ -0,0 +1,23 @@
...
+    type?: UserLabelType;
+    avatar?:
+        | DistributiveOmit<AvatarProps, 'size' | 'className'>
+        | string
+        | React.ReactElement<unknown>;
+    text: UserLabelText;
+    size?: UserLabelSize;
+    view?: UserLabelView;
+    onClick?: (value: string) => void;
+    onClose?: (value: string) => void;
```
amje, line 22: DOM handlers should have native event, like other components do. `value` looks redundant in this component

amje (changes requested):

src/components/UserLabel/types.ts
DakEnviy: No, value is required because it is used for passing parameters to `onClick` and `onClose` callbacks.

src/components/UserLabel/types.ts
DakEnviy: `value` can be useful for cases when we render components from lists

src/components/UserLabel/UserLabel.scss
DakEnviy: I fixed `:focus-visible`. We show a border for the icon because the bounding box for the button is too big and It looks worse.

src/components/UserLabel/UserLabel.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+import i18n from './i18n';
+import type {UserLabelProps} from './types';
+import {extractTextValue, extractTextView} from './utils';
+
+import './UserLabel.scss';
+
+const COMMON_AVATAR_PROPS: Partial<AvatarProps> = {
+    size: 's',
+};
```
amje, line 18: This can be removed I guess?

DakEnviy: I think it is better to fixate avatar size

src/components/UserLabel/UserLabel.tsx
```diff
@@ -0,0 +1,97 @@
...
+    avatar,
+    text,
+    view = 'outlined',
+    onClick,
+    onClose,
+    className,
+    style,
+    qa,
+}: UserLabelProps) => {
+    const closeButtonAriaAttributes: React.AriaAttributes = {
```
ogonkov, line 32: variable seems redundant

src/components/UserLabel/UserLabel.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+const b = block('user-label');
+
+export const UserLabel = ({
+    type = 'person',
+    avatar,
+    text,
+    view = 'outlined',
+    onClick,
+    onClose,
```
ogonkov, line 27: may be `onButtonClick`?

src/components/UserLabel/__stories__/UserLabel.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export default meta;
+
+type Story = StoryObj<typeof UserLabel>;
+
+const person = 'Charles Darwin';
+const email = faker.internet.email(...person.split(' '));
+const personImg = faker.internet.avatar();
+
+export const Default: Story = {
```
ogonkov, line 19: Add story with text that overflow in narrow width container

src/components/UserLabel/UserLabel.tsx
amje: why?

src/components/UserLabel/UserLabel.tsx
```diff
@@ -0,0 +1,97 @@
...
+            } else if (React.isValidElement(avatar)) {
+                avatarView = avatar;
+            } else if (avatar) {
+                avatarView = <Avatar {...avatar} {...COMMON_AVATAR_PROPS} />;
+            }
+            break;
+    }
+
+    return (
+        <div
```
ogonkov, line 64: Can we forward ref to this element please?

src/components/UserLabel/UserLabel.tsx
ogonkov, line 1: Most of the code here is inherited from Persona, but history is removed and component re-added to git, it cause bad developer experience

amje (approved):

PR #1293, opened 2024-01-31
fix(TextInput): add overflow hidden to text-input__content wrapper

shvorak:
Hi

In some browsers (MIcrosoft Edge and others) we have encountered an issue where the input is autocompleted and displays the background outside of the input edges.

PS: Solving this problem by adding a border radius to the input doesn't work as expected, at least because it doesn't match the content radius.

| AS IS | TO BE | With radius |
|--------|--------|--------|
| image | image | image |

### DIFF
--- a/src/components/controls/TextInput/TextInput.scss
+++ b/src/components/controls/TextInput/TextInput.scss
@@ -44,6 +44,10 @@ $block: '.#{variables.$ns}text-input';
 
     &__content {
         box-sizing: border-box;
+        // It's fixing input overflowed behaviour
+        // Input itself doesn't have a border radius, but content has.
+        // And it could cause a render problem,
+        // when input accidentally goes beyond the wrapper
         overflow: hidden;
         display: flex;
         width: 100%;
--- a/src/components/controls/TextInput/TextInput.scss
+++ b/src/components/controls/TextInput/TextInput.scss
@@ -44,6 +44,7 @@ $block: '.#{variables.$ns}text-input';
 
     &__content {
         box-sizing: border-box;
+        overflow: hidden;
         display: flex;
         width: 100%;
     }


korvin89:
Hey @shvorak ! Just for clarifying - u don't set this bg color via css styles?

ogonkov:
This seems like breaking change

input should have border-radius of parent container - border width of container, than it will match perfectly

shvorak:
> Hey @shvorak ! Just for clarifying - u don't set this bg color via css styles?

Yeah, it's just native browser style

shvorak:
> This seems like breaking change
> 
> input should have border-radius of parent container - border width of container, than it will match perfectly

I guess we are misunderstood each over. Input width can't be the same as content width. 
.yc-text-input__content has a border, which increasing it's size.

Could you please tell me that the problem with overflow: hidden?

shvorak:
> > Yeah, it's just a native browser style
> 
> Great. Could u try [this](https://github.com/gravity-ui/uikit/pull/1293#issuecomment-1919298822) approach? `overflow: hidden` pretty dangerous style, we would not like to use it unnecessarily

Ok, the only solution I know is to calculate the border radius for the input.
```
border-radius: calc(var(--_--text-input-border-radius) - var(BORDER_WIDTH_OF_CONTENT));
```
This solution will work. You can try by yourself

korvin89:
> Could you please tell me that the problem with overflow: hidden?

If u will try to use `leftContent` or `rightContent` with node, that will be an anchor for Popup, that will be rendered with `disablePortal: true`, u could not see the most part of this Popup.

But I've check solution with radiuses and it looks to complicated. Let's try ur solution.

Add comment for this new style please.

shvorak:
Hi @korvin89 
Comment was added. 
Afaik `TextArea` cannot be filled the same way as an `input`, so we can't reproduce same behavior. I tried :)

And also here is the example with tooltip on icon inside `TextInput`
image

korvin89 (approved):

PR #1295, opened 2024-01-31
feat!: rename onClose to onCloseClick

amje:
onClose was a bad naming, it has nothing to do with closing


ogonkov (approved):
Still think that it should be `onButtonClick`

DakEnviy (approved):

PR #1296, opened 2024-01-31
refactor(styles): set CSS vars in `pin` mixin

ogonkov:
This will set CSS-vars in private namespace, that will allow to developer reuse `border-radius` in children elements, for example to match nested element `border-radius` by reducing it via `calc()`.

### DIFF
--- a/styles/mixins.scss
+++ b/styles/mixins.scss
@@ -92,20 +92,20 @@ $-pin-mixin-variables-ns: --_--g-pin-mixin-border-radius;
 
     @each $selector in $selectors {
         &#{get-pin-selector(#{$block}_pin_round-round, $selector, $append)} {
-            @include -pin-mixin-variables-from-list($radius $radius $radius $radius);
+            @include -pin-mixin-variables-from-list($radius);
         }
 
         &#{get-pin-selector(#{$block}_pin_brick-brick, $selector, $append)} {
-            @include -pin-mixin-variables-from-list(0 0 0 0);
+            @include -pin-mixin-variables-from-list(0);
         }
 
         &#{get-pin-selector(#{$block}_pin_clear-clear, $selector, $append)} {
-            @include -pin-mixin-variables-from-list(0 0 0 0);
+            @include -pin-mixin-variables-from-list(0);
             border-inline: 0;
         }
 
         &#{get-pin-selector(#{$block}_pin_circle-circle, $selector, $append)} {
-            @include -pin-mixin-variables-from-list(100px 100px 100px 100px);
+            @include -pin-mixin-variables-from-list(100px);
         }
 
         &#{get-pin-selector(#{$block}_pin_round-brick, $selector, $append)} {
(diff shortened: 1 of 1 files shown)


styles/mixins.scss
```diff
@@ -70,69 +72,92 @@
     }
 }
 
+$-pin-mixin-variables-ns: --_--g-pin-mixin-border-radius;
+
+@mixin -pin-mixin-variables-from-list($list) {
```
amje, line 77:
```suggestion
@mixin pin-variables-from-list($list) {
```

styles/mixins.scss
```diff
@@ -70,69 +72,92 @@
     }
 }
 
+$-pin-mixin-variables-ns: --_--g-pin-mixin-border-radius;
```
amje, line 75:
```suggestion
$pin-variables-ns: --_--g-pin-mixin-border-radius;
```

amje (changes requested):

styles/mixins.scss
ogonkov: This will make this variable public, is this really your intention?

styles/mixins.scss
```diff
@@ -63,92 +65,94 @@
...
+}
+
 @mixin pin($block, $selectors, $radius, $append: true) {
+    @each $selector in $selectors {
+        &#{get-pin-selector($block, $selector, $append)} {
+            border-radius: (
+                var(#{$-pin-mixin-variables-ns}-top-left)
+                    var(#{$-pin-mixin-variables-ns}-top-right)
+                    var(#{$-pin-mixin-variables-ns}-bottom-right)
+                    var(#{$-pin-mixin-variables-ns}-bottom-left)
```
amje, line 90: RTL support degradation

styles/mixins.scss
```diff
@@ -63,92 +65,94 @@
...
+            );
+        }
+    }
+
     @each $selector in $selectors {
         &#{get-pin-selector(#{$block}_pin_round-round, $selector, $append)} {
-            border-radius: $radius;
+            @include -pin-mixin-variables-from-list($radius);
```
amje, line 97: This would break customization via CSS API, see "Custom" story of `Button`

amje (changes requested):

styles/mixins.scss
ogonkov: Why?

amje: physical properties instead of logical ones

ogonkov: But property is `border-radius`, it is rtl compatible

amje: There should be not such api like `top-left`, `bottom-right`. Instead we should only use logical names: `start-start`, `end-end`, etc.

ogonkov: Oh, i will rename vars, no problem

amje:
@ogonkov Is PR still relevant?

ogonkov:
I think yes, i will try to rebase and fix bugs if any

styles/mixins.scss
ogonkov: Yes, 2 values syntax for `--g-button-border-radius` makes my refactoring pointless

PR #1304, opened 2024-02-02
feat: add Palette component

Ruminat:
DATAUI-2067

`Palette` — a basic grid with elements you can select/unselect (a grid of checkboxes basically).

image

Can be customized to display anything basically icons/GIFs/images/reactions/colors and so on:

image


korvin89:
Hey @Ruminat! We've discussed ur pr with @amje @ValeraS and have some suggestions for u:
1. Add separate component `Palette` with this interface:
```typescript
interface PaletteOption {
    value: string;
    content: React.ReactNode;
    title?: string; // HTML title
    disabled?: boolean;
}

interface PaletteProps extends DOMProps, QAProps {
    value?: string[];
    defaultValue?: string[];
    onUpdate?: (value: string[]) => void;
    onFocus?: React.FocusEventHandler;
    onBlur?: React.FocusEventHandler;
    disabled?: boolean;
    name?: string;
    options: PaletteOption[];
    size?: ButtonSize;
}
```
`Palette` should have fully keyboard support:
- the first press of the tab key should focus first option in the component
- the second press of the tab key should take the focus away from the component
- navigation between options should be managed by arrow keys
2. Use `Palette` as a base for `EmojiPalette` 
3. Use [this](https://github.com/gravity-ui/uikit/tree/main/src/hooks/useControlledState) hook for value management

Ruminat:
> Add separate component `Palette`

I'll add the `Palette` component (next week, I think)

Do we need the `EmojiPalette` component in `uikit` then?
It seems that we can simply put a usage example of `Palette` for `EmojiPalette` in storybook
But is there a need to make a distinct `EmojiPalette` component?

korvin89:
> Do we need the `EmojiPalette` component in `uikit` then? It seems that we can simply put a usage example of `Palette` for `EmojiPalette` in storybook But is there a need to make a distinct `EmojiPalette` component?

Hmm, we probably don't need it. Example in Storybook sounds quite enough

Ruminat:
Turned the component into `Palette`

Ruminat:
@korvin89 @amje @ValeraS ping

Ruminat:
@korvin89 @amje @ValeraS ping

src/components/Palette/Palette.tsx
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    if (row.length > 0) {
+        rows.push(row);
+    }
+
+    return rows;
+}
+
+function PaletteOptionItem({
```
korvin89, line 231: Suggestions about `PaletteOptionItem` component:
- rename ~~PaletteOptionItem~~ -> `PaletteOption`
- use `Button` as wrapper for option (you're already doing it) and add property `renderOption` (like in `Select`) and use `Button.Icon` by default. In this case, we will not need the following options: `optionClassName`, `iconClassName`
- rename `checked` -> selected

src/components/Palette/Palette.tsx
```diff
@@ -0,0 +1,287 @@
...
+                : [...value, option.value];
+
+            onUpdate(newValue);
+        } else {
+            onUpdate([option.value]);
+        }
+    }, [onUpdate, option.value, value]);
+
+    return (
+        <PaletteControl
```
korvin89, line 268: I think we do not need `PaletteControl` at all. Looks like redundant abstraction

src/components/Palette/Palette.tsx
```diff
@@ -0,0 +1,287 @@
...
+            setFocusedOptionIndex(undefined);
+            event.stopPropagation();
+        },
+        [onBlur],
+    );
+
+    return (
+        <div
+            {...containerProps}
+            role={'grid'}
```
korvin89, line 171: There is no need in brackets in case of constants usage

src/components/Palette/Palette.tsx
```diff
@@ -0,0 +1,287 @@
...
+            onBlur={onBlurGrid}
+            style={style}
+            data-qa={qa}
+            tabIndex={focusedOptionIndex === undefined ? 0 : -1}
+        >
+            {rows.map((row, rowNumber) => (
+                <div
+                    className={paletteClassNames.row(rowClassName)}
+                    key={`row-${rowNumber}`}
+                    role={'row'}
```
korvin89, line 184: The same issue about constants

src/components/Palette/Palette.tsx
```diff
@@ -0,0 +1,287 @@
...
+        (event: React.FocusEvent<HTMLButtonElement>) => {
+            onBlur?.(event);
+            setFocusedOptionIndex(undefined);
+            event.stopPropagation();
+        },
+        [onBlur],
+    );
+
+    return (
+        <div
```
korvin89, line 169: You can check w3c pattern [here](https://www.w3.org/WAI/ARIA/apg/patterns/grid/) to clarify the roles of the elements

src/components/Palette/README.md
```diff
@@ -0,0 +1,219 @@
...
+`PaletteOption`:
+
+| Name     | Description             |      Type      | Default |
+| :------- | :---------------------- | :------------: | :-----: |
+| content  | HTML `class` attribute. |  `ReactNode`   |         |
+| disabled | Disables the button.    |   `boolean`    | `false` |
+| title    | HTML `title` attribute. |    `string`    |         |
+| value    | Control value.          | `PaletteValue` |         |
+
+`PaletteControlProps`:
```
korvin89, line 205: Why do we need a description of the internal interface?

korvin89 (changes requested):

src/components/Palette/Palette.tsx
Ruminat: Fixed

src/components/Palette/Palette.tsx
Ruminat: You're right
It's much simpler without it
Removed it

src/components/Palette/PaletteControl.tsx
```diff
@@ -0,0 +1,70 @@
...
+            qa,
+            checked,
+
+            onUpdate,
+            onFocus,
+            onBlur,
+        } = props;
+
+        const extraProps: React.ButtonHTMLAttributes<HTMLButtonElement> = React.useMemo(
+            () => ({value, role: 'checkbox', 'aria-checked': checked ? 'true' : 'false'}),
```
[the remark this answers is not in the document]
Ruminat: Removed the `checkbox` role and `aria-checked` attribute

src/components/Palette/Palette.tsx
```diff
@@ -0,0 +1,287 @@
...
+    | 'value'
+    | 'disabled'
+    | 'size'
+    | 'optionClassName'
+    | 'iconClassName'
+    | 'onUpdate'
+    | 'onFocus'
+    | 'onBlur'
+>) {
+    const onUpdateOption = React.useCallback(() => {
```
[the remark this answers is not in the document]
Ruminat: Done

src/components/Palette/Palette.tsx
```diff
@@ -0,0 +1,287 @@
...
+            focusedRow,
+            focusedColumn,
+            options.length,
+            focusOnOptionWithIndex,
+            columns,
+        ],
+    );
+
+    React.useEffect(() => {
+        document.addEventListener('keydown', onKeyDown);
```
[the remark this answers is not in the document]
Ruminat: Removed the global listeners and used the `useFocusWithin` hook

src/components/Palette/Palette.tsx
```diff
@@ -0,0 +1,287 @@
...
+
+export type PaletteValue = string | number;
+export type PaletteOption = {
+    value: PaletteValue;
+    content?: React.ReactNode;
+    title?: string; // HTML title
+    disabled?: boolean;
+};
+
+export interface PaletteProps
```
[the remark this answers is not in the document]
Ruminat: Added jsdoc comments to all properties

src/components/Palette/Palette.tsx
Ruminat: I think I used the right roles except for the `checkbox` for the `Button`
I also added a wrapper for the `Button` with `gridcell` role (I'm not sure if it's necessary, but it's recommended in the article you sent if I understand correctly)
The `grid` and `row` roles seem to be right

korvin89:
> @korvin89 ping

@Ruminat  Seems pretty good! Lets wait for at least another one review from @ValeraS or @amje

Ruminat:
@amje ping

korvin89 (approved):

src/components/Palette/utils.ts
```diff
@@ -0,0 +1,23 @@
+import type {PaletteOption} from './Palette';
+
+export function getPaletteRows(options: PaletteOption[], columns: number): PaletteOption[][] {
```
amje, line 3: Let's add some basic tests for this function

src/components/Palette/__stories__/Palette.stories.tsx
```diff
@@ -0,0 +1,129 @@
...
+    {content: '🤥', value: 'value-14', title: 'lying-face'},
+    {content: '🤩', value: 'value-15', title: 'star-struck'},
+    {content: '🤭', value: 'value-16', title: 'face-with-hand-over-mouth'},
+    {content: '🤮', value: 'value-17', title: 'vomiting'},
+    {content: '🥳', value: 'value-18', title: 'partying'},
+    {content: '🥴', value: 'value-19', title: 'woozy'},
+    {content: '🥶', value: 'value-20', title: 'cold-face'},
+];
+
+export const Default = () => {
```
amje, line 40: Please use CSF3 format of stories, see an example in Button's stories

src/components/Palette/Palette.tsx
```diff
@@ -0,0 +1,217 @@
...
+     */
+    onBlur?: (event: React.FocusEvent) => void;
+}
+
+interface PaletteComponent
+    extends React.ForwardRefExoticComponent<PaletteProps & React.RefAttributes<HTMLDivElement>> {}
+
+export const Palette = React.forwardRef<HTMLDivElement, PaletteProps>(function Palette(props, ref) {
+    const {
+        size = 's',
```
amje, line 92: Why not "m"? It's a default for most of our components

src/components/Palette/__stories__/Palette.stories.tsx
```diff
@@ -0,0 +1,129 @@
+import React from 'react';
+
+import * as icons from '@gravity-ui/icons';
```
amje, line 3: Let's import only a few for example, not need to import all

src/components/Palette/Palette.scss
```diff
@@ -0,0 +1,35 @@
...
+        font-size: 18px;
+    }
+    &_size_m &__option {
+        font-size: 20px;
+    }
+    &_size_l &__option {
+        font-size: 24px;
+    }
+    &_size_xl &__option {
+        font-size: 28px;
```
amje, line 33: We need shift the sizes down, they're a bit bigger than they should be. You can check font sizes of Button

src/components/Palette/Palette.tsx
```diff
@@ -0,0 +1,217 @@
...
+     */
+    rowClassName?: string;
+    /**
+     * HTML class attribute for a grid option.
+     */
+    optionClassName?: string;
+    /**
+     * HTML class attribute for the `<Button.Icon />` inside a grid option.
+     */
+    iconClassName?: string;
```
amje, line 72: `icon` name is misleading, icon is a way we center content in the Button. And why we need this class in the first place?

amje (changes requested):

amje:
In RTL mode keyboard navigation is broken

Ruminat:
> In RTL mode keyboard navigation is broken

Fixed

src/components/Palette/Palette.tsx
Ruminat: I don't think we really need it then
Removed it

src/components/Palette/Palette.scss
Ruminat: Shifted the font sizes down according to the `Button`'s `--_--icon-size`: `12px`/`16px`/`16px`/`16px`/`20px`
The font sizes of `Button` are basically the same (except for the `xl`), so I didn't use them.

src/components/Palette/__stories__/Palette.stories.tsx
Ruminat: Done

src/components/Palette/README.md
```diff
@@ -0,0 +1,240 @@
...
+| onUpdate        | Fires when the user changes the state. Provides the new value as a callback's argument. |              `(value: string[]) => void`               |         |
+| optionClassName | HTML `class` attribute for the palette button.                                          |                        `string`                        |         |
+| options         | List of options (palette elements).                                                     |                   `PaletteOption[]`                    |  `[]`   |
+| qa              | HTML `data-qa` attribute, used in tests.                                                |                        `string`                        |         |
+| rowClassName    | HTML `class` attribute for a palette row.                                               |                        `string`                        |         |
+| size            | Sets the size of the elements.                                                          |                    `s` `m` `l` `xl`                    |   `s`   |
```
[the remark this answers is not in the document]
Ruminat: Added

src/components/Palette/__stories__/Palette.stories.tsx
Ruminat: Done

src/components/Palette/utils.ts
Ruminat: Added tests

Ruminat:
@amje ping

amje (approved):

Issue #975, opened 2023-08-31
Unequal text alignment in TextInput


Kelayn:
In TextInput component label and input value appear to have different vertical alignment
![telegram-cloud-photo-size-2-5440448852554533972-y](https://github.com/gravity-ui/uikit/assets/31416046/b524b486-9c50-49d1-875e-05b23266ce26)

expected: to have same alignment

Can be detected on large screens using Yandex.Browser. Depends on page scale (won't be reproduced on 110+% )

PR #1306, opened 2024-02-02
fix(TextInput): unequal text alignment in textinput

DaniilDem:
Closes https://github.com/gravity-ui/uikit/issues/975


src/components/controls/TextInput/TextInput.scss
```diff
@@ -35,6 +35,25 @@
     @include mixins.text-accent;
 }
 
+@mixin input-control($size) {
```
korvin89, line 38: Add here some description please - why we use corrected mixin here

DaniilDem: fixed it

korvin89 (approved):

PR #1309, opened 2024-02-06
feat(Button): added prop 'form' to link a button to a form

AlexKomz:
Added prop 'form' to link a button to a form.


amje:
@AlexKomz Hi! It was intentioal that we haven't all button's attributes in top level props. Consider passing any additional props via `extraProps` prop

AlexKomz:
Yes, found this, thanks, I rushed the PR, I'll close it then! 🤲🏻

AlexKomz:
Suitable use of `extraProps`

PR #1332, opened 2024-02-12
feat: popup prevent alt axis overflow

Kiri111enz:


Kiri111enz:
photo_2024-02-12 11 07 05
photo_2024-02-12 11 07 07

korvin89:
@Kiri111enz Hey! We've discussed this PR with @amje @ValeraS and decided that it is a breaking change, because of new behaviour when popup could overlap anchor instead of take a new place according to the next placement value.

We have suggestions for your case:
- use [Sheet](https://github.com/gravity-ui/uikit/tree/main/src/components/Sheet) component for mobile application view (it's strongly recommended)
- use custom modifier with `altAxis` option

amje:
Not relevant anymore. We moved to the Floating UI

PR #1340, opened 2024-02-12
fix(Popup): pick only used props from PopperProps

ValeraS:


korvin89 (approved):

amje:
Not relevant anymore. We moved to the Floating UI

Issue #1354, opened 2024-02-15
Popup: add afterOpenChange property


korvin89:
It should be a callback that triggered when the animation ends after `Popup` is shown or hidden
```typescript
type AfterOpenChange = (open: boolean) => void;
```

amje:
@korvin89 Alternative: provide callbacks from CSSTransition component

PR #1388, opened 2024-03-01
feat(Popup): add transition callbacks

flops:
Issue: https://github.com/gravity-ui/uikit/issues/1354


src/components/Popup/Popup.tsx
```diff
@@ -20,34 +20,62 @@ import './Popup.scss';
 export type PopupPlacement = PopperPlacement;
 export type PopupAnchorRef = PopperAnchorRef;
 
+type VoidFunction = () => void;
```
amje, line 23: There is already `VoidFunction` type provided by TypeScript

amje (approved):

PR #1404, opened 2024-03-05
fix(DropdownMenu): fix opening from keyboard

### LINKED ISSUE #1373: DropdownMenu not opening from keyboard
https://github.com/gravity-ui/uikit/assets/3368839/f1c84998-42e9-48dd-be8c-e789b8234bb4

Bluepuper:
Fix #1373

### DIFF
--- a/src/hooks/useActionHandlers/useActionHandlers.ts
+++ b/src/hooks/useActionHandlers/useActionHandlers.ts
@@ -16,6 +16,7 @@ export function createOnKeyDownHandler<T>(callback?: AnyFunction) {
             callback &&
             [KeyCode.ENTER, KeyCode.SPACEBAR, KeyCode.SPACEBAR_OLD].includes(event.key)
         ) {
+            event.preventDefault();
             callback(event);
         }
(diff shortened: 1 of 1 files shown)


amje (approved):

src/hooks/useActionHandlers/useActionHandlers.ts
```diff
@@ -16,6 +16,7 @@ export function createOnKeyDownHandler<T>(callback?: AnyFunction) {
             callback &&
             [KeyCode.ENTER, KeyCode.SPACEBAR, KeyCode.SPACEBAR_OLD].includes(event.key)
         ) {
+            event.preventDefault();
```
ogonkov, line 19: Why it is not in callback in Dropdown?

amje: It's useful in every use of useActionHandlers, we want only call the callback when key is pressed, not the default behaviour of this key

ogonkov (approved):
okay

PR #1417, opened 2024-03-11
feat(TreeList): new useList family component

IsaevAlexandr:
- added new component;
- replaced list render method in TreeSelect component;
- grouped `onItemClick` callback arguments;


src/components/TreeSelect/TreeSelect.tsx
```diff
@@ -82,6 +86,7 @@ export const TreeSelect = React.forwardRef(function TreeSelect<T>(
         selectedById: selected,
     });
 
+    // used for control render. Needed a smart way to share list context here
     const listParsedState = useList({
```
IsaevAlexandr, line 90: Here is one big problem: we have to call  `useList` to get same result as inside `TreeList` component.
I will be glad to have ideas on how this can be improved. Perhaps not in this task

src/components/TreeList/TreeList.tsx
```diff
@@ -0,0 +1,140 @@
...
+        size = 'm',
+        items,
+        className,
+        expandedById,
+        disabledById,
+        activeItemId,
+        selectedById,
+        getId,
+        renderItem: propsRenderItem,
+        renderContainer = TreeListContainer,
```
amje, line 33: Suggestion: rename `renderContainer` to `containerSlot` and move `renderItem` to the `children`

src/components/TreeList/__stories__/components/WithGroupSelectionControlledStateAndCustomIcon.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+        listState.setSelected((prevState) => ({
+            [id]: !prevState[id],
+        }));
+
+        listState.setActiveItemId(id);
+    };
+
+    return (
+        <Flex direction="column" gap="3">
```
amje, line 56: Do we need Flex with one child?

src/components/TreeList/types.ts
```diff
@@ -0,0 +1,93 @@
...
+     * Define custom id depended on item data value to use in controlled state component variant
+     */
+    getId?(item: T): ListItemId;
+    /**
+     * Override list item content by you custom node.
+     */
+    renderItem?: TreeListRenderItem<T>;
+    renderContainer?: TreeListRenderContainer<T>;
+    /**
+     * If you wont to disable default behavior pass `disabled` as a value;
```
amje, line 72:
```suggestion
     * If you want to disable default behavior pass `disabled` as a value;
```

src/components/TreeSelect/TreeSelect.tsx
```diff
@@ -3,22 +3,21 @@ import React from 'react';
 import {useForkRef, useUniqId} from '../../hooks';
 import {SelectControl} from '../Select/components';
 import {SelectPopup} from '../Select/components/SelectPopup/SelectPopup';
+import {TreeList as treeList} from '../TreeList';
```
amje, line 6: Why alias to lowercase name?

amje (changes requested):

src/components/TreeList/TreeList.scss
```diff
@@ -0,0 +1,6 @@
+@use '../variables';
```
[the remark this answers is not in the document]
IsaevAlexandr: it corresponds to class name of the components. I can delete it if don't needed

src/components/TreeList/TreeList.tsx
IsaevAlexandr: these are two different possibilities. For most tasks, you don't need to redefine the container

src/components/TreeList/__stories__/components/WithGroupSelectionControlledStateAndCustomIcon.tsx
IsaevAlexandr: why not?) it's just story. I think it stayed when I tested the component with the button

src/components/useList/components/ListItemView/ListItemView.tsx
```diff
@@ -52,6 +52,10 @@ export interface ListItemViewProps extends QAProps {
     className?: string;
     role?: React.AriaRole;
     expanded?: boolean;
+    /**
+     * Add active styles and change selection behavior during dnd is performing
+     */
+    isDragging?: boolean;
```
amje, line 58: Let's call `dragging` like other states with no `is*`

amje (changes requested):

amje (approved):

amje (approved):

Raubzeug (approved):

PR #1425, opened 2024-03-15
feat(Toc): added additional data-qa attributes inside component

vrozaev:


src/components/Toc/Toc.tsx
```diff
@@ -29,6 +29,7 @@ export const Toc = React.forwardRef<HTMLElement, TocProps>(function Toc(props, r
                             content={content}
                             value={value}
                             href={href}
+                            qa={qa ? `${qa}-item` : ''}
```
amje, line 32: Should be undefined by default, empty string results in attribute presence

src/components/Toc/Toc.tsx
```diff
@@ -46,6 +47,7 @@ export const Toc = React.forwardRef<HTMLElement, TocProps>(function Toc(props, r
                                                 value={childrenValue}
                                                 href={childrenHref}
                                                 childItem={true}
+                                                qa={qa ? `${qa}-item` : ''}
```
amje, line 50: Same

src/components/Toc/TocItem/TocItem.tsx
```diff
@@ -44,5 +45,15 @@ export const TocItem = (props: TocItemProps) => {
             </a>
         );
 
-    return <div className={b('section', {child: childItem, active})}>{item}</div>;
+    let testId = '';
```
amje, line 48: Same

amje (changes requested):

PR #1435, opened 2024-03-21
feat(Breadcrumbs): export `LinkBreadcrumbsItem` & `ButtonBreadcrumbsItem` types

smsochneg:


korvin89 (approved):

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -24,12 +24,12 @@ type BaseBreadcrumbsItem = {
     title?: string;
 };
 
-type LinkBreadcrumbsItem = {
+export type LinkBreadcrumbsItem = {
```
amje, line 27: Let's rename them to `BreadcrumbsLinkItem` and `BreadcrumbsButtonItem`, that's `Breadcrumbs` name will be at the beginning

amje (changes requested):

PR #1452, opened 2024-03-25
feat(Button): refactor to flex and center icons

teleginzhenya:
Button component was refactored to use modern flex. Position absolute and complex padding calculations was removed.

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -15,6 +15,7 @@ $block: '.#{variables.$ns}button';
     --_--font-size: var(--g-text-body-1-font-size);
 
     @include mixins.button-reset();
+    position: relative;
     overflow: visible;
     box-sizing: border-box;
     height: var(--g-button-height, var(--_--height));
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -259,13 +259,20 @@ $block: '.#{variables.$ns}button';
     }
 
     &__icon {
-        display: inline-flex;
-        justify-content: center;
+        display: inline-block;
+        position: relative;
         width: var(--g-button-height, var(--_--icon-size));
         height: var(--g-button-height, var(--_--height));
         pointer-events: none;
 
+        &::after {
+            content: '\00a0';
+            visibility: hidden;
+        }
+
         &-inner {
+            position: absolute;
+            inset: 0;
             display: flex;
             align-items: center;
             justify-content: center;
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -261,7 +261,16 @@ $block: '.#{variables.$ns}button';
     &__icon {
         display: inline-block;
         position: relative;
-        width: var(--g-button-height, var(--_--icon-size));
+        margin: 0
+            calc(
+                (
+                        var(--g-button-height, var(--_--height)) - var(
+                                --g-button-icon-size,
+                                var(--_--icon-size)
+                            )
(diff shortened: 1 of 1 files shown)


src/components/Button/Button.scss
```diff
@@ -266,49 +255,27 @@ $block: '.#{variables.$ns}button';
...
-        position: relative;
-        width: var(--g-button-height, var(--_--height));
+        display: inline-flex;
+        justify-content: center;
+        width: var(--g-button-height, var(--_--icon-size));
         height: var(--g-button-height, var(--_--height));
         pointer-events: none;
 
-        &::after {
-            content: '\00a0';
```
teleginzhenya, line 280: Seems like it didn't have any effect before

teleginzhenya (commented):
And seems like we might need to add some tests/stories to the button

* custom width
* max width

src/components/Button/Button.scss
```diff
@@ -12,24 +12,9 @@ $block: '.#{variables.$ns}button';
...
-                        --g-button-icon-size,
-                        var(--_--icon-size)
-                    )
-            ) / 2
-    );
     --_--font-size: var(--g-text-body-1-font-size);
 
     @include mixins.button-reset();
-    display: inline-block;
-    position: relative;
```
amje, line 32: There are children elements with `absolute` position (:before, :after), this prop can't be omitted

src/components/Button/Button.scss
amje: Please rebase your branch, I added test with text interation.

amje (changes requested):

src/components/Button/Button.scss
teleginzhenya: fixed 79ccb7a622dc73a312669206b7b58c3ab85dc9e9

src/components/Button/Button.scss
teleginzhenya: returned and fixed 85ec4446b414efc7d2038ff8efb5307cb5914587

src/components/Button/Button.scss
```diff
@@ -266,13 +256,12 @@ $block: '.#{variables.$ns}button';
...
         display: inline-block;
         white-space: nowrap;
-        padding: 0 var(--g-button-padding, var(--_--padding));
     }
 
     &__icon {
         display: inline-block;
         position: relative;
-        width: var(--g-button-height, var(--_--height));
+        width: var(--g-button-height, var(--_--icon-size));
```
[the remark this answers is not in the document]
teleginzhenya: Fixed here b269fd5886173fdb1c40916559edcb9d09613e27 and I've added some screenshot tests for this scenario in [separate pull request](https://github.com/gravity-ui/uikit/pull/1455)

src/components/Button/Button.scss
```diff
@@ -289,26 +287,12 @@ $block: '.#{variables.$ns}button';
...
+            order: 0;
         }
 
         &_side_end {
-            inset-inline-end: var(--_--icon-position);
-
-            & ~ #{$block}__text {
-                padding-inline-end: var(--_--icon-space);
-            }
+            order: 2;
```
amje, line 295: Default value for order is `0`, let's use`-1` and `1` then?

amje (changes requested):

src/components/Button/Button.scss
teleginzhenya: 4490b68a4c8192e11597fcf3a6044d164705149a

PR #1456, opened 2024-03-27
feat(TreeSelect): rework popup border radius

IsaevAlexandr:
fixes according to review


korvin89 (approved):

korvin89:
I think it looks like `feat` changing

PR #1458, opened 2024-03-28
fix(TextInput): turn off event onBlur on input when use clear button

ananas7:


src/components/Button/Button.tsx
```diff
@@ -76,6 +76,8 @@ export interface ButtonProps extends DOMProps, QAProps {
     onClick?: React.MouseEventHandler<HTMLButtonElement | HTMLAnchorElement>;
     onMouseEnter?: React.MouseEventHandler<HTMLButtonElement | HTMLAnchorElement>;
     onMouseLeave?: React.MouseEventHandler<HTMLButtonElement | HTMLAnchorElement>;
+    onMouseDown?: React.MouseEventHandler<HTMLButtonElement | HTMLAnchorElement>;
+    onMouseUp?: React.MouseEventHandler<HTMLButtonElement | HTMLAnchorElement>;
```
korvin89, line 80: Lets use [extraProps](https://github.com/gravity-ui/uikit/blob/main/src/components/Button/Button.tsx#L73) instead

korvin89 (changes requested):

src/components/Button/Button.tsx
ananas7: a9c43223c53507efbb8678e2ab14bb8503e16c31

korvin89 (approved):

PR #1474, opened 2024-04-02
feat(Overlay): implement component

Vladeeg:


src/components/LoadingOverlay/__stories__/LoadingOverlay.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+                    </React.Fragment>
+                )}
+            </LoadingOverlay>
+            <Button className={spacing({mt: 2})} onClick={() => setShowContent((value) => !value)}>
+                Toggle content
+            </Button>
+        </div>
+    );
+};
+export const Default = DefaultTemplate.bind({});
```
IsaevAlexandr, line 39: pls add more examples in storybook.
I think overlay in dialog content and table overlay would be enough.
It is better to do an example with a table when there is no data at first and they are loaded worse to show how overlay min height work

src/components/LoadingOverlay/LoadingOverlay.tsx
```diff
@@ -0,0 +1,42 @@
...
+    loaderProps?: LoaderProps;
+    background?: OverlayBackground;
+    loading?: boolean;
+    children?: React.ReactNode;
+    className?: string;
+    contentClassName?: string;
+}
+
+export function LoadingOverlay({
+    loaderProps,
```
IsaevAlexandr, line 25: do we realy need override className in Loader component? 
I think it's not worth giving the user this opportunity to mix class name.
instead of `loaderProps` let pass `size` prop only?

src/components/LoadingOverlay/LoadingOverlay.tsx
```diff
@@ -0,0 +1,42 @@
...
+    children,
+    className,
+    contentClassName,
+    ...boxProps
+}: LoadingOverlayProps) {
+    return (
+        <Box className={b(null, className)} {...boxProps}>
+            <div className={b('overlay', {visible: loading})}>
+                <div className={b('background', {type: background})} />
+                <Loader {...loaderProps} />
```
IsaevAlexandr, line 37: how do you think, may be add ability to pass a component insted of hardcode Loader component?

```tsx
const LoadingOverlay = ({
  loaderSlot,
  size = 'm',
}: {loader: React.ReactNode}) => {
  return (
     ...
     {typeof loaderSlot === 'undefined' ? <Loader size={size} /> : loaderSlot}
     ...
  )
}
```
it is important to make it possible to pass `null` so that the display can be completely redefined.

src/components/LoadingOverlay/README.md
```diff
@@ -0,0 +1,32 @@
...
+| ...props         | `BoxProps`         |          |         | Props of a wrapping Box element             |
+
+### Basic usage
+
+```jsx
+<LoadingOverlay loading={loading} contentClassName={b('content')}>
+  <div>I am an example</div>
+  <div>content</div>
+</LoadingOverlay>
+```
```
IsaevAlexandr, line 32: need to add some info about override background by `--g-color-base-background` css custom property

and examples with loader overrides

IsaevAlexandr (changes requested):

src/components/LoadingOverlay/README.md
Vladeeg: Added examples for different loaders, as for overriding `--g-color-base-background`, as I remember, we've decided that it's not really appreciated by designers?

src/components/Overlay/Overlay.tsx
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export function Overlay({
+    className,
+    view = 'base',
+    visible: loading = false,
+    children,
+}: OverlayProps) {
+    return (
+        <div className={b({visible: loading}, className)}>
```
IsaevAlexandr, line 25: You can use `Flex` component here and remove styles with centring from .scss file

src/components/Overlay/README.md
```diff
@@ -0,0 +1,32 @@
...
+| :-------- | :----------------- | :------: | :------ | :---------------------------------- |
+| className | `String`           |          |         | CSS class name of the root element  |
+| visible   | `Boolean`          |          | `false` | Overlay visibility state            |
+| view      | `"base"` `"float"` |          | `base`  | Overlay background style            |
+| children  | `React.ReactNode`  |          |         | Content, usually a Loader component |
+
+### Basic usage
+
+```jsx
+<div style={{position: 'relative'}}>
```
IsaevAlexandr, line 26: When we discussed this component last time, we agreed to add the `position` property to the `Box` component and make an example with this application

src/components/Overlay/README.md
```diff
@@ -0,0 +1,32 @@
...
+### PropTypes
+
+| Name      | Type               | Required | Default | Description                         |
+| :-------- | :----------------- | :------: | :------ | :---------------------------------- |
+| className | `String`           |          |         | CSS class name of the root element  |
+| visible   | `Boolean`          |          | `false` | Overlay visibility state            |
+| view      | `"base"` `"float"` |          | `base`  | Overlay background style            |
+| children  | `React.ReactNode`  |          |         | Content, usually a Loader component |
+
+### Basic usage
```
IsaevAlexandr, line 23: Can use also add `.msx` docs for storybook?

src/components/Overlay/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+# Loader
+
+<!--/GITHUB_BLOCK-->
+
+```tsx
+import {Overlay} from '@gravity-ui/uikit';
+```
+
+The `Overlay` component renders an overlay over the parent element with relative position.
```
IsaevAlexandr, line 11: it should be highlighted here that the component should be used with a parent with `position: relative`

src/components/Overlay/Overlay.tsx
```diff
@@ -0,0 +1,30 @@
...
+export type OverlayBackground = 'base' | 'float';
+
+export interface OverlayProps {
+    className?: string;
+    view?: OverlayBackground;
+    visible?: boolean;
+    children?: React.ReactNode;
+}
+
+export function Overlay({
```
IsaevAlexandr, line 18: @amje @korvin89  Do we want to name this component `Overlay` or `Backdrop`?

src/components/Overlay/README.md
```diff
@@ -0,0 +1,32 @@
...
+| view      | `"base"` `"float"` |          | `base`  | Overlay background style            |
+| children  | `React.ReactNode`  |          |         | Content, usually a Loader component |
+
+### Basic usage
+
+```jsx
+<div style={{position: 'relative'}}>
+  <div>Some content to hide under overlay</div>
+  <Overlay visible={loading}>
+    <Loader />
```
IsaevAlexandr, line 29: @amje @korvin89 are you sure that this use case is really convenient? Do I transfer the same component to the loader every time? It looks like 90% of all uses here will be such a `<Loader />` component and it should be made default by default

src/components/Overlay/README.md
```diff
@@ -0,0 +1,32 @@
...
+```tsx
+import {Overlay} from '@gravity-ui/uikit';
+```
+
+The `Overlay` component renders an overlay over the parent element with relative position.
+For example, it can be used to preserve the desired layout while loading data.
+
+### PropTypes
+
+| Name      | Type               | Required | Default | Description                         |
```
amje, line 16: Please follow the structure we use in other readme files, see `Button`'s readme as an example

src/components/Overlay/__stories__/Overlay.scss
```diff
@@ -0,0 +1,31 @@
+@use '../../variables';
```
amje, line 1: Better to rename the file to `OverlayShowcase` to not confuse with the component's CSS file

src/components/Overlay/__stories__/Overlay.stories.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+import {columns, data} from './utils';
+import type {DataItem} from './utils';
+
+import './Overlay.scss';
+
+const b = block('overlay-showcase');
+
+export default {
+    title: 'Components/Overlays/Overlay',
```
amje, line 24: I would keep it in the `Utils` folder rather than in the `Overlays`

src/components/Overlay/Overlay.tsx
amje: Backdrop is a special case of Overlay use. So I think Overlay name is ok

src/components/Overlay/README.md
amje: Yes, I believe it's better to be that way, Overlay is a utility component, it shouldn't have "default" look. The look will depend on the usage context. Also, turning off the default behaviour would look a bit wierd:
```tsx
<Overlay visible children={null} />
```

src/components/Overlay/Overlay.scss
```diff
@@ -0,0 +1,45 @@
...
+
+    display: flex;
+    visibility: hidden;
+    justify-content: center;
+    align-items: center;
+
+    opacity: 0;
+
+    transition:
+        visibility 0.1s linear,
```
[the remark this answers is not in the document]
Vladeeg: Seems like it works partially, because without this transition `visibility` changes instantly, but with it it toggles only after 0.1s delay, which is intended. I'll look in other ways to implement it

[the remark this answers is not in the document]
Vladeeg: Actually, mozilla's docs state that visibility is animatable, but its interpolation handled as discrete: https://developer.mozilla.org/en-US/docs/Web/CSS/visibility#interpolation

IsaevAlexandr (approved):

src/components/Overlay/__stories__/OverlayShowcase.stories.tsx
amje, line 1: Filename should be `Overlay.stories.tsx` if it's a stories file.

src/components/Overlay/__stories__/OverlayShowcase.scss
amje, line 1: If the file consists of styles for stories then it should be name like `Overlay.stories.scss` and the prefix for classes would be `.overlay-stories`. `<Name>Showcase` name is for special story components that demonstrate multiple states at once.

src/components/Overlay/README.md
```diff
@@ -0,0 +1,46 @@
...
+```tsx
+<Overlay background="base">
+<Overlay background="float">
+```
+
+<!--/GITHUB_BLOCK-->
+
+## Properties
+
+| Name       | Type               | Required | Default | Description                         |
```
amje, line 41: The columns should be the following: Name, Description, Type, Default, in such order. Type and Default columns should be center aligned

amje (changes requested):

src/components/Overlay/Overlay.scss
```diff
@@ -0,0 +1,45 @@
...
+            }
+
+            &_float {
+                background-color: var(--g-color-base-float);
+            }
+        }
+    }
+
+    &__children {
+        z-index: 0;
```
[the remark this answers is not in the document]
Vladeeg: That's because of how opacity works in html. When z-index is not set, container would be drawn on top of its children. For children with opacity, though, it would work even without z-index.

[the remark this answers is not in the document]
amje: In order to not depend on stacking context switching due to opacity change, you should set `isolation: isolate` on the root, and both z-index on background and children elements with values: 0 and 1, respectively. So children will be always on top of background.

amje (changes requested):

amje (approved):

PR #1506, opened 2024-04-11
feat(Portal): take in consideration parent theme

ValeraS:


korvin89 (approved):

amje (approved):

src/components/layout/hooks/useCurrentActiveMediaQuery.tsx
```diff
@@ -81,33 +81,20 @@ export const useCurrentActiveMediaQuery = (
...
 
         return () => {
-            mounted = false;
-            queries.removeListeners(onChange);
+            queries.removeListeners(setState);
         };
-        // don't support runtime breakpoint redefinition. Breakpoints defined only one at LayoutTheme
-        // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, []);
+    }, [breakpointsMap]);
```
IsaevAlexandr, line 97: if user pass layout theme object like
```tsx
<ThemeProvider layout={{
  config: {
    breakpoints: {...}
  }
}} >...</ThemeProvider>
```
and `App` will have some state upper ThemeProvider it will cause this function to be constantly recalculated.

We need good reasons to be able to set breakpoints dynamically in runtime

src/components/theme/types.ts
```diff
@@ -8,4 +8,5 @@ export interface ThemeContextProps {
     theme: Theme;
     themeValue: RealTheme;
     direction: Direction;
+    scoped?: boolean;
```
IsaevAlexandr, line 11: may be some comment here?
why is this property needed?

IsaevAlexandr (changes requested):

src/components/layout/hooks/useCurrentActiveMediaQuery.tsx
ValeraS: if `config` never changes, the user can move its creation outside the component. Otherwise, he can expect the changes to be taken into account.

IsaevAlexandr (approved):

PR #1510, opened 2024-04-12
fix(Dialog): correct width's vars order

saxumcordis:
it is necessary to reverse the order of variables to give opportunity to change dialog width

### DIFF
--- a/src/components/Dialog/Dialog.scss
+++ b/src/components/Dialog/Dialog.scss
@@ -9,7 +9,7 @@ $block: '.#{variables.$ns}dialog';
     position: relative;
     display: flex;
     flex-direction: column;
-    width: var(var(--g-dialog-width), --_--width);
+    width: var(--g-dialog-width, var(--_--width));
 
     &_has-scroll {
         overflow-y: auto;
--- a/src/components/Dialog/Dialog.scss
+++ b/src/components/Dialog/Dialog.scss
@@ -9,7 +9,7 @@ $block: '.#{variables.$ns}dialog';
     position: relative;
     display: flex;
     flex-direction: column;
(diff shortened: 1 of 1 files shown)


amje (approved):

src/components/Dialog/Dialog.scss
```diff
@@ -9,7 +9,7 @@ $block: '.#{variables.$ns}dialog';
     position: relative;
     display: flex;
     flex-direction: column;
-    width: var(--_--width, var(--g-dialog-width));
```
saxumcordis, line 12: why here is double var()

its ok to use `var(--g-dialog-width), --_--width)`, isn't it?

@amje

src/components/Dialog/Dialog.scss
```diff
@@ -9,7 +9,7 @@ $block: '.#{variables.$ns}dialog';
...
+    width: var(var(--g-dialog-width), --_--width);
```
ValeraS, line 12:
```suggestion
    width: var(--g-dialog-width, var(--_--width));
```

ValeraS (changes requested):

src/components/Dialog/Dialog.scss
ValeraS: https://developer.mozilla.org/en-US/docs/Web/CSS/var#syntax

saxumcordis:
not actual

src/components/Dialog/Dialog.scss
amje: @saxumcordis No, @ValeraS is right, syntax is incorrect

ValeraS (approved):

Issue #1490, opened 2024-04-05
Bug(List): broken onLoadMore in virtualized List


DenisVershkov:
way to reproduce:

items: [at least one item]
virtualized: true
loading: true
onLoadMore: () => anything

There is no loading indicator in this configuration => onLoadMore is never triggered.

PR #1513, opened 2024-04-15
fix: add load more functionallity to virtualized list (#1490)

DenisVershkov:


src/components/List/__tests__/List.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+afterEach(() => {
+    cleanup();
+    jest.clearAllMocks();
+});
+
+describe('base List', () => {
+    it('should render loading indicator', () => {
+        setup({virtualized: false, onLoadMore: mockOnLoadMorFn, loading: true});
+
+        const loader = screen.getByRole('list').querySelector('.g-loader');
```
korvin89, line 35: Lets use QA attributes instead of class name

korvin89 (approved):

PR #1515, opened 2024-04-16
feat(List): new list item view for List component

IsaevAlexandr:


amje:

PR #1540, opened 2024-04-23
feat(Select,List): graceful replacement list item view of the List

IsaevAlexandr:


IsaevAlexandr:
@korvin89 @amje @ValeraS Hi, when will it be possible to watch pr?

src/components/theme/types.ts
```diff
@@ -8,4 +8,8 @@ export interface ThemeContextProps {
     theme: Theme;
     themeValue: RealTheme;
     direction: Direction;
+    /**
+     * Ability to define soft migration for new list item view
```
amje, line 12: Let's clarify that the property is not for public usage

```suggestion
     * DO NOT USE. Internally used for soft migration to the new list item view
```

amje (changes requested):

amje:

Issue #1529, opened 2024-04-21
Bug(Alert): the close button is flattened


lu-perfect:
![image](https://github.com/gravity-ui/uikit/assets/104302104/a965ff5a-a433-4ed2-8546-4475ee82dd55)

PR #1542, opened 2024-04-23
fix(Alert): fix close button width

IsaevAlexandr:
related to https://github.com/gravity-ui/uikit/issues/1529

### DIFF
--- a/src/components/Alert/Alert.tsx
+++ b/src/components/Alert/Alert.tsx
@@ -17,10 +17,6 @@ import {DEFAULT_ICON_SIZE, bAlert} from './constants';
 import i18n from './i18n';
 import type {AlertProps} from './types';
 
-const closeBtnExtraProps = {
-    'aria-label': i18n('label_close'),
-};
-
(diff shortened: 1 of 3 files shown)


src/components/Alert/Alert.tsx
```diff
@@ -66,10 +70,9 @@ export const Alert = (props: AlertProps) => {
                     {onClose && (
                         <Button
                             view="flat"
+                            className={bAlert('close-btn')}
                             onClick={onClose}
-                            extraProps={{
```
ValeraS, line 70: Why change this?

ValeraS (changes requested):

src/components/Alert/Alert.tsx
IsaevAlexandr: on each render created new object. You think better to leave it as it is?

ValeraS: yes, the i18n call must be inside the component. Otherwise dynamic language change will not work.

ValeraS (approved):

PR #1554, opened 2024-04-27
fix(Select): show invalid state without passing error message

dfilatov:
Documentations says that error message is optional, but actually Select doesn't have error appearance (red border) without passing error message. This commit fixes that.

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -162,11 +162,11 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function 
     });
     const errorMessageId = useUniqId();
 
+    const isErrorStateVisible = validationState === 'invalid';
     const isErrorMsgVisible =
-        validationState === 'invalid' && Boolean(errorMessage) && errorPlacement === 'outside';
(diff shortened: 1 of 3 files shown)


src/components/Select/Select.tsx
```diff
@@ -166,7 +166,8 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function
         validationState === 'invalid' && Boolean(errorMessage) && errorPlacement === 'outside';
     const isErrorIconVisible =
         validationState === 'invalid' && Boolean(errorMessage) && errorPlacement === 'inside';
-    const isErrorStateVisible = isErrorMsgVisible || isErrorIconVisible;
+    const isErrorStateVisible =
+        validationState === 'invalid' || isErrorMsgVisible || isErrorIconVisible;
```
ValeraS, line 170:
```suggestion
    const isErrorStateVisible = validationState === 'invalid';
```
And maybe use `isErrorStateVisible` in previous checks

ValeraS (changes requested):

src/components/Select/Select.tsx
dfilatov: done

ValeraS (approved):

PR #1555, opened 2024-04-27
feat(Sheet): add aria attributes

mournfulCoroner:
https://github.com/gravity-ui/uikit/issues/690


src/components/Sheet/SheetContent.tsx
```diff
@@ -140,14 +140,20 @@ class SheetContent extends React.Component<SheetContentInnerProps, SheetContentS
 
         return (
             <React.Fragment>
-                {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events */}
                 <div
                     ref={this.veilRef}
                     className={sheetBlock('veil', veilTransitionMod)}
                     onClick={isAnimating ? undefined : this.onVeilClick}
                     onTransitionEnd={this.onVeilTransitionEnd}
+                    role="presentation"
```
amje, line 148: There is no need to specify this. div has no semantics, and hasn't contain any children

amje (changes requested):

src/components/Sheet/SheetContent.tsx
mournfulCoroner: Thank you for information! 
But I don't know how to avoid eslint-disable-next-line jsx-a11y/click-events-have-key-events because this div must be clickable. Maybe I could rewrite the component as a functional and use the useLayout instead of the usual handler? I also think it would be nice to add FocusTrap to the sheet content

amje: Oh I see, then let's keep it

amje (approved):

PR #1565, opened 2024-05-03
feat: added a hook for generating a unique color

gshteynardt:


src/hooks/useGeneratorColor/README.md
```diff
@@ -0,0 +1,28 @@
...
+```tsx
+import {useGeneratorColor} from '@gravity-ui/uikit';
+```
+
+The `useGeneratorColor` a hook that generates a unique (but consistent) background color based on some unique attribute (e.g., name, id, email). The background color remains unchanged with each update.
+
+## Properties
+
+| Name      |                         Description                          |   Type    |   Default   |
+| :-------- | :----------------------------------------------------------: | :-------: | :---------: | ------ | --------- |
```
amje, line 16: Markup is broken. Also description should be left aligned

src/hooks/index.ts
```diff
@@ -2,6 +2,7 @@ export * from './useActionHandlers';
 export * from './useAsyncActionHandler';
 export * from './useBodyScrollLock';
 export * from './useControlledState';
+export * from './useGeneratorColor';
```
amje, line 5: Let's name it `useColorGenarotor`

src/hooks/useGeneratorColor/README.md
```diff
@@ -0,0 +1,28 @@
...
+| colorKeys |               If an array of colors is passed,               | string[]  |  undefined  |        |
+|           |         an index is generated from the token passed,         |           |             |
+|           | and the value from the color array at that index is returned |           |             |
+
+## Result
+
+`useGeneratorColor` returns an object with exactly two values:
+
+1. color - unique color from a token.
+2. oppositeColor - inverted color (black or white), ensuring higher text contrast compared to the current unique color, which is usually better for human perception.
```
amje, line 28: I think `textColor` would be more descriptive. So description will be:
> textColor - text color (dark or light), ensurring higher contrast on generated color

src/hooks/useGeneratorColor/__stories__/Color.tsx
```diff
@@ -0,0 +1,30 @@
...
+import type {UseGeneratorColorProps} from '../types';
+import {useGeneratorColor} from '../useGeneratorColor';
+
+type ColorProps = AvatarProps & {
+    withText: boolean;
+    mode: UseGeneratorColorProps['mode'];
+    token: UseGeneratorColorProps['token'];
+};
+
+export const Color = ({mode, theme, token, withText, ...avatarProps}: ColorProps) => {
```
amje, line 14: `Color` -> `ColoredAvatar`

src/hooks/useGeneratorColor/__stories__/GeneratorColor.scss
amje, line 1: filename -> `UseColorGenerator.scss`

src/hooks/useGeneratorColor/__stories__/UseGeneratorColor.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+const b = block('generator-color');
+
+export default {title: 'Hooks/useGeneratorColor'} as Meta;
+
+const DefaultTemplate: StoryFn = () => {
+    const [tokens, setTokens] = React.useState<string[]>([]);
+    const [mode, setMode] = React.useState<string[]>(['unsaturated']);
+    const [withText, setWithText] = React.useState(false);
+
+    const onClick = React.useCallback(() => {
```
amje, line 26: Let's pregenerate about 10 colors and display them as a table with all mods showing

src/hooks/useGeneratorColor/__stories__/UseGeneratorColor.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+import './GeneratorColor.scss';
+
+const b = block('generator-color');
+
+export default {title: 'Hooks/useGeneratorColor'} as Meta;
+
+const DefaultTemplate: StoryFn = () => {
+    const [tokens, setTokens] = React.useState<string[]>([]);
+    const [mode, setMode] = React.useState<string[]>(['unsaturated']);
+    const [withText, setWithText] = React.useState(false);
```
amje, line 24: We can use Storybook's controls for that. No need to handle state by ourselves

src/hooks/useGeneratorColor/utils/color.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    private getHash(token: string) {
+        const hash = hashFnv32a(token, 0x73_6f_6d_65) ^ hashFnv32a(token, 0x64_6f_72_61);
+
+        return hash;
+    }
+}
+
+export const colorGenerator = (args: ColorProps) => {
```
amje, line 117: That's redundant function. There is no need in simplification of class initializing: `colorGenerator(args) === new Color(args)`

src/hooks/useGeneratorColor/utils/color.ts
amje, line 1: Let's move the file one folder up: `.../useColorGenerator/color.ts`

src/hooks/useGeneratorColor/utils/constants.ts
amje, line 1: Let's move the file one folder up: `.../useColorGenerator/constants.ts`

src/hooks/useGeneratorColor/utils/getHue.ts
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable no-bitwise */
```
amje, line 1: Let's put all the utils into one file: `../useColorGenerator/utils.ts`

src/hooks/useGeneratorColor/utils/randomIndex.ts
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable no-bitwise */
```
amje, line 1: Can we write any tests for utils?

amje (changes requested):

src/hooks/useGeneratorColor/utils/randomIndex.ts
gshteynardt: yes, it's important to test two utilities: getHue and randomIndex

amje (changes requested):

src/hooks/useColorGenerator/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+# useColorGenerator
+
+<!--/GITHUB_BLOCK-->
+
+```tsx
+import {useColorGenerator} from '@gravity-ui/uikit';
+```
+
+The `useColorGenerator` a hook that generates a unique (but consistent) background color based on some unique attribute (e.g., name, id, email). The background color remains unchanged with each update.
```
amje, line 11:
```suggestion
The `useColorGenerator` hook generates a unique (but consistent) background color based on some unique attribute (e.g., name, id, email). The background color remains unchanged with each update.
```

src/hooks/useColorGenerator/types.ts
```diff
@@ -0,0 +1,29 @@
...
+    situration: [number, number];
+};
+
+export type ThemeColorSettings = {
+    light: ColorOptions;
+    medium: ColorOptions;
+    heavy: ColorOptions;
+};
+
+export type INTENSITY = 'light' | 'medium' | 'heavy';
```
amje, line 12: Types are always PascalCase
```suggestion
export type Intensity = 'light' | 'medium' | 'heavy';
```

src/hooks/useColorGenerator/useColorGenerator.ts
```diff
@@ -0,0 +1,43 @@
...
+                {...avatarProps}
+                text={text}
+                color={text ? textColor : undefined}
+                backgroundColor={color}
+            />
+        );
+    };
+```
+*/
+export function useColorGenerator(props: UseColorGeneratorProps) {
```
amje, line 32: There should also be return type of hook:
```ts
interface UseColorGeneratorResult {
  color: string;
  textColor: string;
}
```

src/hooks/useColorGenerator/useColorGenerator.ts
```diff
@@ -0,0 +1,43 @@
...
+    const theme = useThemeType();
+
+    const {color, hue, saturation, lightness} = getPersistentColor({
+        ...props,
+        theme,
+    });
+
+    const textColor = getTextColor(hue, saturation, lightness);
```
amje, line 40: Let's wrap them both in `React.useMemo`, to save computation time during rerenders

src/hooks/useColorGenerator/useColorGenerator.ts
```diff
@@ -0,0 +1,43 @@
...
+ * ensuring higher text contrast compared to the current color, which is usually better for human perception.
+ *
+ * Usage example:
+ ```tsx
+    import React from 'react';
+    import {Avatar} from '@gravity-ui/uikit';
+
+    const Component = ({ token, text, ...avatarProps }) => {
+        const {color, textColor} = useColorGenerator({
+            token,
```
amje, line 18:
```suggestion
            seed,
```

src/hooks/useColorGenerator/useColorGenerator.ts
```diff
@@ -0,0 +1,43 @@
+/* eslint-disable valid-jsdoc */
```
amje, line 1: Let's use correct jsdoc, see the [reference](https://jsdoc.app/tags-example)

src/hooks/useColorGenerator/index.ts
```diff
@@ -0,0 +1,2 @@
+export {useColorGenerator} from './useColorGenerator';
+export type {UseColorGeneratorProps} from './types';
```
amje, line 2: We want to export `UseColorGeneratorResult` as well

src/hooks/useColorGenerator/__stories__/useColorGenerator.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    return (
+        <div className={b()}>
+            <div className={b('grid')}>{items}</div>
+        </div>
+    );
+};
+
+export const View: Story = {
```
amje, line 80: Let's name the story as `Default`, it's the only story for the hook.

src/hooks/useColorGenerator/__stories__/useColorGenerator.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+            control: 'boolean',
+        },
+    },
+};
+
+export default meta;
+
+type Story = StoryObj<typeof ColoredAvatar>;
+
+const views = ['-', 'light', 'medium', 'heavy'] as const;
```
amje, line 27: I think we can remove first header row completely, so it will be three rows, one row for each intensity. Will be much cleaner

src/hooks/useColorGenerator/__stories__/ColorGenerator.scss
amje, line 1: Filename should be `useColorGenerator.stories.scss`

src/hooks/useColorGenerator/__stories__/ColorGenerator.scss
```diff
@@ -0,0 +1,28 @@
+@use '../../../components/variables.scss';
+
+$block: '.#{variables.$ns}color-generator';
```
amje, line 3:
```suggestion
$block: '.#{variables.$ns}color-generator-stories';
```

src/hooks/useColorGenerator/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Properties
+
+| Name      | Description                                                                                                                                | Type                               |   Default   |     |     |
+| :-------- | :----------------------------------------------------------------------------------------------------------------------------------------- | :--------------------------------- | :---------: | --- | --- |
+| mode      | Value to control color saturation                                                                                                          | `saturated` `unsaturated` `bright` | `saturated` |
+| token     | Unique attribute of the entity (e.g., name, id, email)                                                                                     | `string`                           |             |     |     |
+| colorKeys | If an array of colors is passed, an index is generated from the token passed, and the value from the color array at that index is returned | `string[]`                         |             |     |     |
```
amje, line 19: Docs need to be updated with the code

amje (changes requested):

src/hooks/useColorGenerator/useColorGenerator.ts
gshteynardt: similar approach to the documentation in used in the file [https://github.com/gravity-ui/uikit/blob/main/src/hooks/useFileInput/useFileInput.ts](url)

src/hooks/useColorGenerator/useColorGenerator.ts
gshteynardt: i think useMemo will actually consume more memory and computations here. 
all props (intensity and seed) will be listed in the useMemo dependencies, and the functions will always have to be recalculated when any prop changes. So useMemo will be redundant here.

there is also no point in wrapping `getTextColor` in useMemo, because if the hue, saturation or lightness color parameters changed, it means the props have changed.

To avoid unnecessary re-renders, the user should wrap their component in memo.

amje: If props of the hook not changing why we should rerender? It's a rule of thumb, to return memoized vaue from the hooks

src/hooks/useColorGenerator/useColorGenerator.ts
amje: That's also incorrect and should be fixed

src/hooks/useColorGenerator/useColorGenerator.ts
gshteynardt: why will the values be recalculated if the props haven't changed?
in my project, i wrapped the component in `memo`, which uses similar hook to prevent unnecessary re-renders.
in this case, excessive memoization in the hook will only make things worse.
but if that's the standart practice in gravity, then ok

amje: Every component's render will recalculate hook's value. Suggesting users to memoize their components instead of memoizing inside the hook is better? I don't think so

src/hooks/useColorGenerator/README.md
```diff
@@ -8,15 +8,14 @@
...
-| :-------- | :----------------------------------------------------------------------------------------------------------------------------------------- | :--------------------------------- | :---------: | --- | --- |
-| mode      | Value to control color saturation                                                                                                          | `saturated` `unsaturated` `bright` | `saturated` |
-| token     | Unique attribute of the entity (e.g., name, id, email)                                                                                     | `string`                           |             |     |     |
-| colorKeys | If an array of colors is passed, an index is generated from the token passed, and the value from the color array at that index is returned | `string[]`                         |             |     |     |
+| Name      | Description                                            | Type                     | Default |     |     |
+| :-------- | :----------------------------------------------------- | :----------------------- | :-----: | --- | --- |
```
amje, line 16: Type column should also be centered. Also markdown seems broken.

src/hooks/useColorGenerator/useColorGenerator.ts
```diff
@@ -1,43 +1,57 @@
...
-        ...props,
-        theme,
-    });
+    const {color, hue, saturation, lightness} = React.useMemo(
+        () =>
+            getPersistentColor({
+                ...props,
+                theme,
+            }),
+        [props, theme],
```
amje, line 48: `props` is an object, better to pass individual props

amje (changes requested):

PR #1577, opened 2024-05-13
fix(Button): do not shrink button when there is only one icon

teleginzhenya:
Resetting plugin makes button with only icon shrink (https://github.com/gravity-ui/uikit/issues/1529). Started to happen after flex refactoring (https://github.com/gravity-ui/uikit/pull/1452)

Related issue https://github.com/gravity-ui/uikit/issues/1529

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -34,6 +34,7 @@ $block: '.#{variables.$ns}button';
         color 0.15s linear;
     transform: scale(1);
     display: inline-flex;
+    flex-shrink: 0;
     justify-content: center;
     padding: 0 var(--g-button-padding, var(--_--padding));
     gap: var(--g-button-icon-offset, var(--_--icon-offset));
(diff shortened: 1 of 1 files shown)


korvin89:
@teleginzhenya  Could you add test for this case?

teleginzhenya:
@korvin89 sure, I did in separate pr #1579

korvin89 (approved):

src/components/Button/Button.scss
```diff
@@ -34,6 +34,7 @@ $block: '.#{variables.$ns}button';
         color 0.15s linear;
     transform: scale(1);
     display: inline-flex;
+    flex-shrink: 0;
```
amje, line 37: That's not the cause of the problem. The problem is in the negative margins of `.g-button__icon` element. Setting `flex-shrink` to 0 would break user layouts relying on flex

amje (changes requested):

src/components/Button/Button.scss
teleginzhenya: Oh, you're right
Fixed here 9047cd0f7c47b3a88bfe784fa7efa3b16fde5ac7
Test is here https://github.com/gravity-ui/uikit/pull/1579

Issue #1581, opened 2024-05-14
HTML-tag props are not inherited in Text component


chelentos:
For example I want to write something like this:

```tsx
<Text as="label" htmlFor="someId">label text</Text>
```
But i'm getting ts error about not existing prop.

<details><summary>Error example</summary>
<code>Property 'htmlFor' does not exist on type 'IntrinsicAttributes & { as?: "label" | undefined; } & Omit<TextProps, "as"> & { ref?: ((instance: HTMLLabelElement | null) => void) | RefObject<HTMLLabelElement> | null | undefined; }'</code>
</details>

korvin89:
сс @IsaevAlexandr

chelentos:
Resolved [here](https://github.com/gravity-ui/uikit/pull/1584)

baikinkirill:
Hi! We are using version uikit 6.23.0 and components 3.7.0. We still have an error when using the "as" prop in the Text component:

```
export const Text: <C="span" extends any>({ref, ...props}: (TextProps<C> & Omit<React.PropsWithoutRef<React.ComponentProps<C>>, keyof TextProps<T>> & {ref?: TextRef<C> | undefined})) => React.ReactElement
```

PR #1583, opened 2024-05-15
feat(Text, Flex, Box): typed html attributes

IsaevAlexandr:
- fully types DOM attributed depends on `as` props for `Text`, `Box` and `Flex` components with backward compatibility;
- added tests for `Text` component;

related to https://github.com/gravity-ui/uikit/issues/1581


chelentos:
I doubt the correctness of using of `Box` for `Text`. Semantically, these components are needed for different tasks. For example, does the `Text` component need props for its width/height and spacing?

IsaevAlexandr:
> I doubt the correctness of using of `Box` for `Text`. Semantically, these components are needed for different tasks. For example, does the `Text` component need props for its width/height and spacing?

these are reasonable remarks. The use of the `Box` component in this example is necessary in order to work correctly with basic props such as `qa`, `styles`, `className` and `ref` and other. Like base logic fore thouse examples. Out of the box you get the ability to use Text component something like this:
```tsx
<>
  <Text spacing={{mr: 2}}>Some text 1</Text>
  <Text>Some text 1</Text>
<>
```

chelentos:
> Out of the box you get the ability to use Text component something like this:
> 
> ```tsx
> <>
>   <Text spacing={{mr: 2}}>Some text 1</Text>
>   <Text>Some text 1</Text>
> <>
> ```

I think the question then might be why are we adding this behavior for `Text` only?

IsaevAlexandr:
> > Out of the box you get the ability to use Text component something like this:
> > ```tsx
> > <>
> >   <Text spacing={{mr: 2}}>Some text 1</Text>
> >   <Text>Some text 1</Text>
> > <>
> > ```
> 
> I think the question then might be why are we adding this behavior for `Text` only?

`Text` is one more in a row. `Flex`, `Card` and, i hope, it will be more components some what extends from `Box`

src/components/Text/__stories__/Text.stories.tsx
```diff
@@ -101,3 +103,45 @@ export const WordBreak: Story = {
         wordBreak: 'break-all',
     },
 };
+
+const LabelWithControlledFocusStory = () => {
+    const ref = React.useRef<HTMLLabelElement>(null); // don't delete this ref - needed to check currect react html type inshurance
```
korvin89, line 108: currect -> correct

amje (approved):

korvin89 (approved):

PR #1590, opened 2024-05-16
fix(Label): animation lag

itwillwork:
Before:
https://github.com/gravity-ui/uikit/assets/15855766/e9f96baf-7035-4de6-b7ba-e732b0b605c7

After:
https://github.com/gravity-ui/uikit/assets/15855766/53e794c4-f473-45d4-810b-3360cd0c92dc


amje:
@itwillwork I've refactored styles for `Label` in #1615 , it now uses native buttons instead of `Button`, no more conflicts within styles.

PR #1591, opened 2024-05-16
[WIP] feat(Toaster): add ToasterSingletonNew

korvin89:


src/components/Toaster/__stories__/Toaster.stories.tsx
```diff
@@ -93,3 +96,34 @@ export const Default: Story = {
...
+    const toasterSingletonNew2 = new ToasterSingletonNew();
+    toasterSingletonNew2.add({
+        name: `id-${index++}`,
+        theme: 'info',
+        title: 'Hey, toaster!',
+        content: <ToastContent />,
+    });
+};
+
+export const NewSingleton: Story = {
```
korvin89, line 119: I'll delete this story after review

ogonkov:
This PR illustrates problem with using `useTheme` hook with Toaster. Toaster renders in separate node, without all required context, that makes useTheme hook to fail.

Looks like this will be fixed via #1987

PR #1594, opened 2024-05-20
feat(Button): allow to pass any valid `<button>` or `<a>` HTML attributes

amje:
This PR introduces an easier way to pass any extra properties to the component and deprecates `extraProps` prop:
```jsx
// Before
<Button extraProps={{onKeyDown: () => {}}} />

// After
<Button onKeyDown={() => {}} />
```

By default, extra properties extend `React.ButtonHTMLAttributes` and when `href` property is passed they extend `React.AnchorHTMLAttributes`.

PR potentially will break types because `ButtonProps` can no longer be extended in interface declaration:

```ts
// Can't extend in interface
interface MyButton extends ButtonProps {
    myProp: string;
}

// This works fine
type MyButton = ButtonProps & {
    myProp: string;
}
```

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import type {DOMProps, QAProps} from '../types';
+import type {QAProps} from '../types';
 import {block} from '../utils/cn';
 import {isIcon} from '../utils/common';
 import {eventBroker} from '../utils/event-broker';
@@ -53,7 +53,7 @@ export type ButtonPin =
 
 export type ButtonWidth = 'auto' | 'max';
 
-export interface ButtonProps extends DOMProps, QAProps {
+interface ButtonBaseProps extends QAProps {
     /** Button appearance */
     view?: ButtonView;
     size?: ButtonSize;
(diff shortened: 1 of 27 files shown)


Raubzeug (approved):

src/components/Button/Button.tsx
```diff
@@ -125,64 +122,81 @@ const ButtonWithHandlers = React.forwardRef<HTMLElement, ButtonProps>(function B
                     view,
                 },
             });
+
+            if (onClickCapture) {
+                onClickCapture(event as any);
```
ogonkov, line 127: Why is type cast used?

src/components/Button/__tests__/Button.test.tsx
```diff
@@ -105,11 +105,11 @@ describe('Button', () => {
     test('should render custom component', () => {
         const text = 'Button with custom component';
 
-        const ButtonComponent = (props: ButtonProps) => {
+        const ButtonComponent = (props: React.HTMLAttributes<HTMLElement>) => {
             return (
-                <button {...props} style={{boxShadow: '2px 2px 2px 2px deepskyblue'}}>
```
ogonkov, line 110: Why element type changed?

ogonkov (approved):

src/components/Button/__tests__/Button.test.tsx
amje: To ensure any root tag is valid

src/components/Button/Button.tsx
amje: It requires weird type, intersection of `HTMLButtonElement` and `HTMLAnchorElement`

goshander:
@amje `aria-label` props pass to React component as is, without camelCase converting?

amje:
> @amje `aria-label` props pass to React component as is, without camelCase converting?

Yes, it's valid

goshander (approved):

amje:
Closed in favor of #2015

PR #1604, opened 2024-05-22
fix(Select): id property

GermanVor:


src/components/Select/Select.tsx
```diff
@@ -316,6 +316,7 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function
             className={selectBlock(mods, className)}
             {...focusWithinProps}
             style={inlineStyles}
+            id={selectId}
```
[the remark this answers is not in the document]
GermanVor: https://github.com/gravity-ui/uikit/pull/1604/commits/d8a23400f257ed3ac9135da443401517ae5ca9d8

src/components/Select/Select.tsx
```diff
@@ -283,6 +282,11 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function
         return null;
     };
 
+    const uniqId = useUniqId();
+    const controlId = id ?? uniqId;
+    const selectId = `select-${controlId}`;
```
korvin89, line 287: Do we really need both ids `selectId` and `controlId`?

GermanVor: Honestly, I do not know.

Seems like `SelectControlProps.selectId` was about only `aria-activedescendant` not `id` attribute. 
Also it is required property, while native `id` attribute is optional.

src/components/TreeSelect/TreeSelect.tsx
```diff
@@ -231,7 +233,7 @@ export const TreeSelect = React.forwardRef(function TreeSelect<T>(
                 handleClose={handleClose}
                 disablePortal={popupDisablePortal}
                 mobile={mobile}
-                id={`tree-select-popup-${treeSelectId}`}
+                id={popupId}
```
korvin89, line 236: We don't use this property, it could be deleted

korvin89 (approved):

PR #1607, opened 2024-05-22
refactor: enable `id-length` eslint rule

itwillwork:


korvin89 (changes requested):
It does not looks convenient for code supporting. And there are a lot of new eslint-ignores.

PR #1614, opened 2024-05-23
test(Slider): CSF 3 & visual testing

itwillwork:


Arucard89:
I suggest grouping single and range components in one page for each case. This will help reduce the number of menu items in the showcase section. I think it will be more convenient for watching.

amje:
> I suggest grouping single and range components in one page for each case. This will help reduce the number of menu items in the showcase section. I think it will be more convenient for watching.

Agreed, can we write the stories so we have both single and range slider, but with shared controls.`value` control will be the value for the single slider and `rangeValue` control will be the value for the ranged slider?

playwright/core/types.ts
```diff
@@ -13,7 +15,7 @@ import type {
 interface ComponentFixtures {
     mount<HooksConfig extends JsonObject>(
         component: JSX.Element,
-        options?: MountOptions<HooksConfig>,
+        options?: MountOptions<HooksConfig> & {wrapDivStyles?: React.CSSProperties},
```
amje, line 18:
```suggestion
        options?: MountOptions<HooksConfig> & {wrapperStyle?: React.CSSProperties},
```

amje (changes requested):

amje:
> > > I suggest grouping single and range components in one page for each case. This will help reduce the number of menu items in the showcase section. I think it will be more convenient for watching.
> > 
> > 
> > Agreed, can we write the stories so we have both single and range slider, but with shared controls.`value` control will be the value for the single slider and `rangeValue` control will be the value for the ranged slider?
> 
> @amje, Maybe we could omit the value property altogether? I don't think it is really necessary for the showcases(excluding the default showcase).
> 
> UPD: So my point is to have: -single page for slider(with value prop) -single page for range slider(with value prop) -single and range components in one page for each other special case(without value prop)

@Arucard89  Like this?
`Default` (single default with value)
`Range` (range default with value)
`Size (both single and range without value)`
`Disabled (both single and range without value)`
`// etc`

Arucard89:
> @Arucard89 Like this?
> `Default` (single default with value)
> `Range` (range default with value)
> `Size (both single and range without value)`
> `Disabled (both single and range without value)`
> `// etc`

@amje, yes. I believe that it would be helpful for users to play around with the components using the default settings and explore the special, predefined cases.

PR #1630, opened 2024-06-04
fix(Button): normal-contrast loading view

GermanVor:
there is no loading effect of normal-contrast button

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -228,8 +228,12 @@ $block: '.#{variables.$ns}button';
         &_normal-contrast {
             --_--text-color: var(--g-color-text-dark-primary);
             --_--background-color: var(--g-color-base-light);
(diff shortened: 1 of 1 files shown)


src/components/Button/Button.scss
```diff
@@ -228,8 +228,12 @@ $block: '.#{variables.$ns}button';
         &_normal-contrast {
             --_--text-color: var(--g-color-text-dark-primary);
             --_--background-color: var(--g-color-base-light);
-            --_--background-color-hover: var(--g-color-base-light-hover);
             --_--focus-outline-color: var(--g-color-line-light);
+
+            --_--background-color-hover: var(--g-color-base-light-hover);
```
amje, line 233: Keep this on current line, please. This helps to not to lose it

amje (changes requested):

amje (approved):

Issue #1587, opened 2024-05-16
Remove scrolbars custom styling


amje:
UIKit shouldn't force any styling on scrollbars. We'll remove default styling, but keep styles in mixin for futher usages

PR #1633, opened 2024-06-08
fix(theme): bring back root classname helper and fix docs

amje:


korvin89 (approved):

src/components/theme/getRootClassName.ts
```diff
@@ -0,0 +1,10 @@
+import {block} from '../utils/cn';
+import type {CnMods} from '../utils/cn';
+
+import {ROOT_CLASSNAME} from './constants';
+
+const b = block(ROOT_CLASSNAME);
+
+export function getRootClassName(mods: CnMods, className: string) {
```
ValeraS, line 8: It seems the first parameter is too general, maybe let's make it more strict?

We are not using this method yourself, isn't it?

amje: You mean instead of object allow only known keys, like `theme`?

ValeraS: Yes

src/components/theme/getRootClassName.ts
```diff
@@ -0,0 +1,13 @@
+import {block} from '../utils/cn';
+
+import {ROOT_CLASSNAME} from './constants';
+
+const b = block(ROOT_CLASSNAME);
+
+interface RootMods {
+    theme?: string;
```
ValeraS, line 8:
```suggestion
    theme?: RealTheme;
```

src/components/theme/getRootClassName.ts
```diff
@@ -0,0 +1,13 @@
+import {block} from '../utils/cn';
+
+import {ROOT_CLASSNAME} from './constants';
+
+const b = block(ROOT_CLASSNAME);
+
+interface RootMods {
```
ValeraS, line 7:
```suggestion
interface RootMods {
    nativeScrollbar?: boolean;
```

src/components/theme/getRootClassName.ts
```diff
@@ -0,0 +1,13 @@
...
+    theme?: string;
+}
+
+export function getRootClassName({theme}: RootMods = {}, className?: string) {
+    return b({theme}, className);
```
ValeraS, line 12:
```suggestion
    return b({theme, 'native-scrollbar': nativeScrollbar}, className);
```

ValeraS (changes requested):

src/components/theme/getRootClassName.ts
amje: We're planning to remove scroll customization, #1587

PR #1637, opened 2024-06-10
fix(Slider): export  function for preparing initial value

Arucard89:
-small refactor of initial value function
-add prepareSliderInnerState function to export

### DIFF
--- a/src/components/Slider/utils.ts
+++ b/src/components/Slider/utils.ts
@@ -134,24 +134,26 @@ export function getInnerState({
         state.marks = createMarks(calculateInfoPoints({count: marksCount, max, min}));
     }
 
-    if (value && Array.isArray(value)) {
-        state.range = true;
-        state.value = prepareArrayValue({min: state.min, max: state.max, value});
-    } else if (defaultValue && Array.isArray(defaultValue)) {
-        state.range = true;
-        state.defaultValue = prepareArrayValue({
-            min: state.min,
-            max: state.max,
-            value: defaultValue,
-        });
-    } else if (value) {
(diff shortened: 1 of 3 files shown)


src/components/Slider/utils.ts
```diff
@@ -88,7 +88,7 @@ function createMarks(points: number[]): RcSliderProps['marks'] {
     return marks;
 }
 
-export function getInnerState({
+export function prepareSliderInnerState({
```
korvin89, line 91: Could you provide some jsdoc here?

amje, line 91: Why user might needed this function?

amje (changes requested):

src/components/Slider/utils.ts
Arucard89: > Why user might needed this function?

I need it to use in RangeSliderPicker component to get inner state. And I don't want to duplicate the code. 
Also it could be helpful for other users in the development of components based on the slider.

Arucard89: @amje, I don't know is it really "fix" but not "chore"?
Is it important enough to get into the changelog?

Arucard89: Yes, forgot to do it, thanks
https://github.com/gravity-ui/uikit/pull/1637/commits/143584fa5100fdf50b41a55846ab5e66581c9aec

amje: Yes, it's important to note this change in the changelog because you have changed the implementation

Arucard89: It's just refactoring to make code more readable without any changes in implemenation

amje (approved):

PR #1665, opened 2024-06-19
fix(select): connect combobox to listbox

ValeraS:


IsaevAlexandr:
Hi, what is the motivation of that changes?
I'm asking because we have the same logic in `TreeSelect` component

ValeraS:
The `combobox` element should set [aria-controls](https://w3c.github.io/aria/#aria-controls) to a value that refers to the element that serves as the popup. The popup is an element that has role [listbox](https://w3c.github.io/aria/#listbox), [tree](https://w3c.github.io/aria/#tree), [grid](https://w3c.github.io/aria/#grid), or [dialog](https://w3c.github.io/aria/#dialog). In case of this component it should be a `listbox`.

korvin89 (approved):

PR #1670, opened 2024-06-19
feat(Slider)!: support form

ValeraS:


src/components/Slider/BaseSlider/BaseSlider.tsx
```diff
@@ -13,7 +13,7 @@ import './BaseSlider.scss';
 const b = block('base-slider');
 
 type BaseSliderProps = {stateModifiers: StateModifiers} & Omit<
-    SliderProps,
+    SliderProps<number | [number, number]>,
```
Arucard89, line 16: Why is this done here? this leads to the need to specify the @ts-expect-error directive at line 25

src/components/Slider/Slider.tsx
```diff
@@ -2,14 +2,15 @@
 
 import React from 'react';
 
-import debounce from 'lodash/debounce';
-
+import {useControlledState} from '../../hooks';
```
Arucard89, line 5: is it necessary here? BaseSlider component has out-of-the-box props for controlled and uncontrolled behaviour without this hook

src/components/Slider/Slider.tsx
```diff
@@ -103,41 +108,59 @@ export const Slider = React.forwardRef(function Slider(
...
+                            <React.Fragment>
+                                {handle}
+                                <input
+                                    ref={inputRef}
+                                    type="hidden"
+                                    name={name}
+                                    form={form}
+                                    value={handleProps.value}
+                                    disabled={disabled}
+                                />
```
Arucard89, line 133: is it correct to put hidden input inside the handler? 
may be it will be better to move it outside the component and connect ti Slider thru Slider's props (onChange, value, etc)?

src/components/Slider/types.ts
```diff
@@ -8,7 +8,7 @@ export type SliderValue = number | [number, number];
 
 export type RcSliderValueType = number | number[];
 
-export type SliderProps<ValueType = number | [number, number]> = {
+export interface SliderProps<ValueType = number | [number, number]> extends DOMProps, QAProps {
```
Arucard89, line 11: why did you change type to interface?

Arucard89:
@ValeraS hi. 

I see deep refactoring of basic component which adds breaking change and very specific case of usage .

May be you consider to make HOC (withForm) or new FormSlider component based on Slider component? in this case, you could add changes you need and move away from deep refactoring of the Slider and avoid breaking changes.

@korvin89 , @amje , what do you think about it?

korvin89:
@Arucard89 
> I see deep refactoring of basic component which adds breaking change

I think it is okay for the major version I suppose

> May be you consider to make HOC (withForm) or new FormSlider wrapper component based on Slider component? 

We are trying to add this functionality to all uikit controls into the core code

Arucard89:
> @Arucard89
> 
> > I see deep refactoring of basic component which adds breaking change
> 
> I think it is okay for the major version
> 
> > May be you consider to make HOC (withForm) or new FormSlider wrapper component based on Slider component?
> 
> We are trying to add this functionality to all uikit controls into the core code

oh, I see. There was no description, so I didn't know that it is planned work

src/components/Slider/Slider.tsx
Arucard89: > is it correct to put hidden input inside the handler? may be it will be better to move it outside the component and connect ti Slider thru Slider's state?

Also it is necessary to take into account the probability of two values appearing [number, number]. At my point it will be easier to manage such case.

For example like this:

```ts
...
const inputRef = useFormResetHandler({
    initialValue: Array.isArray(innerValue) ? innerValue[0] : innerValue,
    onReset: handleReset,
});
const inputSecondRef = useFormResetHandler({
    initialValue: Array.isArray(innerValue) ? innerValue[1] : innerValue,
    onReset: handleReset,
});
...
return (
        <div
            {...filterDOMProps(otherProps)}
            id={id}
            className={b(null, className)}
            ref={ref}
            style={style}
            data-qa={qa}
        >
            <input
                ref={inputRef}
                ... //
            />
            {Array.isArray(innerValue)? <input 
                ref={inputSecondRef}
                ... 
                /> : null}
            <div className={b('top', {size, hasTooltip})}></div>
            <BaseSlider
        ... />
        ...

```

src/components/Slider/Slider.tsx
```diff
@@ -30,46 +31,26 @@ export const Slider = React.forwardRef(function Slider(
...
         tabIndex,
         className,
+        style,
         qa,
         apiRef,
         'aria-label': ariaLabelForHandle,
         'aria-labelledby': ariaLabelledByForHandle,
+        name,
+        form,
+        id,
```
Arucard89, line 48: Are these props(`name`, `id`) fit to case with range slider, when value is [number, number]?

src/components/Slider/Slider.tsx
ValeraS: Normally, a slider component uses input with a type range under the hood, but not in this library. IMHO, integration with forms should be a base functionality for the slider component.

src/components/Slider/Slider.tsx
ValeraS: Yes, because we control state of the BaseSlider component.

src/components/Slider/Slider.tsx
ValeraS: Yes. `id` goes to container. And several inputs with the same name are valid (see tests).

src/components/Slider/Slider.tsx
Arucard89: @ValeraS , yes I agreed with you. And I suggest to put input near the BaseSlider instead of inside handle rendering function. I think it will be ore logical. 
Input still will be inside the Slider component. Because (in my opinion) input looks strange inside render prop.

src/components/Slider/BaseSlider/BaseSlider.tsx
Arucard89: Sorry for misunderstanding) 
I'd like to know why do we need this changes in types at all?)

src/components/Slider/Slider.tsx
ValeraS: There is should be an input for each handle, so it's convenient to render the input alongside the handle.

Arucard89: ok, thanks)

src/components/Slider/BaseSlider/BaseSlider.tsx
ValeraS: Changes give more correct types for `value` and `onUpdate`. Before this, type of value always `number | number[]` for `BaseSlider` and `number | [number | number]` for `Slider`. After this, if `value` has type 'number' then `onUpdate` also receives only values of type `number`. The same thing applies to the `[number, number]` type.

Arucard89: ok, I see. thanks)

amje (approved):

src/components/Slider/Slider.tsx
```diff
@@ -95,47 +69,71 @@ export const Slider = React.forwardRef(function Slider(
...
+    const inputRef = useFormResetHandler({initialValue: innerValue, onReset: handleReset});
+
+    const stateModifiers: StateModifiers = {
+        size,
+        error: validationState === 'invalid' && !disabled,
+        disabled,
+        'tooltip-display': innerState.tooltipDisplay,
+        rtl: direction === 'rtl',
+        'no-marks': Array.isArray(marks) ? marks.length === 0 : marks === 0,
+    };
```
Arucard89, line 95: why did you remove memoization? 
it can help prevent creating object on each render and could prevent rerenders inside components

src/components/Slider/Slider.tsx
```diff
@@ -95,47 +69,71 @@ export const Slider = React.forwardRef(function Slider(
...
+
+    const handleRender = (
```
Arucard89, line 97: why did you remove memoization? 
It could prevent some rerenders

src/components/Slider/__tetsts__/Slider.form.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+                </form>
+            );
+        }
+
+        render(<Test />);
+        const form = screen.getByTestId('form');
+        expect(form).toHaveFormValues({slider: ['5', '10']});
+
+        const sliderHandle = screen.getAllByRole('slider')[1];
+        fireEvent.keyDown(sliderHandle, {key: 'ArrowRight', keyCode: 39, code: 'ArrowRight'});
```
ogonkov, line 114: May be use userEvent.keyboard? Could be shorter

ValeraS: `userEvent.keyboard` does not set `keyCode` for the event, but `rs-slider` checks it.

amje:
@ValeraS @Arucard89 Going to merge this PR as we needed this feature. Please continute your discussion about memoization later if you want to.

PR #1682, opened 2024-06-28
feat(Card): use common constants for card's radiuses

Arucard89:
- add new radius constant --g-border-radius-xxl
- replace string values to common radius constants in styles
source:
https://github.com/gravity-ui/landing/pull/220#discussion_r1657214751

### DIFF
--- a/src/components/Card/Card.scss
+++ b/src/components/Card/Card.scss
@@ -157,11 +157,11 @@ $block: '.#{variables.$ns}card';
 
     &_size {
         &_m {
-            --_--border-radius: var(--g-border-radius-l);
+            --_--border-radius: var(--g-border-radius-card-m);
         }
 
         &_l {
-            --_--border-radius: var(--g-border-radius-xxl);
+            --_--border-radius: var(--g-border-radius-card-l);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/components/Card/Card.scss
```diff
@@ -157,11 +157,11 @@ $block: '.#{variables.$ns}card';
 
     &_size {
         &_m {
-            --_--border-radius: 8px;
+            --_--border-radius: var(--g-border-radius-l);
```
korvin89, line 160: Why we use `--g-border-radius-l` for `size_m`?

Arucard89: Card's border sizes don't have straight accordance to radius constants. I have choosen constants by values.
https://github.com/gravity-ui/uikit/blob/main/styles/themes/common/_index.scss#L13

korvin89: @amje I think it is a not good idea to use symantic variables in such cases

Arucard89: What do you think about adding variables for cards? 
For example:
—g-border-radius-card-m
—g-border-radius-card-l

Arucard89: > What do you think about adding variables for cards to _index.scss?
> For example:
> —g-border-radius-card-m
> —g-border-radius-card-l

@korvin89 , @amje, I mean like this: https://github.com/gravity-ui/uikit/pull/1682/commits/a50b4604eade6bc6ab47d30fe7d86ceadc100ed6

In my opinion, these variables also helps to make themisation easier, without using CSS API

PR #1685, opened 2024-07-01
chore<Table>: Header layout

GermanVor:
it is now
image

in this pr
image

and in our project (I do not know why the difference so clear, we do not have css injections in this place)
image


src/components/Table/hoc/withTableSorting/SortIndicator/SortIndicator.scss
```diff
@@ -3,6 +3,8 @@
 $block: '.#{variables.$ns}sort-indicator';
 
 #{$block} {
+    display: flex;
+
     &__icon {
         vertical-align: -2px;
```
Raubzeug, line 9: It seems that for now this style is redundant.

Raubzeug (changes requested):

src/components/Table/hoc/withTableSorting/SortIndicator/SortIndicator.scss
GermanVor: 127920ef77c007cbe38e71805229c977c98fb006

Raubzeug (approved):

PR #1690, opened 2024-07-01
fix(TextArea): fix content width & height styles

GermanVor:
there is no way how to custom height and width of TextArea (not through rows count)


GermanVor:
may be it have sense add new property - `maxheight` for cases `height: 100%`

korvin89:
@GermanVor Hey! Could you remove all the tests? We will add them a little bit later

korvin89 (approved):

Issue #1696, opened 2024-07-08
TreeList generate invalid HTML


ogonkov:
TreeList component put `div`, inside `ul`

Div should be li according to spec https://html.spec.whatwg.org/multipage/grouping-content.html#the-ul-element

Also, roles are different from ARIA APG https://www.w3.org/WAI/ARIA/apg/patterns/treeview/

korvin89:
cc @IsaevAlexandr

PR #1698, opened 2024-07-09
fix(ListItemView): div -> li default list item html tag

IsaevAlexandr:
related to https://github.com/gravity-ui/uikit/issues/1696

Second attempt to return valid html tag to list items


korvin89 (approved):

src/components/useList/__stories__/components/ListWithDnd.tsx
```diff
@@ -98,6 +98,7 @@ export const ListWithDnd = ({size, itemsCount, 'aria-label': ariaLabel}: ListWit
                                                     dragging={snapshot.isDragging}
                                                     ref={provided.innerRef}
                                                     endSlot={<Icon data={Grip} size={16} />}
+                                                    role="option"
```
ogonkov, line 101: Roles different from recommended practices, see link in issue

PR #1714, opened 2024-07-18
chore(TextArea): rows logic

GermanVor:
It will be very useful to have y scrollbar depending not only on `rows` or `maxRows` values, but on `height` of the `TextArea`.

now it is possible to pass `height` value via `TextAreaProps.className` to see y-scrollbar


GermanVor:
may be [it](https://github.com/gravity-ui/uikit/blob/75446dd1fb665527b9d32d7ba30283e565f47a35/src/components/controls/mixins.scss#L5) is extra now.

I did not check

src/components/controls/TextArea/TextAreaControl.tsx
```diff
@@ -54,46 +40,26 @@ export function TextAreaControl(props: Props) {
...
-                height: control.scrollHeight,
-                paddingTop,
-                paddingBottom,
-                lineHeight,
-            });
+    const realRows = React.useMemo(() => {
+        if (rows) return rows;
 
-            control.style.height = 'auto';
+        let currentRowsNumber = (innerValue?.match(/\n/g) || []).length + 1;
```
korvin89, line 49: It does not work in case of long strings without `\n`

korvin89 (changes requested):

src/components/controls/TextArea/TextAreaControl.tsx
GermanVor: you are right, better calculate rows number depend on height

[d370a17](https://github.com/gravity-ui/uikit/pull/1714/commits/d370a17e942669816fa5c67ece2b634d1e502efa)

amje:
@GermanVor Is this PR still relevant?

PR #1726, opened 2024-07-30
feat(ListItemView): ability to pass custom react node as a content prop

IsaevAlexandr:
…ent prop

What has been done:

- Added the ability to transfer a custom react component as a `content` prop in `ListItemView`;
- Removed `Flex` component from root of `ListItemView` component. Flex uses inline styles what prevent class override approach of `ListItemView`;

Now ListItemView look very close to previous approach of current `List` component

### DIFF
--- a/src/components/useList/components/ListItemView/ListItemView.tsx
+++ b/src/components/useList/components/ListItemView/ListItemView.tsx
@@ -1,96 +1,59 @@
 import React from 'react';
 
-import {Check, ChevronDown, ChevronUp} from '@gravity-ui/icons';
-
-import {Icon} from '../../../Icon';
-import {Text, colorText} from '../../../Text';
-import {Flex, spacing} from '../../../layout';
-import type {FlexProps} from '../../../layout';
+import {spacing} from '../../../layout';
 import type {QAProps} from '../../../types';
-import {block} from '../../../utils/cn';
 import {LIST_ITEM_DATA_ATR, modToHeight} from '../../constants';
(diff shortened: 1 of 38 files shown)


src/components/useList/components/ListItemView/ListItemView.tsx
```diff
@@ -102,104 +65,66 @@ export const ListItemView = React.forwardRef(function ListItemView<
...
     const onClick = disabled ? undefined : _onClick;
     const activeOnHover =
         typeof propsActiveOnHover === 'boolean' ? propsActiveOnHover : Boolean(onClick);
+    const style = {
+        minHeight: `var(--g-list-item-height, ${
+            height ??
+            modToHeight[size][
+                Number(Boolean(isListItemContentPropsGuard(content) ? content?.subtitle : false))
+            ]
+        }px)`,
```
IsaevAlexandr, line 91: Here may be too match) but it's works and do exactly what i wont

src/components/useList/components/ListItemView/ListItemView.tsx
```diff
@@ -102,104 +65,66 @@ export const ListItemView = React.forwardRef(function ListItemView<
...
-                        </Text>
-                    ) : (
-                        subtitle
-                    )}
-                </div>
-            </Flex>
-
-            {endSlot}
-        </Flex>
+            {isListItemContentPropsGuard(content) ? (
```
IsaevAlexandr, line 117: here entry point in all changes of this pr

amje (approved):

korvin89 (approved):

PR #1733, opened 2024-08-02
feat(layout): added more spacing sizes

IsaevAlexandr:


styles/themes/common/spacing.scss
```diff
@@ -3,7 +3,7 @@
 
     --g-spacing-0: calc(var(--g-spacing-base) * 0);
     --g-spacing-half: calc(var(--g-spacing-base) * 0.5);
-    --g-spacing-1: var(--g-spacing-base);
+    --g-spacing-1: var(--g-spacing-base * 1);
```
IsaevAlexandr, line 6: @amje @korvin89 
pure magic - somehow this change provide screenshot test to broke))

funny is that in AvatarStack components  does not use this size (`--g-spacing-1`)

IsaevAlexandr: I think the problem is that the storybrooke examples don't use a Theme provider.

amje:
Why we need that? We don't have such sizes

Fatal1stZ:
ping

PR #1735, opened 2024-08-06
feat(ClipboardButton): delay tooltip's closing animation after copying

### LINKED ISSUE #1739: Delay clipboard button's closing animation
Internal ticket UXRFC-400

After some research and attempt to implement it with what we have, I think it needs a bit more control over ActionTooltip to implement it. Anyone got any ideas? So, it could be a new internal component that looks like ActionTooltip or existing one but has open/close methods.

lxndr:
Closes #1739


Raubzeug:
@lxndr Hi! Let's try not to add new props and make the case with those that are already defined in the component's api?
For example, it seems that the problem with the rapidly hiding `Copied` title can be solved by adding the prop to the ActionTooltip `closeDelay={status=== 'success' ? 1000 : 0}`.
And to force hide the tooltip, you can pass the `disabled={true}` prop in the ActionTooltip.

lxndr:
@Raubzeug Hi! That's what I tried here https://github.com/gravity-ui/uikit/pull/1735/commits/626aa919e5aa2a4d40060906d62b5477a662134d. Looked OK but not quite what the design envisioned. So, I asked for a greater control over the tooltip - `disabled` can only hide. I agree that it's better off with existing props. Refer to UXRFC-400 for more info.

Raubzeug:
@lxndr as for me, thats much better! Lets stick to this implementation and bring it to the final?

src/components/ClipboardButton/ClipboardButton.tsx
```diff
@@ -81,10 +100,69 @@ const ClipboardButtonComponent = (props: ClipboardButtonComponentProps) => {
...
+            window.clearTimeout(timerIdRef.current);
+
+            timerIdRef.current = window.setTimeout(() => {
+                setShowTooltip(false);
+            }, timeout - TOOLTIP_ANIMATION);
+        },
+        [onCopy, timeout],
+    );
+
+    const resetTooltip = React.useCallback(() => {
```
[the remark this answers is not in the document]
lxndr: Sorry, I don't get it. I believe there's not much difference between enter or leave here. onMouseEnter worked better for me while I was fixing bugs that's why I left it.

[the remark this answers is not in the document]
Raubzeug: Sorry, I double checked and behaviour is as expected!

src/components/ClipboardButton/ClipboardButton.tsx
```diff
@@ -81,10 +98,73 @@ const ClipboardButtonComponent = (props: ClipboardButtonComponentProps) => {
...
+        options,
+        hasTooltip = true,
+        onMouseEnter,
+        onFocus,
+        ...buttonProps
+    } = props;
+
+    const timerIdRef = React.useRef<number>();
+    const [tooltipCloseDelay, setTooltipCloseDelay] = React.useState<number | undefined>(undefined);
+    const [tooltipDisabled, setTooltipDisabled] = React.useState<boolean>(false);
```
Raubzeug, line 114: no need in generic here

src/components/ClipboardButton/ClipboardButton.tsx
```diff
@@ -81,10 +98,73 @@ const ClipboardButtonComponent = (props: ClipboardButtonComponentProps) => {
...
+
+    React.useEffect(() => window.clearTimeout(timerIdRef.current), []);
+
+    const handleCopy: OnCopyHandler = React.useCallback(
+        (...args) => {
```
Raubzeug, line 119: no need in destructuring here

src/components/ClipboardButton/ClipboardButton.tsx
```diff
@@ -81,10 +98,73 @@ const ClipboardButtonComponent = (props: ClipboardButtonComponentProps) => {
...
+            onCopy?.(...args);
+            setTooltipDisabled(false);
```
Raubzeug, line 121: do we really need to set it on copy?

src/components/ClipboardButton/ClipboardButton.tsx
```diff
@@ -81,10 +98,73 @@ const ClipboardButtonComponent = (props: ClipboardButtonComponentProps) => {
...
+            }, timeout - TOOLTIP_ANIMATION);
+        },
+        [onCopy, timeout],
+    );
+
+    const resetTooltip = React.useCallback(() => {
+        if (tooltipDisabled) {
+            setTooltipDisabled(false);
+            setTooltipCloseDelay(undefined);
+        }
```
Raubzeug, line 137:
```suggestion
        if (tooltipDisabled) {
            setTooltipDisabled(false);
        }
        if (tooltipCloseDelay) {
            setTooltipCloseDelay(undefined);
        }
```

Raubzeug (changes requested):

src/components/ClipboardButton/ClipboardButton.tsx
lxndr: yes

src/components/ClipboardButton/ClipboardButton.tsx
lxndr: why?

Raubzeug: > why?

generally tooltipCloseDelay doesn't depends on tooltipDisabled? Why should we reset it if it's already undefined?

lxndr: it said `if (!showTooltip) {` before, so we only reset if tooltip is made hidden.

src/components/ClipboardButton/ClipboardButton.tsx
lxndr: done

src/components/ClipboardButton/ClipboardButton.tsx
lxndr: done

src/components/ClipboardButton/ClipboardButton.tsx
Raubzeug: why? what if we decide not to hide tooltip? it seems we will still need to reset closingDelay.

lxndr: What do you mean by deciding not to hide the tooltip? It has to be hidden even if the user quickly leaves and hovers the button again. We only show the tooltip if it's gone and user hovers the button.

Raubzeug (approved):

PR #1741, opened 2024-08-09
feat(PlaceholderContainer): add component and showcases

Marginy605:


src/components/PlaceholderContainer/PlaceholderContainer.tsx
```diff
@@ -0,0 +1,124 @@
...
+import './PlaceholderContainer.scss';
+
+const b = block(componentClassName);
+
+export class PlaceholderContainer extends React.Component<
+    PlaceholderContainerInnerProps,
+    PlaceholderContainerState
+> {
+    static Direction = Direction;
+    static Align = Align;
```
amje, line 23: Let's remove these static fields and use raw strings like we did for `size`

src/components/PlaceholderContainer/PlaceholderContainer.tsx
```diff
@@ -0,0 +1,124 @@
...
+    PlaceholderContainerImageSettingsProps,
+    PlaceholderContainerInnerProps,
+    PlaceholderContainerState,
+} from './types';
+
+import './PlaceholderContainer.scss';
+
+const b = block(componentClassName);
+
+export class PlaceholderContainer extends React.Component<
```
amje, line 18: Can we rewrite to function component?

src/components/PlaceholderContainer/PlaceholderContainerAction.tsx
```diff
@@ -0,0 +1,45 @@
...
+    size?: ButtonSize;
+    handler?: (event: React.MouseEvent<HTMLButtonElement>) => void;
+    href?: string;
+}
+
+interface PlaceholderContainerActionProps {
+    action: Action;
+}
+
+export const PlaceholderContainerAction = ({action}: PlaceholderContainerActionProps) => {
```
amje, line 27: There is no advantages to keep this markup as separate file, it has no styles and simply proxy props for Button

src/components/PlaceholderContainer/PlaceholderContainer.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+        if (renderAction) {
+            return renderAction();
+        }
+
+        if (!action) {
+            return null;
+        }
+
+        if (Array.isArray(action)) {
```
amje, line 104: Let's name it `actions` and type it always as array. And we could also merge two props `action` and `renderAction`:

```
actions?: PlaceholderContainerActionProps[] | React.ReactNode;
```

src/components/PlaceholderContainer/PlaceholderContainerAction.tsx
```diff
@@ -0,0 +1,45 @@
...
+    loading?: boolean;
+    disabled?: boolean;
+    view?: ButtonView;
+    size?: ButtonSize;
+    handler?: (event: React.MouseEvent<HTMLButtonElement>) => void;
+    href?: string;
+}
+
+interface PlaceholderContainerActionProps {
+    action: Action;
```
amje, line 24: `Action` could be just `PlaceholderContainerActionProps `

src/components/PlaceholderContainer/__stories__/PlaceholderContainer.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+                        enabled: false,
+                    },
+                ],
+            },
+        },
+    },
+} as Meta;
+
+const ShowcaseTemplate: StoryFn = () => <PlaceholderContainerShowcase />;
+export const Showcase = ShowcaseTemplate.bind({});
```
amje, line 28: Let's have separate stories for each prop of the component (size, direction, image, action) like we do in most other component's stories

src/components/PlaceholderContainer/constants.ts
```diff
@@ -0,0 +1,11 @@
+export const componentClassName = 'placeholder-container';
+
+export const Direction = {
+    Row: 'row',
+    Column: 'column',
+} as const;
+
+export const Align = {
```
amje, line 8: Suggested earlier to remove these and use raw strings

src/components/PlaceholderContainer/types.ts
```diff
@@ -0,0 +1,38 @@
+import type React from 'react';
+
+import type {Action} from './PlaceholderContainerAction';
+import type {Align, Direction} from './constants';
+
+type Size = 's' | 'm' | 'l' | 'promo';
+
+type PlaceholderContainerImageNodeProps = NonNullable<React.ReactNode>;
+
+export type PlaceholderContainerImageSettingsProps = {
```
amje, line 10:
```suggestion
export type PlaceholderContainerImageProps = {
```

src/components/PlaceholderContainer/types.ts
```diff
@@ -0,0 +1,38 @@
...
+    url: string;
+    alt?: string;
+};
+
+interface PlaceholderContainerGeneralProps {
+    title?: string;
+    description?: React.ReactNode;
+    renderContent?: () => React.ReactNode;
```
amje, line 18: Maybe change this to `content?: React.ReactNode`?

src/components/PlaceholderContainer/types.ts
```diff
@@ -0,0 +1,38 @@
...
+    action?: Action | Action[];
+    renderAction?: () => React.ReactNode;
+    className?: string;
+    image: PlaceholderContainerImageNodeProps | PlaceholderContainerImageSettingsProps;
+}
+
+export interface PlaceholderContainerDefaultProps {
```
amje, line 25: We don't use separate props for defaults

src/components/PlaceholderContainer/types.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export type PlaceholderContainerInnerProps = PlaceholderContainerGeneralProps &
+    PlaceholderContainerDefaultProps;
+
+export interface PlaceholderContainerProps
+    extends PlaceholderContainerGeneralProps,
+        Partial<PlaceholderContainerDefaultProps> {}
+
+export interface PlaceholderContainerState {}
```
amje, line 38: Let's remove empty state type

amje (changes requested):

src/components/PlaceholderContainer/PlaceholderContainer.tsx
```diff
@@ -0,0 +1,107 @@
...
+        if (!description) {
+            return null;
+        }
+
+        return <div className={b('description')}>{description}</div>;
+    };
+
+    const renderImage = (): NonNullable<React.ReactNode> => {
+        if (typeof image === 'object' && 'url' in image) {
+            return <img src={image.url} alt={image.alt || ''} />;
```
amje, line 59: If we refer to attributes when object is used then we should call it the same, `src` instead of `url`. Maybe we'll rename it?

src/components/PlaceholderContainer/__stories__/PlaceholderContainer.stories.tsx
```diff
@@ -0,0 +1,266 @@
...
+    };
+
+const descriptionProps = {
+    description:
+        'Some long descriptionProps text that can contain of long long very long text etc. It can be repeated like this. Some long descriptionProps text that can contain of long long very long text etc.',
+    promoDescription:
+        "Comparing to 'L' size promo size has full width of the content block, a larger title size and alignment",
+};
+
+export const Size: Story = {
```
amje, line 141: First story should be named `Default` with the default props, so user can play around with the component

src/components/PlaceholderContainer/__stories__/PlaceholderContainer.stories.tsx
```diff
@@ -0,0 +1,266 @@
...
+                        description={descriptionProps.description}
+                        direction="row"
+                        size="s"
+                    />
+                </ShowcaseItem>
+                <ShowcaseItem title="Size m">
+                    <PlaceholderContainer
+                        {...placeholderContainerProps}
+                        description={descriptionProps.description}
+                        direction="row"
```
amje, line 157: `direction` cases should be covered in its own story, same is for `align`. So basically there should be at least 5 stories: Default, Size, Direction, Align, Actions.

amje (changes requested):

src/components/PlaceholderContainer/PlaceholderContainer.tsx
```diff
@@ -0,0 +1,107 @@
...
+            </Button>
+        </div>
+    );
+};
+
+export const PlaceholderContainer = ({
+    direction = 'row',
+    align = 'center',
+    size = 'l',
+    className = b(),
```
amje, line 35: This className should be an addition to `b()`, not replacement: `b(mods, className)`

amje (changes requested):

src/components/PlaceholderContainer/PlaceholderContainer.tsx
```diff
@@ -0,0 +1,107 @@
...
+        return (
+            <div className={b('content', {size})}>
+                {contentNode}
+                {renderAction()}
+            </div>
+        );
+    };
+
+    return (
+        <div className={b({direction, align, size}, [className])}>
```
amje, line 100: Second argument could be just string

PR #1742, opened 2024-08-10
feat(RadioButton): redesign to increase contrast in component

### LINKED ISSUE #1732: More contrast look of RadioButton
Internal ticket - UXRFC-314

benax-se:
Closes #1732


src/components/RadioButton/RadioButton.scss
```diff
@@ -3,45 +3,94 @@
...
+            background-color: var(--g-color-line-generic);
+        }
+
+        &::after {
             content: '';
             position: absolute;
-            z-index: -1;
-            inset: 3px;
-            border-radius: var(--_--border-radius-inner);
+            z-index: -2;
```
zamkovskaya, line 39: In which case is z-index: -1 not sufficient?

zamkovskaya (approved):

src/components/RadioButton/RadioButton.scss
benax-se: I forgot to lower it to -1 after testing different layout options.
Fixed it

amje:
This PR might be fixing the issue #1923

ogonkov:
@amje looks like this PR could soulve this problem, yes, is there eta when this PR could be merged?

src/components/RadioButton/RadioButton.scss
```diff
@@ -3,45 +3,94 @@
...
         text-align: center;
-        border-radius: var(--_--border-radius-inner);
+
         cursor: pointer;
         transform: scale(1);
-        transition: color 0.15s linear;
+        transition: color var(--_--transition-time) linear;
 
-        &-outline {
+        &::before {
```
ogonkov, line 27: What is the reason to use pseudo-elements? Why just don't change border and of element itself?

PR #1764, opened 2024-08-22
feat(MobileProvider): calculate device type if mobile prop is not provided

viladimiru:
This pr is part of [UXRFC-278](https://st.yandex-team.ru/UXRFC-278)

I think this pr provide breaking changes, which need to release with major version


src/components/mobile/MobileProvider.tsx
```diff
@@ -49,6 +55,14 @@ export function MobileProvider({
         [useHistory],
     );
 
+    const [mobile, setMobile] = useControlledState(controlledMobile, isDeviceTypeMobile());
+
+    React.useEffect(() => {
+        if (controlledMobile === undefined) {
+            setMobile(isDeviceTypeMobile());
+        }
+    }, [controlledMobile, width]);
```
ValeraS, line 64: How does device type depend on width? UAParser uses only the user agent string to determine the device and its type.

ValeraS (changes requested):

src/components/mobile/MobileProvider.tsx
viladimiru: Useragent may change if we change device type from devtools

ValeraS: But users won't  change useraget. Is it something just for development purpose?

viladimiru: I think if user have ability to change useragent through devtools, uikit have to handle this behaviour, but it's also for development/testing purposes

Do you think it's expensive?

ValeraS: I still don't understand how it depends on the width. And if the user changes the useragent, they can refresh the page to apply the changes.

viladimiru: I think you right, it's maybe wrong to observing width because it's doesn't mean that useragent was changed, but I didn't find how to observe user agent changes directly, so I think better to detect device type only once, thank you

viladimiru: I invited @amje with @korvin89 in this discussion, I sure that your point is still right, but I think we can loose flexibility if we will not recalculate this property somehow, for example depending of width

viladimiru: @amje agree with your point, so current implementation is actual, can you check please?

PR #1765, opened 2024-08-23
chore: add utility function to detect touch screen

viladimiru:
This pr is part of [UXRFC-278](https://st.yandex-team.ru/UXRFC-278)


src/components/mobile/MobileProvider.tsx
```diff
@@ -53,14 +57,29 @@ export function MobileProvider({
         document.body.classList.toggle(rootMobileClassName, mobile);
     }, [rootMobileClassName, mobile]);
 
+    const touch = React.useMemo(() => isTouchScreen(), [width]);
```
ValeraS, line 60: How does this depend on width?

ValeraS (changes requested):

src/components/mobile/MobileProvider.tsx
viladimiru: If we change device type from devtools, navigator can update context, for example if I toggle from desktop version to mobile, `navigator.maxTouchPoints` toggling value between 1 and 0

viladimiru: We discuses this theme in more details [here](https://github.com/gravity-ui/uikit/pull/1764#discussion_r1728742337).
This case is same, devtools may change `navigator.maxTouchPoints` value, but we can't subscribe on this value directly. I deleted width observing from here because it's wrong direction for observing, unfortunately I didn't find how to subscribe on property changes of `isTouchScreen` method

viladimiru: We discussed with @amje and @korvin89 should we use `MobileProvider` for inner statement, which uncontrolled by user, and decided to get rid of this and make special helper

src/components/mobile/helpers.ts
amje, line 1: Let's place this file in `src/utils/dom.ts`

src/components/mobile/helpers.ts
```diff
@@ -0,0 +1,11 @@
+export function isTouch(): boolean {
```
amje, line 1:
```suggestion
export function isTouchDevice(): boolean {
```

amje, line 1: Let's add check for window presence, if no window then return false

src/components/mobile/helpers.ts
```diff
@@ -0,0 +1,11 @@
...
+    if (window.PointerEvent) {
+        return navigator?.maxTouchPoints > 0;
+    }
```
amje, line 4: Is this correct condition?

Let's something like this:
```js
if (typeof navigation !== 'undefined' && navigator.maxTouchPoints > 0) {
    return true;
}
```

amje (changes requested):

src/components/mobile/helpers.ts
viladimiru: I think it's more effective, thanks!

src/utils/dom.ts
```diff
@@ -0,0 +1,15 @@
+export function isTouchDevice(): boolean {
+    if (!window) {
```
amje, line 2: This would error if window is not in the scope at all. Need something like that: `typeof window === 'undefined'`

src/utils/dom.ts
```diff
@@ -0,0 +1,15 @@
...
+
+    if (typeof navigator !== 'undefined' && navigator.maxTouchPoints > 0) {
+        return true;
+    }
+
+    if (window.matchMedia && window.matchMedia('(any-pointer:coarse)').matches) {
+        return true;
+    }
+
+    return Boolean(window.TouchEvent || 'ontouchstart' in window);
```
ValeraS, line 14: It looks like this check will be true for non-touch devices too, right?

ValeraS (changes requested):

src/utils/dom.ts
viladimiru: Seem's like browser provides `TouchEvent` API also [for devices with trackpad](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent#:~:text=touch%20screen%20or%20trackpad), so you right! I deleted `window.TouchEvent` from this condition

ValeraS (approved):

amje (approved):

PR #1772, opened 2024-08-26
test(Breadcrumbs): add smoke visual tests

itwillwork:


korvin89:
We don't need tests for this component, because we already have a new one in lab section

itwillwork:
> We don't need tests for this component, because we already have a new one in lab section

ok, rejected )

PR #1776, opened 2024-08-26
test(ControlLabel): add smoke visual tests

itwillwork:


korvin89:
@itwillwork @amje Hey! I suggest to skip test for inner components like this. We check it's functionality in other public components

PR #1813, opened 2024-08-26
test(Overlay): add smoke visual tests

itwillwork:


Vladeeg (approved):

Vladeeg:
@amje just in case, could you please review this PR too because you have more experience here?

PR #1826, opened 2024-08-28
feat(NumberInput): add new component

DaryaLari:


src/components/lab/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,320 @@
...
+    } = {...defaultBaseTextInputProps, ...props};
+
+    const {
+        min,
+        max,
+        step: baseStep,
+        value,
+        defaultValue,
+        shiftMultiplier,
+    } = getInternalVariables({
```
amje, line 148: `getInternalState` seems to be more descriptive name

src/components/lab/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,320 @@
...
+            handleDecrement();
+        } else if (e.key === KeyCode.ARROW_UP) {
+            handleIncrement();
+        }
+        onKeyDown?.(e);
+    };
+
+    const handleKeyUp: React.KeyboardEventHandler<HTMLInputElement> = (e) => {
+        if (e.key === KeyCode.SHIFT) {
+            setStep(baseStep);
```
amje, line 225: We also should reset step on input's blur, otherwise it stucks on Tab/Shift+Tab navigation

src/components/lab/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,320 @@
...
+                max,
+            });
+            if (value !== clampedValue) {
+                onUpdate?.(clampedValue);
+            }
+        }
+        onBlur?.(e);
+    };
+
+    const handleChange: React.ChangeEventHandler<HTMLInputElement> = (e) => {
```
amje, line 251: We can only use `onUpdate` for updating the value. `onChange` handler is redundant

src/components/lab/NumberInput/NumericArrows/NumericArrows.scss
```diff
@@ -0,0 +1,49 @@
...
+$block: '.#{variables.$ns}numeric-arrows';
+
+#{$block} {
+    --_--numeric-arrows-border-color: var(
+        --_-g-number-input-numeric-arrows-border-color,
+        var(--g-text-input-border-color, transparent)
+    );
+
+    --_--border-width: var(--g-text-input-border-width, 1px);
+    border-style: solid;
```
amje, line 15: Seems like these styles are duplicated

src/components/lab/NumberInput/NumericArrows/NumericArrows.scss
```diff
@@ -0,0 +1,49 @@
+@use '../../../../../styles/mixins';
+@use '../../../variables';
+@use '../../../controls/mixins.scss' as control-mixins;
+@use '../../../controls/variables.scss' as control-variables;
```
amje, line 4: We don't use any of these in the file

src/components/lab/NumberInput/index.ts
```diff
@@ -0,0 +1,5 @@
+export {NumberInput} from './NumberInput';
+export type {NumberInputProps} from './NumberInput';
+
+export {getNumericInputValidator} from './utils';
+export type {GetNumericInputValidatorParams} from './utils';
```
amje, line 5: Aren't these internal helpers? Why we export them?

src/components/lab/NumberInput/utils.ts
```diff
@@ -0,0 +1,262 @@
...
+            return smallerPossibleValue;
+        }
+        if (greaterPossibleValue <= max) {
+            return greaterPossibleValue;
+        }
+    }
+    return clampedValue;
+}
+
+export function warnAboutInvalidProps({
```
amje, line 195: This will span a lot. What's the benefit of this? Component won't allow invalid value anyway.

amje (changes requested):

src/components/controls/TextInput/TextInput.tsx
```diff
@@ -94,7 +104,7 @@ export const TextInput = React.forwardRef<HTMLSpanElement, TextInputProps>(
             note,
             onUpdate,
             onChange,
-        } = props;
+        } = propsWithDefaults;
```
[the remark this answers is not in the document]
DaryaLari: Because some of these TextInput defaults (which were moved to a separate object) are also required in NumberInput

src/components/lab/NumberInput/NumericArrows/NumericArrows.scss
DaryaLari: Which styles and where is their duplicate is located?

src/components/lab/NumberInput/index.ts
DaryaLari: No, they are not used in component. This utility function is for external use only. NumberInput itself doesn't show any validation errors

src/components/lab/NumberInput/utils.ts
DaryaLari: It will spam only with invalid props passed.
I think it would be better to notify programmer about invalid component usage6 but not just fix props silently

src/components/lab/NumberInput/NumericArrows/NumericArrows.scss
amje: border styles between NumericArrows.scss and NumberInput.scss

src/components/controls/TextInput/TextInput.tsx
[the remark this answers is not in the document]
amje: Can we use it in React way then? `TextInput.defaultProps = {...}`

src/components/lab/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,320 @@
...
+
+    return (
+        <TextInput
+            {...props}
+            className={b({size, view, state}, className)}
+            controlProps={{
+                onInput: handleInput,
+                ...props.controlProps,
+                onWheel: allowMouseWheel && active ? handleWheel : undefined,
+                role: 'spinbutton',
```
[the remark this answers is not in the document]
DaryaLari: Native input type number has spinbutton role

src/components/controls/TextInput/TextInput.tsx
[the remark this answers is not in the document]
DaryaLari: If you'd help me to deal with fc typings with defaultProps^ then I'll change it to react way
I would really appreciate you help

src/components/lab/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,320 @@
...
+    endContent?: React.ReactNode;
+    /** An optional element displayed under the lower right corner of the control and sharing the place with the error container */
+    note?: React.ReactNode;
+    /**Describes the validation state */
+    validationState?: 'invalid' | undefined;
+
+    /** Shows inncrement/decrement buttons at the end of control
+     * @default true
+     */
+    hasControls?: boolean;
```
ValeraS, line 51: Let's use a reverse flag.

src/components/lab/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,320 @@
...
+     */
+    max?: number;
+    /** Delta for incrementing/decrementing entered value with arrow keyboard buttons or component controls
+     * @default 1
+     */
+    step?: number;
+    /** Step multiplier when shift button is pressed
+     * @default 10
+     */
+    shiftMultiplier?: number;
```
ValeraS, line 67: Why use Shift and not Page Up / Page Down ?

src/components/lab/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,320 @@
...
+    const handleWheel: React.WheelEventHandler<HTMLInputElement> = (e) => {
+        const delta = e.shiftKey ? e.deltaX : e.deltaY;
+        if (delta > 0) {
+            handleIncrement();
+        } else if (delta < 0) {
+            handleDecrement();
+        }
+    };
+
+    const handleKeyDown: React.KeyboardEventHandler<HTMLInputElement> = (e) => {
```
ValeraS, line 218: If min/max are set, we can set the value to min/max on Home/End.

src/components/lab/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,320 @@
...
+        updateCursorPosition(innerControlRef, v, preparedStringValue);
+        const {valid, value: parsedNumberValue} = getParsedValue(preparedStringValue);
+        if (valid && parsedNumberValue !== value) {
+            onUpdate?.(parsedNumberValue);
+        }
+    };
+
+    const handleInput: React.FormEventHandler<HTMLInputElement> = (e) => {
+        const preparedStringValue = getPossibleNumberSubstring(e.currentTarget.value, allowDecimal);
+        updateCursorPosition(innerControlRef, e.currentTarget.value, preparedStringValue);
```
ValeraS, line 270: When you increase the value (press the up arrow), the cursor moves to the beginning and then to the end.

ValeraS (changes requested):

src/components/lab/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,312 @@
...
+        setInputValue((currentInputValue) => {
+            if (!areStringRepresentationOfNumbersEqual(currentInputValue, stringPropsValue)) {
+                return stringPropsValue;
+            }
+            return currentInputValue;
+        });
+    }, [value]);
+
+    const clamp = true;
+    const [step, setStep] = React.useState(baseStep);
```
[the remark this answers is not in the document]
DaryaLari: I thought that it would be helpful for determining current step while pressing arrow buttons (controls in component) with shift button pressed (realized) 
But then I realized, that it would require adding new wrapperRef prop for detecting focusWithin, and change step on focusWithin state change. 
If you think that shift pressing shouldn't change the step value for component
If I'm not mistaken, pointer events don't have info, if shift is pressed or not now

src/components/lab/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,312 @@
...
+        clampToNearestStepValue({value: value + step, step, min, max}) > value;
+
+    const canDecrementNumber =
+        clampToNearestStepValue({value: value - step, step, min, max}) < value;
+
+    const innerControlRef = React.useRef<HTMLInputElement>(null);
+    const fieldRef = useFormResetHandler({
+        initialValue: defaultValue,
+        onReset: onUpdate ?? voidFunction,
+    });
```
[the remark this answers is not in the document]
DaryaLari: uncontrolled type of component is not supported here yet

src/components/lab/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,320 @@
...
+    const handleRef = useForkRef(props.controlRef, innerControlRef, fieldRef);
+
+    const handleIncrement = () => {
+        if (canIncrementNumber) {
+            onUpdate?.(
```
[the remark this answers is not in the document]
DaryaLari: Yes, it is not supported (yet?)

src/components/lab/NumberInput/NumberInput.tsx
korvin89: We've discussed it with @ValeraS @amje and decided to use `hiddenControls` without default

src/components/lab/NumberInput/NumberInput.tsx
korvin89: We've discussed it with @ValeraS @amje and decided to keep this name

src/components/lab/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,312 @@
...
+    };
+
+    const handleDecrement = () => {
+        if (canDecrementNumber) {
+            onUpdate?.(clampToNearestStepValue({value: value - step, step: baseStep, min, max}));
+        }
+    };
+
+    const handleWheel: React.WheelEventHandler<HTMLInputElement> = (e) => {
+        const delta = e.shiftKey ? e.deltaX : e.deltaY;
```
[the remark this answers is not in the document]
DaryaLari: Maybe it depends on system scroll settings?
On my computer scroll to the right increases value as expected

src/components/lab/NumberInput/NumberInput.tsx
DaryaLari: But in native input type=number on End/Home just the caret moves to the end/beginning of the content...
Wouldn't this difference in behaviour with native control be confusing?

DaryaLari:
@ValeraS 
There is no ability to reply to this comment directly:
> From default props only size and view are needed, right? Let's set default only for them ('m', and 'normal') in NumberInput.

Also disabled is used in NumberInput directly.
But, wouldn't it be better to store default props in one object to be sure that they wouldn't differ in different controls one day? And what if we'll need to use  more default props from TextInput in NUmberInput itself?

src/components/controls/TextInput/TextInput.tsx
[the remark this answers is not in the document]
DaryaLari: Also disabled is used in NumberInput directly.
But, wouldn't it be better to store default props in one object to be sure that they wouldn't differ in different controls one day? And what if we'll need to use more default props from TextInput in NumberInput itself?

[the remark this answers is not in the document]
DaryaLari: Ok, I've checked that other controls define default values of these props themselves, so I reverted these changes in TextInput

src/components/lab/NumberInput/NumberInput.tsx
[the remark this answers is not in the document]
DaryaLari: @ValeraS I've added uncontrolled input type support

DaryaLari:
@ValeraS @amje @korvin89 
Hopefully I've fixed all previous comments and I'm waiting for your review again

src/components/lab/NumberInput/NumberInput.scss
```diff
@@ -0,0 +1,44 @@
...
+            --_--textinput-end-padding: 3px;
+        }
+
+        &_xl {
+            --_--textinput-end-padding: 3px;
+        }
+    }
+
+    &_view_normal {
+        --_-g-number-input-numeric-arrows-border-color: var(--g-color-line-generic);
```
amje, line 25:
```suggestion
        --_--arrows-border-color: var(--g-color-line-generic);
```

It's a private var so no need to specify full name

src/components/lab/NumberInput/NumberInput.scss
```diff
@@ -0,0 +1,44 @@
...
+        --_--border-color-hover: var(--g-color-line-generic-hover);
+        --_--border-color-active: var(--g-color-line-generic-active);
```
amje, line 27: What these variables for? They duplicate TextInput styles

src/components/lab/NumberInput/NumberInput.scss
```diff
@@ -0,0 +1,44 @@
...
+        &#{$block}_state_error {
+            --_-g-number-input-numeric-arrows-border-color: var(--g-color-line-danger);
+        }
+    }
+
+    &_view_clear {
+        --_-g-number-input-numeric-arrows-border-color: transparent;
+    }
+
+    &__numeric-arrows {
```
amje, line 38:
```suggestion
    &__arrows {
```

Let's be compact here

src/components/lab/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,326 @@
...
+    /** Indicates that the user cannot change control's value */
+    readOnly?: boolean;
+    /** User`s node rendered before label and input node */
+    startContent?: React.ReactNode;
+    /** User`s node rendered after input node and clear button */
+    endContent?: React.ReactNode;
+    /** An optional element displayed under the lower right corner of the control and sharing the place with the error container */
+    note?: React.ReactNode;
+    /**Describes the validation state */
+    validationState?: 'invalid' | undefined;
```
amje, line 46: This prop already in the `BaseInputControlProps`, no need to specify it again

src/components/lab/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,326 @@
...
+     */
+    step?: number;
+    /** Step multiplier when shift button is pressed
+     * @default 10
+     */
+    shiftMultiplier?: number;
+    /** Enables changing value by scrolling mousewheel on with cursor on the input
+     * @default  false
+     */
+    allowMouseWheel?: boolean;
```
amje, line 70: We've decided to remove the wheel behaviour completely. It turned out that currently it brings a lot of bugs and inconsistency between mouse and touchpad devices. Maybe we reimplement it in the future.

src/components/lab/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,326 @@
...
+                min,
+                max,
+                direction: 'up',
+            });
+            setValue?.(newValue);
+            setInputValue(newValue.toString());
+        }
+    };
+
+    const handleDecrement = (
```
amje, line 185: handleIncrement and handleDecrement look very alike. Let's combine them into one function with the signature `handleValueDelta(delta: number)`, and use like `handleValueDelta(1)`, `handleValueDelta(-10)` (if shift is pressed, and derection is negative).

src/components/lab/NumberInput/NumberInput.tsx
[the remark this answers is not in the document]
amje: Ok, let's keep it.

src/components/lab/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,326 @@
...
+    return (
+        <TextInput
+            {...props}
+            className={b({size, view, state}, className)}
+            controlProps={{
+                onInput: handleInput,
+                ...props.controlProps,
+                onWheel: allowMouseWheel && active ? handleWheel : undefined,
+                role: 'spinbutton',
+                inputMode: 'numeric',
```
amje, line 290: when `allowDecimal = true`, we should pass `decimal` here to have correct keyborad on mobile devices

src/components/lab/NumberInput/NumericArrows/NumericArrows.scss
```diff
@@ -0,0 +1,45 @@
+@use '../../../variables';
+
+$block: '.#{variables.$ns}numeric-arrows';
+
+#{$block} {
+    --_--numeric-arrows-border-color: var(
```
amje, line 6:
```suggestion
    --_--arrows-border-color: var(
```

src/components/lab/NumberInput/NumericArrows/NumericArrows.tsx
```diff
@@ -0,0 +1,73 @@
...
+    onDownClick,
+    ...restProps
+}: NumericArrowsProps) {
+    const commonBtnProps: Partial<ButtonProps> = {
+        size: 's',
+        pin: 'brick-brick',
+        view: 'flat-secondary',
+        disabled,
+        tabIndex: -1,
+        width: 'max',
```
amje, line 42: Let's hide the buttons from screenreaders via `aria-hidden="true"`

src/components/lab/NumberInput/index.ts
amje: Then why keep this code? We cannot maintain it properly without specific usecases

amje (changes requested):

amje:
Extra zero digits not deleted on blur:

https://github.com/user-attachments/assets/2c277406-7053-4c09-bfba-44262c4fbb64

Typing dot when `allowDecimal = false` deletes one digit at time, should do nothing:

https://github.com/user-attachments/assets/cedd6164-5acb-40d6-8c4f-6726004b2f68

Leading dot not deleted on blur:

https://github.com/user-attachments/assets/d48529ad-bf53-4546-ae5b-e129184efb2b

amje (approved):

ValeraS (approved):

Issue #1737, opened 2024-08-07
Switching language on Firefox Ubuntu close select


dsleepo:
1. Open filterable select
2. Press `super` + `space`

Actual result: option's popup closed
Expected: option's popup remains open

[DELL-LATITUDE-E7440-(1920x1080,-Intel-Core-i5-1-6-ГГц,-RAM-8-ГБ,-SSD-128-ГБ,-Win7-Pro-64)_Ubuntu-20-04_07-08-2024_14-54-24.webm](https://github.com/user-attachments/assets/3e3e951a-d91b-4a41-ac85-6dcd1d196c4e)

lxndr:
I can reproduce it having GNOME 46, Wayland, Firefox 129. If I have time, I could take a closer look what's going on. Not a lot of Linux users around here.

lxndr:
I can reproduce it having GNOME 46, Wayland, Firefox 129. If I have time, I could take a closer look what's going on. Not a lot of Linux users around here.

lunaticenslaved:
If you do not mind I take part.

I reproduced the issue on Manjaro + Firefox 130 with `super` + `space`. Moreover I emulated it with Macbook:
1. In Firefox open any select.
2. Click `control` + `space` to open search bar.

Apparently blur event is fired when FIrefox's window loses focus. It was discussed here: https://stackoverflow.com/questions/19700476/avoid-element-blur-handler-when-window-blur-invoked-browser-loses-focus

If you please I want to suggest a solution: #1836

PR #1836, opened 2024-09-05
fix(useFocusWithin): ignore blur event when window loses focus

lunaticenslaved:
Issue #1737

Apparently when Firefox's window loses focus a blur event is fired. 

It can be reproduced on Linux GNOME with `super` + `space` to switch language. I do not know whether or not it works on KDE.
It can be emulated on Mac with `control` + `space` to open search bar.

On Windows with switching language by `win` + `space` it works fine.

Maybe this solution can help.

### DIFF
--- a/src/hooks/useFocusWithin/useFocusWithin.ts
+++ b/src/hooks/useFocusWithin/useFocusWithin.ts
@@ -185,9 +185,10 @@ function useFocusEvents({
     const onBlurHandler = React.useCallback(
         (event: React.FocusEvent) => {
             if (
-                event.relatedTarget === null ||
-                event.relatedTarget === document.body ||
-                event.relatedTarget === (document as EventTarget)
+                document.activeElement !== event.target &&
+                (event.relatedTarget === null ||
+                    event.relatedTarget === document.body ||
+                    event.relatedTarget === (document as EventTarget))
             ) {
                 onBlur(event);
                 targetRef.current = null;


src/hooks/useFocusWithin/useFocusWithin.ts
```diff
@@ -96,6 +96,17 @@ export function useFocusWithin(props: UseFocusWithinProps) {
...
+                document.activeElement === event.target &&
+                (event.target instanceof HTMLButtonElement ||
+                    event.target instanceof HTMLInputElement ||
+                    event.target instanceof HTMLTextAreaElement ||
+                    event.target instanceof HTMLSelectElement) &&
+                !event.target.disabled
+            ) {
+                return;
+            }
+
```
ValeraS, line 109: Let's remove this and add `document.activeElement !== event.target && (...)` to that check https://github.com/gravity-ui/uikit/blob/4c3d070a74cee72a5017ac6df37b3f8cf22b2122/src/hooks/useFocusWithin/useFocusWithin.ts#L187-L191

ValeraS (changes requested):

korvin89 (commented):
Add comment, please. It looks a little bit weird without context :)

lunaticenslaved:
> Add comment, please. It looks a little bit weird without context :)

Removed the block of code. It looks more obvious now I hope.

src/hooks/useFocusWithin/useFocusWithin.ts
lunaticenslaved: Done.

ValeraS (approved):

PR #1840, opened 2024-09-06
feat(Card): enable onClick event handler on selected card

DaryaLari:


amje:
@DaryaLari Seems there is an issue with styles. Selected styles should not respect hover state
https://github.com/user-attachments/assets/a46bd8c2-aa32-49eb-aa90-9e81a7de999b

DaryaLari:
> @DaryaLari Seems there is an issue with styles. Selected styles should not respect hover state

@amje Why do you consider it as an issue? In my universe, hover is an indication that you can interact with the element and it is not just a picture. Isn't it?..

amje:
> > @DaryaLari Seems there is an issue with styles. Selected styles should not respect hover state
> 
> @amje Why do you consider it as an issue? In my universe, hover is an indication that you can interact with the element and it is not just a picture. Isn't it?..

When selected state is not distinguished from not selected when it's hovered - it's bad. To know the currest state user must unhover first.

amje (approved):

Lunory (approved):

PR #1845, opened 2024-09-10
feat(TextInput): add unstable_endContent prop

DaryaLari:


korvin89:
Something wrong with `s` size

https://github.com/user-attachments/assets/f5e6a9d8-a046-4fb6-bdac-35aaa079a72b

DaryaLari:
I think, it would be nice also to make endContent padding a part of public css-api, so that someone (like me 🙄) would have a legal way to render some control close to the component border without spacing (now there would always be some space between endContent and border)
image

korvin89 (approved):

PR #1852, opened 2024-09-12
feat!: use native copy method

### LINKED ISSUE #658: Get rid of `react-copy-to-clipboard`
I guess we can use native browser APIs instead.

ogonkov:
Breaking changes:

- native copy method used
- `options` of `copy-to-clipboard` removed from props
- allow `children` to be just `ReactElement`

Fixes #658


src/utils/copyText.ts
```diff
@@ -0,0 +1,7 @@
+export function copyText(text: string) {
+    if (navigator.clipboard.writeText) {
```
Raubzeug, line 2: `navigator?.clipboard?.writeText` ?

src/components/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -27,12 +33,46 @@ export function CopyToClipboard(props: CopyToClipboardProps) {
...
     );
 
+    const onClickWithCopy: React.MouseEventHandler<HTMLElement> = React.useCallback(
+        (event) => {
+            textRef.current = text;
+
+            copyText(text).then(
+                () => {
+                    if (text === textRef.current) {
+                        handleCopy(text, true);
```
Raubzeug, line 43: maybe move it to a helper like

```
const copyText = (result: boolean, event: MouseEvent) => {
  handleCopy(text, result);
  if (typeof content.props?.onClick === 'function') {
    content.props.onClick(event);
  }
}
```

Raubzeug (changes requested):

src/utils/copyText.ts
ogonkov: I think `if` is better for possible stackatrace investigation

Raubzeug: but it is possible that application will fail, if there's no `clipboard` in `navigator` proto. We don't use `try ... catch` when using `copyText`.
I suggested not removing `if`, but add checks like `if (navigator?.clipboard?.writeText) { ... }`

Raubzeug (changes requested):

src/components/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -2,23 +2,27 @@
...
+    const textRef = React.useRef(text);
     const [status, setStatus] = React.useState<CopyToClipboardStatus>(INITIAL_STATUS);
 
     const timerIdRef = React.useRef<number>();
 
-    const content = React.useMemo(() => children(status), [children, status]);
+    const content = React.useMemo<React.ReactElement<React.HTMLAttributes<HTMLElement>>>(
+        () => children(status),
+        [children, status],
+    );
```
amje, line 22: It's no benefit of useMemo at all, children as function typically inline in jsx, so every render is new instance

src/components/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -27,15 +31,37 @@ export function CopyToClipboard(props: CopyToClipboardProps) {
...
 
+    const onClickWithCopy: React.MouseEventHandler<HTMLElement> = React.useCallback(
+        (event) => {
+            textRef.current = text;
+
+            function copy(result: boolean) {
+                if (text === textRef.current) {
+                    handleCopy(text, result);
+
+                    content.props?.onClick?.(event);
```
amje, line 42: Shouldn't we call this `onClick` unconditionally?

src/demo/colors/ColorPanel.tsx
```diff
@@ -39,18 +38,20 @@ export function ColorPanel(props: ColorPanelProps) {
...
             return (
                 <div className="color-panel__card" key={color.name}>
-                    <ReactCopyToClipboard text={copyText}>
-                        <div
-                            className={`color-panel__card-box ${boxBorders}`}
-                            style={{background: `var(${varName})`}}
-                        />
-                    </ReactCopyToClipboard>
+                    <CopyToClipboard text={copyText}>
+                        {() => (
```
amje, line 42: Let's support normal children type as well as it was in `ReactCopyToClipboard `? I find it usefull, if I don't need status feedback

src/utils/copyText.ts
```diff
@@ -0,0 +1,7 @@
+export function copyText(text: string) {
```
amje, line 1: This file is not needed to be at the root utils, place it near the component please

amje (changes requested):

src/components/CopyToClipboard/CopyToClipboard.tsx
ogonkov: This was here before, i just add types here

src/components/CopyToClipboard/CopyToClipboard.tsx
ogonkov: This conditional is to check that props were not changed after promise resolve

src/demo/colors/ColorPanel.tsx
ogonkov: This PR is aimed to replace `react-copy-to-clipboard`, we can make this change in separate PR

src/components/CopyToClipboard/CopyToClipboard.tsx
amje: Let's remove this memo wrap

amje (changes requested):

SeqviriouM (approved):

amje (approved):

PR #1856, opened 2024-09-13
fix(Button): selected styles override disabled styles in button

mariamelior:


amje (approved):

ogonkov:
looks like case with bug is not covered by screenshot tests

PR #1859, opened 2024-09-16
feat(i18n): functionality for adding translations into other languages

ValeraS:


src/components/utils/addComponentKeysets/addComponentKeysets.ts
amje, line 1: Let's name the directory `utils/i18n`?

amje (approved):

PR #1873, opened 2024-09-23
feat(ActionsPanel): add component and showcases

jhoncool:
Component moved from `@gravity-ui/components`


src/components/ActionsPanel/components/hooks/useObserveIntersection.ts
```diff
@@ -0,0 +1,77 @@
...
+        });
+
+        Array.from(parentRef.current?.children || []).forEach((item) => {
+            if (item.hasAttribute(OBSERVER_TARGET_ATTR)) {
+                observer.observe(item);
+            }
+        });
+
+        return () => observer.disconnect();
+    }, [updateObserveKey + direction]);
```
korvin89, line 74: Why there is no `handleIntersection`?

src/components/ActionsPanel/types.ts
```diff
@@ -0,0 +1,34 @@
+import type {ButtonProps} from '../Button';
+import type {DropdownMenuItem} from '../DropdownMenu';
+import type {QAProps} from '../types';
+
+export interface ActionItem {
```
korvin89, line 5: I suggest to rename this type to `ActionsPanelItem` or `ActionsPanelActionItem`. We try to name the types associated with a component from the name of that component. This helps to avoid collisions in the general index file and have more transparent type names.

korvin89 (approved):

PR #1880, opened 2024-09-28
feat(FilePreview): add component

KirillDyachkovskiy:


src/components/FilePreview/index.ts
```diff
@@ -0,0 +1,4 @@
+export {FilePreview} from './FilePreview';
+export type {FilePreviewProps} from './FilePreview';
+export {getFileType} from './utils';
```
korvin89, line 3: Why we need it as a public interface?

vvtimofeev:
@korvin89 please come back to this PR, Kirill added 4 commits after the most recent comments, i think he implemented your suggestions.

korvin89 (approved):

PR #1917, opened 2024-10-25
feat(Select): add OnOpenChange options

angryermine:


korvin89:
@angryermine Hi! Why didn't the option `onOpenChange` work for you?

angryermine:
> @angryermine Hi! Why didn't the option `onOpenChange` work for you?

onOpenChange triggers after popup closing, in my case a want to leave popup open

korvin89:
> > @angryermine Hi! Why didn't the option `onOpenChange` work for you?
> 
> onOpenChange triggers after popup closing, in my case a want to leave popup open

@angryermine  Yep, I got it. I've edited my message

angryermine:
> I think I've figured out what you need. Do you want to close the selector when selecting an option in multiple mode?

yep

korvin89:
> > I think I've figured out what you need. Do you want to close the selector when selecting an option in multiple mode?
> 
> yep

@angryermine Got it. You can handle `open` property by yourself in this case. I suppose it could help to get what you need

angryermine:
> @angryermine Got it. You can handle `open` property by yourself in this case. I suppose it could help to get what you need

@korvin89 

this results in a blink behaviour in a single selection mode

i have two cases:
1. don't close popup in single selection 
2. close popup in multiple selection

it's the same as in Select2 (https://select2.org/dropdown#forcing-the-dropdown-to-remain-open-after-selection)

korvin89:
@angryermine I'll try to write some demo code in sandbox a little bit later

korvin89:
> 1. don't close popup in single selection
> 2. close popup in multiple selection

@angryermine  About your cases:
1. This case will be completed after [this modification ](https://github.com/gravity-ui/uikit/issues/1920)
2. This case could be completed [in this way](https://stackblitz.com/edit/vitejs-vite-em5vdi?file=src/App.tsx,package.json)

Would it work for you?

PS: you can add the second argument in this PR

angryermine:
@korvin89 

> Would it work for you?

looks like what i've been looking for

> PS: you can add the second argument in this PR

done, but i'm not sure about default value

korvin89:
> @korvin89
> 
> > Would it work for you?
> 
> looks like what i've been looking for
> 
> > PS: you can add the second argument in this PR
> 
> done, but i'm not sure about default value

@angryermine  What about default value?

amje:
Alternate proposal: #2717

PR #1950, opened 2024-11-14
fix(useBodyScrollLock): correctly restore styles

sofiushko:


korvin89:
What problem was here?

sofiushko:
> What problem was here?

`CSSStyleDeclaration.removeProperty()` not working with properties in camelCase
https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/removeProperty#property

So `restoreBodyStyles()` was broken

korvin89 (approved):

PR #1959, opened 2024-11-22
feat(Breadcrumbs): add disabledCurrent prop

beliarh:
In some cases, it would be preferable to keep the last element in the navigation active (clickable).

For example, when opening a page in a side panel.
The last breadcrumb element in this case refers to the embedded page rather than the current one.
Clicking this link should navigate to that page. But currently the link in the last item is disabled.

### DIFF
--- a/src/components/lab/Breadcrumbs/BreadcrumbItem.tsx
+++ b/src/components/lab/Breadcrumbs/BreadcrumbItem.tsx
@@ -13,6 +13,7 @@ interface BreadcrumbProps extends BreadcrumbsItemProps {
     current?: boolean;
     itemType?: 'link' | 'menu';
     disabled?: boolean;
+    disabledLink?: boolean;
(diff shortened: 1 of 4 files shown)


korvin89:
@beliarh Hi! Could you provide description for this feature? We haven't figured it out yet

beliarh:
@korvin89 Updated the description

korvin89:
@beliarh This behaviour have fixed [here](https://github.com/gravity-ui/uikit/pull/1962#event-15449880618).

[Usage example](https://github.com/gravity-ui/uikit/pull/1962/files#diff-5087b4c72a21110ac6d1385a6e0a7f47ffb69cb2df3bb7703aee7e64eaae4e65R149).

Could I close your PR?

beliarh:
> Could I close your PR?

@korvin89 Yes, sure

korvin89:
This function was added in this PR https://github.com/gravity-ui/uikit/pull/1962#event-15449880618

PR #1966, opened 2024-11-28
fix(Dialog): fixed dialog overflow, added header caption control to story

nataliadskch:
If `Dialog` header has a long word, layout breaks:
image
I added `word-break: break-word` so that it looks a bit nicer:
image

Tested on phrase «Are you sure you want to launch aaa/bbbbbbbbbb/ccccccccccccc/ddddddddddd/eeeeeeeeeee/fffffffffff?».


nataliadskch:
Ooops, I didn't mean to try to merge without review. Mixed it up with rebase button.

PR #1974, opened 2024-12-02
feat(Select): add title on mobile

iapolya:
📱 Add title to mobile Select popup

image


korvin89:
@iapolya  Hi! There is no need in such property. You can do it via `renderPopup` option, here is an example https://stackblitz.com/edit/vitejs-vite-kffgtv?file=src/App.tsx,src/main.tsx

PR #1979, opened 2024-12-05
feat!: refactor components to use Floating UI API

amje:


src/components/Toaster/ToasterComponent/ToasterPortal.tsx
```diff
@@ -31,15 +31,13 @@ export function ToasterPortal({children, className, mobile}: Props) {
...
-            return;
-        }
-
-        el.current.className = b({mobile}, className);
-    }, [className, mobile]);
-
-    return <Portal container={el.current}>{children}</Portal>;
+    return (
+        <Portal container={el.current}>
+            <div className={b({mobile}, className)} aria-live="assertive">
```
ogonkov, line 36: floating ui render it's own `div`, why we add our own div?

amje: What div are your referring to? You mean we don't need the effect above?

korvin89 (approved):

PR #1987, opened 2024-12-11
refactor(Toaster)!: use single component for Toast render

icekimi23:
PR is made to implement this RFC https://github.com/gravity-ui/rfc/issues/15

Breaking changes:

- **ToasterProvider now requires a mandatory `toaster` property**, where you must pass an instance of the `Toaster` class.
- The `Toaster` class is now responsible for managing toast state but no longer handles rendering. As a result, there is no need to call `Toaster.injectReactDOMClient(ReactDOMClient)` anymore.
- Calling `new Toaster()` now returns a new object each time. If you want to use toasts within React through the `useToaster()` hook or use the `Toaster` object directly outside React (e.g., in Redux side effects, etc.) while ensuring that toasts appear in a same screen area without overlapping, you need to make sure that the object passed to `ToasterProvider` and the one used outside React is the same instance.
- To get a single shared instance of `Toaster`, you can import it from `@gravity-ui/uikit/toaster-singleton`.
- The `toaster-singleton-react-18.js` file no longer exists, but it is unnecessary since there is no longer any specific dependency on a particular React version.
- Toaster singleton methods is no longer binded to singleton instance

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,16 +45,6 @@
                 "default": "./build/cjs/toaster-singleton.js"
             }
         },
-        "./toaster-singleton-react-18": {
-            "import": {
-                "types": "./build/esm/toaster-singleton-react-18.d.ts",
-                "default": "./build/esm/toaster-singleton-react-18.js"
-            },
-            "require": {
-                "types": "./build/cjs/toaster-singleton-react-18.d.ts",
-                "default": "./build/cjs/toaster-singleton-react-18.js"
-            }
-        },
         "./i18n": {
             "import": {
                 "types": "./build/esm/i18n/index.d.ts",


src/components/Toaster/ToasterSingleton.tsx
```diff
@@ -22,95 +14,86 @@ declare global {
...
 
-    destroy() {
-        // eslint-disable-next-line react/no-deprecated
-        ReactDOM.unmountComponentAtNode(this.rootNode);
-        document.body.removeChild(this.rootNode);
+    remove(name: string) {
+        this.toasts = removeToast(this.toasts, name);
+
+        this.notify();
     }
```
ogonkov, line 51: Where is destroy method gone?

src/components/Toaster/Provider/ToasterProvider.tsx
```diff
@@ -2,105 +2,32 @@
...
-import {removeToast} from '../utilities/removeToast';
+import type {ToasterSingleton} from '../ToasterSingleton';
+import type {InternalToastProps} from '../types';
 
 import {ToasterContext} from './ToasterContext';
 import {ToastsContext} from './ToastsContext';
 
-type Props = React.PropsWithChildren<{}>;
+type Props = React.PropsWithChildren<{
+    toaster: ToasterSingleton;
```
ogonkov, line 12: This is definitely breaking change, please change target branch to `next`

src/components/Toaster/ToasterSingleton.tsx
```diff
@@ -22,95 +14,86 @@ declare global {
...
+        };
     }
 
-    private setRootNodeClassName() {
-        this.rootNode.className = bToaster({mobile: this.mobile}, this.className);
+    private notify() {
+        for (const listener of this.listeners) {
+            listener(this.toasts);
+        }
     }
```
ogonkov, line 98: Please separate subscribe pattern from Toaster class to separate base class `EventEmitter`

icekimi23:
@ogonkov I have updated PR, check it once again, please, when you can. I removed Singleton pattern from Toaster (in separate commit), I think it is more clear and flexible to return new object every time, with this behaviour users can create multiple Providers for toasts if they want. But it would be nice to hear your opinion:)

src/components/Toaster/Toaster.tsx
```diff
@@ -0,0 +1,78 @@
+'use client';
```
ogonkov, line 1: I think this code could run on both client and server, isn't it?

ogonkov (approved):

ogonkov (changes requested):
This story seems broken https://preview.gravity-ui.com/uikit/1987/?path=/story/components-feedback-toaster--toast-playground

ogonkov (approved):

Issue #1965, opened 2024-11-28
Add autoFocus prop to Button and Select


Ruminat:
It's really inconvenient, that these components don't have that prop
In `Button` you can set it via `extraProps` (which is better, but still can be improved by adding the property directly to the component)
In `Select` you don't even have the `extraProps`, the only options you're left with are
- focusing with `ref` on `useEffect`
- or modifying the Select button with `renderControl`

amje:
In the next major update we will add ability to pass all HTML props to Button, Link and Select

PR #1988, opened 2024-12-12
chore(Button): add autoFocus property

GermanVor:
https://github.com/gravity-ui/uikit/issues/1965


amje:
@GermanVor Hi, thanks for contribution, but this change will be not compatible with the changes we did for upcoming major update, see [comment](https://github.com/gravity-ui/uikit/issues/1965#issuecomment-2549009335)

PR #1995, opened 2024-12-16
feat(modal)!: use floating ui in modal

oynikishin:


src/components/Dialog/DialogFooter/DialogFooter.tsx
```diff
@@ -55,128 +53,103 @@ function getButtonView(preset: ButtonPreset): ButtonView {
...
-                    <Popup
-                        open={showError}
-                        anchorRef={this.errorTooltipRef}
-                        placement={['bottom', 'top']}
-                        disablePortal
-                        hasArrow
-                    >
-                        <div className={b('error')}>{errorText}</div>
-                    </Popup>
+                    <div className={b('error')}>{errorText}</div>
```
ogonkov, line 134: Is it really should be part of base component? May be we should consider some render prop?

src/components/Modal/Modal.tsx
```diff
@@ -79,85 +139,159 @@ export function Modal({
...
+                        >
+                            <div
+                                className={b(
+                                    'content',
+                                    {'has-scroll': contentOverflow === 'auto'},
+                                    contentClassName,
+                                )}
+                                ref={handleFloatingRef}
+                                aria-label={ariaLabel}
+                                aria-labelledby={ariaLabelledBy}
```
ogonkov, line 283: Does this node have role? Without role labels doesnt make sense

Issue #1877, opened 2024-09-26
Support `React.ReactNode` type for `TextInput['label']` property


korvin89:
https://github.com/gravity-ui/uikit/blob/main/src/components/controls/TextInput/TextInput.tsx#L40

PR #1997, opened 2024-12-17
feat(TextInput): change type prop label

u4aew:
### Description
Added type React.ReactNode for label in the TextInput component.

Issue:  #1877


amje:
@u4aew Hi, thanks for contribution. But, unfortunately, we've changed our mind regarding this feature. The inital problem we had could be resolved another way, not expanding the API. For now we don't want to have `ReactNode` there, because it could be a potential lost of control in this place, we want there to be a simple `string`

PR #2002, opened 2024-12-19
feat(tabs): add new tabs components

sofiushko:


ogonkov:
All this changes for tabs?

korvin89:
@sofiushko Hey! Could you rebase your branch?

Arucard89 (approved):

sofiushko:
> @sofiushko Hey! Could you rebase your branch?

Done
Is visual test `[DefinitionList › smoke [direction: horizontal]]` stable?

sofiushko:
@amje, @korvin89 
Hey, rebased this branch again

src/components/tabs/TabList.scss
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        &:focus-visible {
+            #{$block}__item-content {
+                outline: 2px solid var(--g-color-line-focus);
+                outline-offset: -2px;
+            }
+        }
+
+        &-title {
```
korvin89, line 93: There is one `&-title` above, do we actually need to repeat it here?

src/components/tabs/TabPanel.tsx
```diff
@@ -0,0 +1,32 @@
...
+    children: React.ReactNode;
+}
+
+export const TabPanel = (props: TabPanelProps) => {
+    const {children, value, qa, id} = props;
+    const {activeTabId} = React.useContext(TabContext);
+
+    return (
+        <div
+            {...filterDOMProps(props, {labelable: true})}
```
korvin89, line 23: Is it assumed that users always should set `aria-labelledby` by themselves?

korvin89 (commented):
There is a [w3c design pattern for tabs](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/). For now there are some differences between this pattern and current code behaviour:
- `Enter` key doesn't activate tab
- Arrows don't move focus within tabs
- I'm not quite sure about moving focus between tabs by `Tab` key pressing

sofiushko:
> There is a [w3c design pattern for tabs](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/). For now there are some differences between this pattern and current code behaviour:

I relied on https://github.com/gravity-ui/rfc/issues/16
There is a fundamental difference between these components behaviour. 

> * `Enter` key doesn't activate tab
It`s ok, we can do this

> * Arrows don't move focus within tabs
it`s not ok. If we want this behaviour, we need to make tabs component uncontrolled. It goes against the issue. Maybe we should change the interface to make it possible?

> * I'm not quite sure about moving focus between tabs by `Tab` key pressing
I repeated the current component in this moment

amje:
@sofiushko I agree that focus behaviour should follow w3c pattern (arrow keys navigation instead of tabs). But I don't understand why we should change the API to make it possible

src/components/tabs/TabList.tsx
```diff
@@ -0,0 +1,46 @@
...
+    ({size = 'm', value, children, className, onUpdate, qa}, ref) => {
+        const activeTabId = React.useContext(TabContext).activeTabId || value;
+
+        const tabInnerContextValue = React.useMemo(
+            () => ({onUpdate, activeTabId}),
+            [onUpdate, activeTabId],
+        );
+
+        return (
+            <div className={bTabList({size}, className)} data-qa={qa} ref={ref}>
```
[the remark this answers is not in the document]
sofiushko: For a possibility to render adaptive tabs in future we need some wrapper

sofiushko:
> @sofiushko I agree that focus behaviour should follow w3c pattern (arrow keys navigation instead of tabs). But I don't understand why we should change the API to make it possible

So sorry
I think you want to activate the newly tab on Arrow Click, according to
 `Left Arrow: moves focus to the previous tab. If focus is on the first tab, moves focus to the last tab. Optionally, activates the newly focused tab`
If only focus - it possible, i will think about realization =)

sofiushko:
@amje, @korvin89 

Add some behaviour from [w3c design pattern for tabs](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/)
- focus the active tab item on focus by `tab` 
- next `tab` moves for next focusable elem
- focus on arrows
- activate tab on `space` and `enter`
- focus on home/end

korvin89 (approved):

PR #2004, opened 2024-12-21
Add support NextLink and other custom link components

jadnast:
Added support for using custom link components (e.g., NextLink) in the Button component via the new `hrefComponent` prop. If `hrefComponent` is not specified, it defaults to `<a>`.

### Details:
- Introduced a new prop `hrefComponent` in `ButtonProps`.
- If `hrefComponent` is provided, it will render the specified component for links instead of  `<a>`.
- Maintains backward compatibility by defaulting to `<a>` when `hrefComponent` is not set.
- Ensures flexibility for projects using custom routing libraries like Next.js.

### Related Tickets & Documents
- Issues: #2007 

I hereby agree to the terms of the CLA available at: https://yandex.ru/legal/cla/?lang=en

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -50,6 +50,7 @@ export interface ButtonProps extends DOMProps, QAProps {
     type?: 'button' | 'submit' | 'reset';
     component?: React.ElementType;
     href?: string;
+    hrefComponent?: React.ElementType;
     target?: string;
     rel?: string;
     extraProps?:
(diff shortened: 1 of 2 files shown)


amje:
@jadnast Hi, we've already attempted to support this in #1594 and we'll add it to the `next` branch (that will be 7 release) soon. You can be able to use only `component` prop, without need of extra props

amje:
@jadnast Hi, could you try this feature in v7 release?

jadnast:
@amje Of course

jadnast:
@amje Everything is great, thank you!

Issue #2003, opened 2024-12-20
Migrate to new JSX Transform


amje:
1. Change `jsx` option in TSConfig from `"react"` to `"react-jsx"`
2. Remove not used React imports from files

ogonkov:
Also, codemods available

PR #2006, opened 2024-12-21
feat: migrate to new JSX transform

u4aew:
### Description 
Added `react-jsx` to **tsconfig.json** | **tsconfig.test.json**, updated rules in **.eslintrc**, updated presets in **.babelrc**  and removed `React` imports.

[docs](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)

issue: #2003


.eslintrc
```diff
@@ -32,7 +28,8 @@
         "import/consistent-type-specifier-style": ["error", "prefer-top-level"],
         "@typescript-eslint/prefer-ts-expect-error": "error",
         "@typescript-eslint/consistent-type-imports": ["error", {"prefer":  "type-imports", "fixStyle": "separate-type-imports"}],
-        "complexity": "off"
+        "complexity": "off",
+        "react/react-in-jsx-scope": "off"
```
ogonkov, line 32: Guide also suggest `react/jsx-uses-react`

.eslintrc
```diff
@@ -14,10 +14,6 @@
         ],
         "no-restricted-syntax": [
             "error",
-            {
-                "selector": "ImportDeclaration[source.value='react'] :matches(ImportNamespaceSpecifier, ImportSpecifier)",
-                "message": "Please use import React from 'react' instead."
-            },
```
amje, line 20: Using `React` namespace is our codestyle. So `import * as React from 'react'` should be the only option.
To accomplish this, change the rule to:
```json
{
    "selector": "ImportDeclaration[source.value='react'] :matches(ImportDefaultSpecifier, ImportSpecifier)",
    "message": "Please use `import * as React from 'react'` instead."
}
```

.babelrc
```diff
@@ -1,16 +1,21 @@
...
-  "presets": [
-    [
-      "@babel/preset-env",
-      {
-        "targets": {
-          "chrome": 100
-        }
-      }
+    "sourceType": "unambiguous",
+    "presets": [
```
amje, line 3: Please keep current preset order

.eslintrc
amje: We can simply extend `plugin:react/jsx-runtime` which does the job. No need to off the rules here.

amje (changes requested):

amje (approved):

amje:
@u4aew Thanks for contribution!

PR #2008, opened 2024-12-21
feat: add select autofocus props

iapolya:
Optional autofocus. This is useful for mobiles when you don't want to open the keyboard.

### DIFF
--- a/src/components/Select/README.md
+++ b/src/components/Select/README.md
@@ -284,6 +284,8 @@ LANDING_BLOCK-->
 
 To enable filter section, use the `filterable` property. Its default value is `false`.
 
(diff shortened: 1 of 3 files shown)


korvin89:
@iapolya 
Could you describe your case? Why you need focused control on mobile device?

iapolya:
@korvin89, instead, I don't want a focus. It's currently on by default.

There are cases where I use a select with a filter, but I don't want to open the keyboard right away.

korvin89:
@iapolya 
Current behaviour is the behaviour by design. You can achieve you goal in this way https://stackblitz.com/edit/vitejs-vite-kffgtv?file=src/App.tsx,src/main.tsx

PR #2009, opened 2024-12-21
feat(Modal): add CSS API for width and height

iapolya:
Modal window state stretched to full page, no mardin, no border radius.  Useful on mobile.

### DIFF
--- a/src/components/Modal/Modal.scss
+++ b/src/components/Modal/Modal.scss
@@ -2,8 +2,10 @@
 
 $block: '.#{variables.$ns}modal';
 $modal-default-border-radius: 5px;
-$modal-default-height: auto;
-$modal-default-width: auto;
+$modal-content-height: auto;
+$modal-content-width: auto;
+$modal-wrapper-height: auto;
+$modal-wrapper-width: auto;
 
 #{$block} {
     display: block;
@@ -32,7 +34,7 @@ $modal-default-width: auto;
     &__content-aligner {
         min-width: 100%;
         min-height: 100%;
-        height: var(--g-modal-height, #{$modal-default-height});
+        height: var(--g-modal-wrapper-height, #{$modal-wrapper-height});
 
         display: inline-flex;
         align-items: center;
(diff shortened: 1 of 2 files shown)


amje:
@iapolya Hi, Modal is not supposed to be open in fullscreen by design. It has no "close" element except its overlay. So it won't be possible to close the Modal on mobile device in such scenario. There is a dedicated component for mobile cases - `Sheet`, which should be used instead

src/components/Modal/Modal.scss
```diff
@@ -20,19 +22,25 @@ $modal-default-border-radius: 5px;
...
+        width: var(--g-modal-width, #{$modal-default-width});
+        height: var(--g-modal-height, #{$modal-default-height});
+
         margin: var(--g-modal-margin, #{variables.$modal-default-margin});
         border-radius: var(--g-modal-border-radius, #{$modal-default-border-radius});
         overflow-x: hidden;
     }
 
     &__content {
+        height: var(--g-modal-height, #{$modal-default-height});
```
amje, line 42: Let's apply these styles only for `&__content`:
```
width: var(--g-modal-width, #{$modal-default-width});
height: var(--g-modal-height, #{$modal-default-height});
```

Applying it to all elements could result in unexpected behaviour.

amje (changes requested):

src/components/Modal/Modal.scss
iapolya: Yes, valid point. This works for the case when I want to use vh, but it's not sufficient when using %.

amje: content and wrapper represent inner Modal elements, their only purpose it to center the modal, they are not for public tweaking. For the end user they should not be exposed. Why `--g-modal-width` and `--g-modal-height` is not enough?

iapolya: In the project, we [used 100% for the mobile modal](https://developer.chrome.com/blog/url-bar-resizing?hl=ru), and the styles were like this: 

```
.g-modal__content-aligner {
    height: 100%;
}

.g-modal__content-wrapper,  .g-modal__content {
    height: 100%;
    width: 100%;
}
```

We will switch to [dvh](https://caniuse.com/?search=dvh) in the project. It works correctly on mobile and does not require nested styles. I simplified CSS API

amje (approved):

PR #2018, opened 2024-12-25
feat: require React version 16.14.0 or higher

amje:


korvin89 (approved):

ogonkov (commented):
I guess package-lock.json should be updated too

I think running npm ci should do the trick

Issue #1408, opened 2024-03-06
Remove deprecated `switcher` prop from `DropdownMenu`


amje:

amje:
DropdownMenu will be removed in v8

PR #2019, opened 2024-12-25
feat(DropdownMenu): remove deprecated prop switcher

u4aew:
### Description
The deprecated `switcher` property has been removed from the `DropdownMenu` component.

issue #1408


amje:
@u4aew We're currently working on complete refactor of this component. So this change is irrelevant

PR #2031, opened 2024-12-28
feat(Popover)!: replace legacy Popover with the new one

amje:
### Summary of changes

- Created `legacy` package entrypoint for deprecated and obsolete components, which can't be removed for a while.
- Replaced Popover component with the component from Lab
- Moved old Popover component to the legacy entrypoint


src/components/Popover/Popover.tsx
```diff
@@ -1,249 +1,130 @@
...
+        | ((props: Record<string, unknown>, ref: React.Ref<HTMLElement>) => React.ReactElement)
+        | React.ReactElement;
+    open?: boolean;
+    onOpenChange?: (open: boolean) => void;
+    disabled?: boolean;
+    content?: React.ReactNode;
+    trigger?: 'click';
+    openDelay?: number;
+    closeDelay?: number;
+    enableSafePolygon?: boolean;
```
korvin89, line 44: Could you provide description for this option?

korvin89 (approved):

src/components/Popover/Popover.tsx
amje: There is a note about it in Properties table in readme

PR #2037, opened 2025-01-13
fix(NumberInput): disable rounding decimal to integer without passed step

DaryaLari:


amje:
@DaryaLari Could you please make the same changes to the `next` branch? The component is moved from the lab there.

amje (approved):

PR #2041, opened 2025-01-14
fix(NumberInput): disable rounding decimal to integer without passed step

DaryaLari:


DaryaLari:
Pr to next instead of/addition to https://github.com/gravity-ui/uikit/pull/2037
@amje

amje (approved):

PR #2051, opened 2025-01-17
feat(HelpMark)!: use default popover component

amje:


src/components/HelpMark/HelpMark.tsx
```diff
@@ -13,33 +13,30 @@ import './HelpMark.scss';
...
-                    type="button"
-                    {...buttonProps}
-                    className={b('button', buttonProps.className)}
-                >
-                    <Icon data={CircleQuestion} size={ICON_SIZE} />
-                </button>
-            )}
+        <Popover
+            content={<div className={b('popover')}>{children}</div>}
+            hasArrow
```
Raubzeug, line 28: what do you think if we add default for `delayClosing` to keep previous behaviour?

Raubzeug (approved):

src/components/HelpMark/HelpMark.tsx
Raubzeug: discussed in pm, default will be set in `Popover`.

Raubzeug (approved):

src/components/HelpMark/HelpMark.tsx
amje: New Popover has the default delay, not 300ms but 250ms

PR #2052, opened 2025-01-20
feat(Slider): remove deprecated fields

Arucard89:
**Ticket:** UXRFC-503

### Slider

- removed deprecated field `availableValues`: set `step` to null and use `marks`, `min` and `max` props to specify available values to select instead
- removed deprecated field `marksCount`: use `marks` instead
- removed deprecated field `hasTooltip`: use `tooltipDisplay` instead
- removed all this deprecated fields from parameters of `prepareSliderInnerState`


Arucard89:
@korvin89 , I wrote the description. Could you check, please, is it enough?

korvin89:
> @korvin89 , I wrote the description. Could you check, please, is it enough?

@Arucard89  Yep. Check failed tests as well please

korvin89 (approved):

PR #2064, opened 2025-01-24
fix(Alert): allow for the scrollable content to have scroll

ArturAbdullin:


src/components/Alert/__stories__/Alert.stories.tsx
```diff
@@ -143,3 +143,25 @@ export const Align: Story = {
         ...Default.args,
     },
 };
+
+export const WithScrollableMessage: Story = {
```
ArturAbdullin, line 147: This story is just to show that it works. To be deleted.

src/components/Alert/Alert.tsx
```diff
@@ -44,9 +44,28 @@ export const Alert = (props: AlertProps) => {
...
                         props.icon // ability to pass `null` as `icon` prop value
                     )}
-                    <Flex direction={layout === 'vertical' ? 'column' : 'row'} gap="5" grow>
-                        <Flex gap="2" grow className={bAlert('text-content')}>
-                            <Flex direction="column" gap="1" grow justifyContent={align}>
+                    <Flex
+                        direction={layout === 'vertical' ? 'column' : 'row'}
+                        gap="5"
+                        grow
+                        shrink={1}
```
amje, line 51: `flex-shrink` defaults to 1. No need to specify it

src/components/Alert/__stories__/Alert.stories.tsx
```diff
@@ -143,3 +143,25 @@ export const Align: Story = {
...
+    render: ({message, ...restArgs}) => (
+        <Showcase>
+            <ShowcaseItem title={''}>
+                <Alert
+                    {...restArgs}
+                    message={
+                        <div style={{overflowX: 'auto'}}>
```
amje, line 154: Add tabIndex={0} to fix a11y warning

amje (changes requested):

ArturAbdullin:
`min-width={0}` for some reason breaks the Alert's title alignment with the icon.
Found another solution to make the scrollable content - the content of the message should be wrapped with the `display: grid` container.

PR #2065, opened 2025-01-24
feat: add fixed position on body scroll lock

iapolya:
Before:

https://github.com/user-attachments/assets/bf03cd9d-2a2d-4c2f-9cf4-070c1aeac238

After:

https://github.com/user-attachments/assets/c728a01c-1394-4ca5-af24-78aca3d09ca0


amje:
@iapolya Hi, in upcoming v7 release, we have removed this hook. We're fully rely on Floading UI [implementation](https://floating-ui.com/docs/FloatingOverlay#lockscroll) now.

PR #2072, opened 2025-01-27
fix(Button): additional to #2070

ma-efremoff:
Unfortunately typescript cannot determine which type is used when we
don't provide `href` property. As result it brings some compilation
errors with `ref` and `component` fields.

So we need to bring back the conditional inferring of props type.
It was a mistake to think we can avoid it.

Additionaly `Exclude<ElementType, 'a' | 'button'>` is extracted
as separate type `ButtonCustomElementType`

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -50,7 +50,6 @@ export interface ButtonButtonProps
     extends ButtonCommonProps,
         Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, 'disabled'> {
     component?: never;
-    ref?: React.Ref<HTMLButtonElement>;
     href?: never;
     /**
      * @deprecated Use additional props at the root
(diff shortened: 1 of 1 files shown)


src/components/Button/Button.tsx
```diff
@@ -142,7 +141,11 @@ const _Button = React.forwardRef(function Button<T extends React.ElementType>(
         'data-qa': qa,
     };
 
-    if (props.component) {
+    function isButtonComoponentProps(p: ButtonProps<T>): p is ButtonComponentProps<T> {
```
amje, line 144: Let's move this function out of render

amje (changes requested):

src/components/Button/Button.tsx
ma-efremoff: done

ma-efremoff:
```
1 failed
    [chromium] › ../playwright/core/index.ts:19:5 › ClipboardButton › smoke  @ClipboardButton @smoke 
```
It looks like all the difference in the popup view:
4e2272983f064452b43829aaaa408dac1b1ec5b8

It should be affected by changes from another PR.

amje (approved):

PR #2076, opened 2025-01-29
feat(Modal): height animation

roberthovsepyan:
UXRFC-608


src/components/Modal/Modal.tsx
```diff
@@ -83,6 +83,50 @@ export function Modal({
...
+        // Set previous height first for the transition to work, because it doesn't work with 'auto'
+        contentRef.current.style.height = `${previousHeight.current}px`;
+        isTransitioningHeight.current = true;
+
+        requestAnimationFrame(() => {
+            if (contentRef.current) {
+                contentRef.current.style.height = `${contentHeight}px`;
+                previousHeight.current = contentHeight;
+            }
+        });
```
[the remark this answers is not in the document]
roberthovsepyan: We have to first set the height to the current height and then set the new height (which happens in this callback)
alternatively setTimeout can be used to achieve the same result

src/components/Modal/Modal.tsx
```diff
@@ -96,8 +96,8 @@ export interface ModalProps extends DOMProps, AriaLabelingProps, QAProps {
     /** Callback called when `Popup` is closed and "out" transition is completed */
     onTransitionOutComplete?: () => void;
     contentOverflow?: 'visible' | 'auto';
-
     floatingRef?: React.RefObject<HTMLDivElement>;
+    disableHeightTransition?: boolean;
```
amje, line 100: What about adding the same prop for `Dialog` too?

src/components/Modal/Modal.tsx
```diff
@@ -182,6 +183,11 @@ export function Modal({
     const {isMounted, status} = useTransitionStatus(context, {duration: TRANSITION_DURATION});
     const previousStatus = usePrevious(status);
 
+    const animateHeightProps = useAnimateHeight({
+        ref: !disableHeightTransition && initialFocusRef.current ? initialFocusRef : undefined,
```
amje, line 187: Let's remove `initialFocusRef` variable and use `refs.floating` instead in both places. Now it's a bit confusing to see refs for focus in animation hook

src/hooks/private/useAnimateHeight/useAnimateHeight.ts
```diff
@@ -0,0 +1,77 @@
...
+    }, [enabled]);
+
+    const handleContentTransitionEnd = React.useCallback(
+        (event: React.TransitionEvent<HTMLElement>) => {
+            if (event.propertyName !== 'height') {
+                return;
+            }
+
+            // ResizeObserver final resize event fires before this, so we have to delay with timeout
+            setTimeout(() => {
```
amje, line 33: RAF callback doesn't work in this case?

amje (changes requested):

src/components/Modal/Modal.tsx
roberthovsepyan: fixed

src/components/Modal/Modal.tsx
roberthovsepyan: fixed

src/hooks/private/useAnimateHeight/useAnimateHeight.ts
```diff
@@ -0,0 +1,77 @@
...
+    }, [ref, enabled]);
+
+    useResizeObserver({ref, onResize: handleResize});
+
+    const result = React.useMemo<UseAnimateHeightResult>(
+        () => ({onTransitionEnd: handleContentTransitionEnd}),
+        [handleContentTransitionEnd],
+    );
+
+    return result;
```
ValeraS, line 76: It seems we can make this hook self-contain

```suggestion
    const handleResize = React.useCallback(() => {
        const node = ref?.current;
        if (!node || isTransitioningHeight.current || !enabled) {
            return;
        }

        const contentHeight = node.clientHeight;
        if (!previousHeight.current || previousHeight.current === contentHeight) {
            previousHeight.current = contentHeight;
            return;
        }

        // Set previous height first for the transition to work, because it doesn't work with 'auto'
        node.style.height = `${previousHeight.current}px`;
        isTransitioningHeight.current = true;
        const overflowY = node.style.overflowY;
        node.style.overflowY = 'clip';

        const handleTransitionEnd = (event: TransitionEvent) => {
            if (event.propertyName !== 'height') {
                return;
            }

            node.removeEventListener('transitionend', handleTransitionEnd);

            // ResizeObserver final resize event fires before this, so we have to delay with timeout
            setTimeout(() => {
                node.style.height = 'auto';
                node.style.overflowY = overflowY;
                isTransitioningHeight.current = false;
            }, 0);
        };

        node.addEventListener('transitionend', handleTransitionEnd);

        requestAnimationFrame(() => {
            node.style.height = `${contentHeight}px`;
            previousHeight.current = contentHeight;
        });
    }, [ref, enabled]);

    useResizeObserver({ref: enabled ? ref : undefined, onResize: handleResize});
```

src/components/Modal/Modal.tsx
```diff
@@ -182,6 +178,11 @@ export function Modal({
     const {isMounted, status} = useTransitionStatus(context, {duration: TRANSITION_DURATION});
     const previousStatus = usePrevious(status);
 
+    const animateHeightProps = useAnimateHeight({
+        ref: !disableHeightTransition && refs.floating.current ? refs.floating : undefined,
+        enabled: open,
+    });
```
ValeraS, line 184: With proposed changes this will be
```suggestion
    useAnimateHeight({
        ref: refs.floating,
        enabled: status === 'open' && !disableHeightTransition,
    });
```

ValeraS (changes requested):

src/hooks/private/useAnimateHeight/useAnimateHeight.ts
roberthovsepyan: fixed it
thanks!

src/components/Dialog/__stories__/DialogShowcase.tsx
```diff
@@ -123,6 +124,90 @@ function OtherDialog() {
...
+            }, 2000);
+
+            setTimeout(() => {
+                setIsSecondDynamicPartOpen(true);
+            }, 4000);
+        } else {
+            setIsFirstDynamicPartOpen(false);
+            setIsSecondDynamicPartOpen(false);
+        }
+    }, [open]);
```
ValeraS, line 149: Here and in the Modal story:
```suggestion
    React.useEffect(() => {
        const timers: number[] = [];
        if (props.open) {
            timers[0] = window.setTimeout(() => {
                setIsFirstDynamicPartOpen(true);
            }, 2000);

            timers[1] = window.setTimeout(() => {
                setIsSecondDynamicPartOpen(true);
            }, 4000);
        } else {
            setIsFirstDynamicPartOpen(false);
            setIsSecondDynamicPartOpen(false);
        }
        return () => {
            timers.forEach((t) => {
                window.clearTimeout(t);
            });
        };
    }, [props.open]);
```

ValeraS (changes requested):

src/components/Dialog/__stories__/DialogShowcase.tsx
roberthovsepyan: done

ValeraS (approved):

amje (approved):

PR #2082, opened 2025-01-31
fix(Select): keyboard behavior on mobile

iapolya:
There are cases when the select with an open keyboard behaves inconveniently for the user

1. Auto-closing of Select when the keyboard is closed. Reproduced in Safari and Yandex Browser on iOS

Click on "Готово" button

https://github.com/user-attachments/assets/7413fd16-2cd4-4ecc-ac3a-823058a57693

2. The keyboard opens after selecting item. This is inconvenient because it requires the user to close the keyboard if they want to select an item from the list located nearby

https://github.com/user-attachments/assets/d41aed2a-bc9a-4834-b827-95c37eb6b556

**Questions:**

1. For which cases might we need _useFocusWithin_ in a mobile select?
2. Сan I remove `filterRef.current?.focus() `on mobile?

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -164,7 +164,10 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function 
 
             if (multiple) {
                 const activeItemIndex = listRef?.current?.getActiveItem();
-                filterRef.current?.focus();
+
+                if (!mobile) {
+                    filterRef.current?.focus();
+                }
 
                 if (typeof activeItemIndex === 'number') {
                     // prevent item deactivation in case of multiple selection
@@ -178,7 +181,7 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function 
 
             handleSelection(option);
         },
-        [handleSelection, multiple],
+        [handleSelection, mobile, multiple],
     );
 
     const handleControlKeyDown = React.useCallback(
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -244,6 +244,7 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function 
             },
             [handleClose, onBlur],
         ),
+        isDisabled: mobile,
     });
 
     const uniqId = useUniqId();
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -240,11 +240,13 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function 
         onBlurWithin: React.useCallback(
             (e: React.FocusEvent) => {
                 onBlur?.(e);
-                handleClose();
+
+                if (!mobile) {
+                    handleClose();
+                }
             },
-            [handleClose, onBlur],
+            [handleClose, mobile, onBlur],
         ),
-        isDisabled: mobile,
     });
 
     const uniqId = useUniqId();
--- a/src/components/Sheet/SheetContent.tsx
+++ b/src/components/Sheet/SheetContent.tsx
@@ -2,6 +2,7 @@
 
 import * as React from 'react';
 
+import {SelectQa} from '../Select';
 import {Platform, withMobile} from '../mobile';
 import type {History, Location, MobileContextProps} from '../mobile';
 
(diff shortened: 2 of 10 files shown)


iapolya:
@korvin89 ping

src/hooks/useFocusWithin/useFocusWithin.ts
```diff
@@ -186,15 +190,15 @@ function useFocusEvents({
         (event: React.FocusEvent) => {
             if (
                 document.activeElement !== event.target &&
-                (event.relatedTarget === null ||
+                ((!mobile && event.relatedTarget === null) ||
```
ValeraS, line 193: It might be better to disable `useFocusWithin` in `Select` on mobile devices.

ValeraS (changes requested):

src/hooks/useFocusWithin/useFocusWithin.ts
iapolya: @ValeraS, when I disable the hook, the tests for closing the select fail because onClose is not being called. It looks like the hook is needed on mobile

Screenshot 2025-02-05 at 16 03 32

ValeraS: We can replace click on body with `await userEvent.click(screen.getByTestId(SelectQa.SHEET));` in tests

iapolya: Yes, but userEvent.click(...) doesn't trigger [onTransitionEnd](https://github.com/gravity-ui/uikit/blob/feat/select-close-on-mobile/src/components/Sheet/SheetContent.tsx#L148) we need to close Sheet. The test worked because (with useFocusWithin) click triggered blur. I'll research this moment

src/components/Sheet/SheetContent.tsx
```diff
@@ -144,9 +145,10 @@ class SheetContent extends React.Component<SheetContentInnerProps, SheetContentS
                 <div
                     ref={this.veilRef}
                     className={sheetBlock('veil', veilTransitionMod)}
-                    onClick={isAnimating ? undefined : this.onVeilClick}
+                    onClick={this.onVeilClick}
                     onTransitionEnd={this.onVeilTransitionEnd}
                     role="presentation"
+                    data-qa={SelectQa.SHEET_VEIL}
```
mournfulCoroner, line 151: Why is the `Select` qa used in a separate `Sheet` component?

iapolya: By analogy with [this](https://github.com/gravity-ui/uikit/blob/feat/select-close-on-mobile/src/components/Select/constants.ts#L37). But I agree with the remark, I will move it to `Sheet`

src/hooks/useFocusWithin/useFocusWithin.ts
ValeraS: why are these changes still here?

iapolya: Sorry, missed it after rebase

src/components/Select/Select.tsx
```diff
@@ -241,6 +244,7 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function
             },
             [handleClose, onBlur],
         ),
+        isDisabled: mobile,
```
korvin89, line 247: Lets remove this line and don't call `handleClose` in `onBlurWithin` in case of `mobile` device

src/components/Select/Select.tsx
```diff
@@ -305,7 +308,8 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function
         <div
             ref={controlWrapRef}
             className={selectBlock(mods, className)}
-            {...focusWithinProps}
+            onFocus={mobile ? onFocus : focusWithinProps.onFocus}
+            onBlur={mobile ? onBlur : focusWithinProps.onBlur}
```
ValeraS, line 312: How is this different from disabling the hook? @korvin89 suggested the following
```js
    const {focusWithinProps} = useFocusWithin({
        onFocusWithin: onFocus,
        onBlurWithin: React.useCallback(
            (e: React.FocusEvent) => {
                onBlur?.(e);
                if (!mobile) {
                    // close on blur only on desktop
                    handleClose();
                }
            },
            [handleClose, onBlur],
        ),
    });
```

ValeraS (changes requested):

ValeraS (approved):

mournfulCoroner (approved):

PR #2083, opened 2025-02-03
feat(TabList): use hooks for logic

amje:
- Move all logic into hooks
- Handle aria attributes automatically
- Support RTL
- Use `button` as element for Tab
- Support alternate activation mode by focus


src/components/tabs/hooks/useTabPanel.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    const currentValue = tabContextProps.value;
+    const parentId = tabContextProps.id;
+    const tabId = `${parentId}:t:${tabPanelProps.value}`;
+    const panelId = `${parentId}:p:${tabPanelProps.value}`;
+
+    const isSelected = currentValue === tabPanelProps.value;
+
+    return {
+        ...filterDOMProps(tabPanelProps, {labelable: true}),
```
sofiushko, line 19: We will rewrite `aria-labelledby` below?

src/components/tabs/hooks/useTabList.ts
```diff
@@ -0,0 +1,127 @@
...
+            case KeyCode.END: {
+                event.preventDefault();
+                activateOnFocus(focusFurthestTab(event, false));
+                break;
+            }
+        }
+    };
+
+    return {
+        ...filterDOMProps(tabListProps, {labelable: true}),
```
sofiushko, line 117: We will rewrite `aria-labelledby` below?

sofiushko (commented):
is it ok if user will add `TabPovider` without panels?
There will be not valid aria-attrs

amje:
> is it ok if user will add `TabPovider` without panels? There will be not valid aria-attrs

We can't force it anyway. So, yes, there will be invalid attributes

src/components/tabs/hooks/useTabPanel.ts
amje: I think we should change the order here

amje: Seems we must always rewrite this prop to ensure correct a11y here

sofiushko: In this case users should not be able to set it

amje: changed

korvin89 (approved):

PR #2088, opened 2025-02-05
feat(Sheet): add maxHeight prop

MrAndrushik:
The `Sheet` component does not have the ability to occupy the maximum height with a small amount of content. 

Added the `maxHeight` prop to the component


mournfulCoroner (changes requested):

mournfulCoroner:
Thanks for the pull request! I would like to note that this behavior can also be achieved using the `contentClassName` prop and specifying `min-height: 90vh`

mournfulCoroner (approved):

PR #2101, opened 2025-02-08
fix(Dialog): fixed dialog overflow, added header caption control to story

nataliadskch:
If `Dialog` header has a long word, layout breaks:
image
I added `word-break: break-word` so that it looks a bit nicer:
image

Tested on phrase «Are you sure you want to launch aaa/bbbbbbbbbb/ccccccccccccc/ddddddddddd/eeeeeeeeeee/fffffffffff?».


korvin89:
@nataliadskch Hey! We believe that current proposal may worse the current situation, as it seems incorrect to break words in such an important element as the dialog title.
`Dialog` already knows how to hyphenate words, but only if they have a normal width. For your case, you can use props `capture`, which can work with React node.

PR #2112, opened 2025-02-13
feat: accept text as function in clipboard button

astandrik:
In this task we have frequently updated data that can be copied to clipboard

https://github.com/ydb-platform/ydb-embedded-ui/issues/1837

we want to process it only on clipboard button click


src/components/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -10,7 +10,7 @@ const INITIAL_STATUS: CopyToClipboardStatus = 'pending';
 export function CopyToClipboard(props: CopyToClipboardProps) {
     const {children, text, timeout, onCopy} = props;
 
-    const textRef = React.useRef(text);
+    const textRef = React.useRef('');
```
astandrik, line 13: didnt actually understand why do we need text here
looks like it not used anywhere outside of onClickWithCopy

Raubzeug (approved):

Raubzeug (approved):

astandrik:
@SeqviriouM ping

SeqviriouM (approved):

PR #2126, opened 2025-02-20
fix: remove unused `react-dom` from `peerDependencies`

ogonkov:
`react-dom` was used in Toaster


amje:
@ogonkov  I think keeping it has no harm. In case if we need it again later we won't be locked by major change adding a peer

ogonkov:
@amje in projects without react-dom this cause error on install, because react-dom@19 resolved by npm

I dont see point to keep in peers dependency that we didnt use in code

amje:
> I dont see point to keep in peers dependency that we didnt use in code

I've noted my point above

PR #2151, opened 2025-03-07
fix(Progress)!: component margin removed

### LINKED ISSUE #2150: [Progress] component default margin
### Description

I got unexpected alignment when using component Progress cause it has its default [margin](https://github.com/gravity-ui/uikit/blob/4e7b797c88565dede7e3c22a7ab5cf69fa404867/src/components/Progress/Progress.scss#L26).

### Suggested solution

Does it really need it or we can remove it?

### Alternative

_No response_

### Additional context

_No response_

### Validations

chelentos:
Fixes #2150

Should be merged in major, because can cause changes in layout where it's used.

### DIFF
--- a/src/components/Progress/Progress.scss
+++ b/src/components/Progress/Progress.scss
@@ -23,7 +23,6 @@ $progress-xs-height: 4px;
 
     position: relative;
 
-    margin: 0 auto;
     overflow: hidden;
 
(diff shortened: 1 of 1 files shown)


Lunory:
Switch target branch to `next` please if PR with breaking changes

PR #2153, opened 2025-03-07
fix(HelpMark): icon alignment

ogonkov:
Because of inline document flow, inline icon inside inline-block button adds more vertical space than expected

Screenshot 2025-03-07 at 18 15 12


Raubzeug (changes requested):
Please update visual tests!

Raubzeug (approved):

PR #2157, opened 2025-03-12
feat(FilePreview): add compact view

kseniya57:
Add the compact file preview without text for the cases like table or gallery


amje (approved):

mournfulCoroner (changes requested):
We need to work out the appearance options from the design side. Already requested

kseniya57:
> We need to work out the appearance options from the design side. Already requested

Why we can't do the changes after the design is ready?

mournfulCoroner:
> > We need to work out the appearance options from the design side. Already requested
> 
> Why we can't do the changes after the design is ready?

If you're talking about the design for the gallery, it wasn't the design for the File Preview component. The changes have not been agreed upon for him. 
If you just add the "compact" prop now, what will you have to do if you need any more versions of the component's appearance? Then in the future you may have to work out the usual look and compact for each version of view. It is better to think more deeply about possible design options now

amje (approved):

mournfulCoroner (approved):

PR #2179, opened 2025-03-24
Bug showcase: not working ref inside Dialog in uikit 7

Ruminat:
### Bug demonstration (not an actual Pull Request)

Open [Overlays/Dialog/Default](https://preview.gravity-ui.com/uikit/2179/?path=/story/components-overlays-dialog--default) and take a look at the console
it states that `buttonRef.current is null`
Although in uikit 6 it wouldn't be `null`

## Summary by Sourcery

Bug Fixes:
- Demonstrates a bug where a ref inside a Dialog component is null when accessed in a useEffect hook.

### DIFF
--- a/src/components/Dialog/__stories__/Dialog.stories.tsx
+++ b/src/components/Dialog/__stories__/Dialog.stories.tsx
@@ -26,6 +26,9 @@ const DefaultTemplate: StoryFn<DialogProps & {showError: boolean}> = ({showError
             <Button view="normal" onClick={() => setOpen(true)}>
                 Show dialog
             </Button>
+
+            <NotWorkingRefShowcase />
+
             <Dialog
                 {...args}
                 onClose={() => setOpen(false)}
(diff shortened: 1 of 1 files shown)


Ruminat:
> ref works, but in this example it won't be set on the initial render, since the Modal is only mounted into the DOM on the second render. If you need ref on initial render of Button, you can extract dialog's body to separate component:
> 
> ```js
> function DialogContent() {
>     const ref = React.useRef<HTMLButtonElement>(null);
> 
>     React.useEffect(() => {
>         console.log(ref.current);
>     }, []);
> 
>     return <Button ref={ref}>{'ref.current is defined inside useEffect : > )'}</Button>
> }
> 
> <Dialog open={true} onClose={() => console.log('close')}>
>     <Dialog.Header caption="Ref example" />
>     <Dialog.Body>
>         <DialogContent />
>     </Dialog.Body>
>     <Dialog.Footer textButtonApply=":)"></Dialog.Footer>
> </Dialog>
> ```
> 
> or you can use `useState` instead of `useRef`:
> 
> ```js
>     const [node, setNode] = React.useState<HTMLButtonElement | null>(null);
> 
>     React.useEffect(() => {
>         console.log(node);
>     }, [node]);
> ```

But that way I won't be able to pass this ref to Dialog's `initialFocus` and all the code that worked in uikit 6 with refs in Dialogs now doesn't work in uikit 7

Is there a reason for this change?

Ruminat:
Both solutions require some workarounds which feels quite wrong and unstable to me

ValeraS:
initialFocus works with ref as expected. If it doesn't work for you, please  provide reproduction.

Ruminat:
> initialFocus works with ref as expected. If it doesn't work for you, please provide reproduction.

If I need to use ref from `initialFocus` in `useEffect`, I won't be able to (unless I use `useState`)

ValeraS:
> > initialFocus works with ref as expected. If it doesn't work for you, please provide reproduction.
> 
> If I need to use ref from `initialFocus` in `useEffect`, I won't be able to (unless I use `useState`)

It is technically incorrect to assume that ref is filled on initial render in a situation where you have no control over when the content is mounted into the DOM. And the fact that it worked before is just a happy coincidence. 

BTW, you can pass number to the initialFocus to focus an element by index in tab-order?

Ruminat:
Ok, I guess I'll just use some workarounds for this

Issue #2192, opened 2025-03-28
Do not set `title` to Breadcrumbs.Item by default


ogonkov:
### Description

In Breadcrumbs.Item, HTML attribute `title` is set by default to element. This make tricky to set `Tooltip` component to Breadcrumb.Item, and not get 2 tooltips

https://github.com/gravity-ui/uikit/blob/862419ffe1c78c79849fe304fa6a8437555d37d8/src/components/Breadcrumbs/BreadcrumbsItem.tsx#L42-L45

### Suggested solution

I guess it is better to not set `title` at all, this is more predictable behavior

### Validations

- [x] Check that there isn't [already an issue](https://github.com/gravity-ui/uikit/issues) that request the same feature to avoid creating a duplicate.

PR #2196, opened 2025-03-30
fix(Breadcrumbs): handle empty string in breadcrumbs item title

korvin89:
Closes https://github.com/gravity-ui/uikit/issues/2192


src/components/Breadcrumbs/BreadcrumbsItem.tsx
```diff
@@ -32,18 +32,14 @@ function BreadcrumbsItem(props: BreadcrumbsItemProps, ref: React.ForwardedRef<HT
         ping,
         referrerPolicy,
         children,
+        title,
```
ValeraS, line 35: let's remove the title from `linkProps`, it will remain in `restProps`

ValeraS (changes requested):

ValeraS (approved):

PR #2201, opened 2025-03-31
feat(Tab): add `onClick` handler

ogonkov:
## Summary by Sourcery

Add optional `onClick` handler to Tab component to support custom click event handling

New Features:
- Introduce an optional `onClick` prop to the Tab component that allows custom event handling for mouse and keyboard interactions

Enhancements:
- Update `useTab` hook to invoke the optional `onClick` handler before updating the tab context

### DIFF
--- a/src/components/tabs/hooks/useTab.ts
+++ b/src/components/tabs/hooks/useTab.ts
@@ -21,14 +21,18 @@ export function useTab(tabProps: TabProps) {
     const isDisabled = tabProps.disabled;
     const isFocused = tabContext.isFocused;
 
-    const onClick = () => {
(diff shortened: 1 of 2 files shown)


sofiushko:
@amje 
Hi, is there some common solution for router navigation in components?

amje:
> @amje Hi, is there some common solution for router navigation in components?

There is an approach: add `component` prop, which accepts any React component, and spread all props to it. See Button as an example

amje:
@ogonkov I've implement passing all props in #2216 as discussed

ogonkov:
We will use @amje approach

Issue #2219, opened 2025-04-07
Modal component errors when changing size


PahaN47:
### Describe the bug

Every time Modal component changes size there is an error: `Uncaught ResizeObserver loop completed with undelivered notifications.`

### Reproduction

https://preview.gravity-ui.com/uikit/iframe.html?globals=&args=&id=components-overlays-modal--default&viewMode=story

### Steps to reproduce

Open modal with dynamic content in storybook, then check console

### Environment

MacOS 14, Safari and Chromium (Yandex.Browser)

### Validations

- [x] Check that there isn't [already an issue](https://github.com/gravity-ui/uikit/issues) that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

amje:
@roberthovsepyan Take a look please

roberthovsepyan:
@amje this seems to be a benign warning. It occurs when resizing the element inside ResizeObserver callback, which is exactly what we do here. [Source](https://trackjs.com/javascript-errors/resizeobserver-loop-completed-with-undelivered-notifications/)

> JavaScript on your page is using the browser’s [ResizeObserver API](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) to invoke a callback when a DOM element’s size changes. The callback can modify the same element that originally triggered the resize event, triggering a second resize event. That new event is not sent to your callback until the next render cycle. The browser sends the ResizeObserver loop completed with undelivered notifications. error to inform you that this deferral has happened.

There doesn't seem to be a good workaround, but the warning doesn't affect functionality. Maybe we should just catch this error so it isn't logged to console?

PR #2247, opened 2025-04-24
fix(Dialog): handle ResizeObserver loop error

roberthovsepyan:
## Summary by Sourcery

Bug Fixes:
- Add an error event listener to prevent a harmless ResizeObserver loop error from being reported, which occurs during height changes but does not impact functionality


roberthovsepyan:
fixes this https://github.com/gravity-ui/uikit/issues/2219#issuecomment-2786992736

amje:
@roberthovsepyan Let's instead of skipping the error, we unsubscribe from the observer for the moment we manipulate height, like MUI does: https://github.com/mui/material-ui/blob/master/packages/mui-material/src/TextareaAutosize/TextareaAutosize.tsx#L182

src/hooks/useResizeObserver/useResizeObserver.ts
```diff
@@ -2,7 +2,7 @@ import * as React from 'react';
 
 interface UseResizeObserverProps<T> {
     ref: React.RefObject<T | null | undefined> | undefined;
-    onResize: () => void;
+    onResize: (observer?: ResizeObserver) => void;
```
amje, line 5: It's a bit odd to have observer inside onResize callback. Maybe we simply get it from the hook as a return value?

amje (changes requested):

src/hooks/useResizeObserver/useResizeObserver.ts
roberthovsepyan: The thing is we can't cleanly return it from the hook because we first need to pass `onResize` to the hook itself, so it's a cyclical dependency. We can only bypass this if we don't use `useCallback`, and it works but triggers the hook more times so observer is recreated on every render, which I don't like

I think passing it in is more elegant because we need to access observer in onResize callback itself
However if you disagree I can change it to this
Screenshot 2025-05-06 at 16 12 23

amje (approved):

PR #2256, opened 2025-04-30
feat(Label): added "info" type label

loktionov94:
## Summary by Sourcery

Enhance Label component with new 'info' type and improved interactive behavior

New Features:
- Added 'info' type label with an information icon
- Introduced more granular interactive prop with multiple interaction modes

Enhancements:
- Expanded interactive prop to support 'hover' and 'icon' interaction modes
- Updated documentation and type definitions to reflect new label capabilities


korvin89:
@loktionov94 Hey! Could you provide the key of the design ticket?

loktionov94:
@korvin89 Hi! Design ticket key - UXRFC-600

amje:
"Close" story is working incorrectly, only the close button should be hovered if `interactive = undefined`

loktionov94:
@amje about "Close" type, is this behavior typical only for `interactive === undefined` (hover effect only when hovering over the button) or for any interactive value?

amje:
> @amje about "Close" type, is this behavior typical only for `interactive === undefined` (hover effect only when hovering over the button) or for any interactive value?

I think we could simplify the API here. We don't need to change `interactive` prop. It could remain in boolean type. For `type = info`, we add icon on the right and toggle `isInteractive` to true and that's all. `type = close` and `interactive` props should work independently like before

goshander (approved):

src/components/Label/Label.tsx
```diff
@@ -17,12 +17,12 @@ import './Label.scss';
...
-type SizeMapType = {copyIconSize: number; closeIconSize: number};
+type SizeMapType = {copyIconSize: number; closeIconSize: number; infoIconSize: number};
 
 const sizeMap: Record<string, SizeMapType> = {
-    xs: {copyIconSize: 12, closeIconSize: 12},
-    s: {copyIconSize: 14, closeIconSize: 14},
-    m: {copyIconSize: 16, closeIconSize: 16},
+    xs: {copyIconSize: 12, closeIconSize: 12, infoIconSize: 12},
+    s: {copyIconSize: 14, closeIconSize: 14, infoIconSize: 14},
+    m: {copyIconSize: 16, closeIconSize: 16, infoIconSize: 16},
```
amje, line 25: Let's simplify to `Record<string, number>` type, all icons should be the same size if they're from gravity icons

loktionov94: done

amje (approved):

Issue #2253, opened 2025-04-30
Add custom icon size to HelpMark


kvestus:
### Description

I want to change icon size of HelpMark

### Suggested solution

Add custom icon size to HelpMark

### Validations

- [x] Check that there isn't [already an issue](https://github.com/gravity-ui/uikit/issues) that request the same feature to avoid creating a duplicate.

PR #2257, opened 2025-05-02
feat(HelpMark): add size prop

uuxxx:
resolve https://github.com/gravity-ui/uikit/issues/2253

## Summary by Sourcery

Add size prop to HelpMark component to allow customizable icon sizes

New Features:
- Add configurable size prop to HelpMark component
- Introduce size variant story in Storybook

Enhancements:
- Update Showcase component to support alignment prop
- Replace hardcoded icon size with configurable size

### DIFF
--- a/src/components/HelpMark/HelpMark.tsx
+++ b/src/components/HelpMark/HelpMark.tsx
@@ -2,25 +2,28 @@ import * as React from 'react';
 
 import {CircleQuestion} from '@gravity-ui/icons';
 
-import type {IconProps} from '../Icon';
 import {Icon} from '../Icon';
 import {Popover} from '../Popover';
 import type {PopoverProps} from '../Popover';
 import type {QAProps} from '../types';
 import {block} from '../utils/cn';
 
+import {ICON_SIZE_MAP} from './constants';
+
 import './HelpMark.scss';
 
 const b = block('help-mark');
 
+type IconSize = keyof typeof ICON_SIZE_MAP;
+
 export interface HelpMarkProps extends QAProps, React.ButtonHTMLAttributes<HTMLButtonElement> {
-    iconProps?: Omit<IconProps, 'data'>;
+    iconSize?: IconSize;
     popoverProps?: Omit<PopoverProps, 'children'>;
     children?: React.ReactNode;
 }
 
 export const HelpMark = React.forwardRef<HTMLButtonElement, HelpMarkProps>(function HelpMark(
-    {children, qa, className, popoverProps, iconProps, ...restProps},
+    {children, qa, className, iconSize = 'm', popoverProps, ...restProps},
     ref,
 ) {
     return (
@@ -36,7 +39,7 @@ export const HelpMark = React.forwardRef<HTMLButtonElement, HelpMarkProps>(funct
                 className={b(null, className)}
                 data-qa={qa}
             >
-                <Icon data={CircleQuestion} className={b('icon')} {...iconProps} />
+                <Icon data={CircleQuestion} size={ICON_SIZE_MAP[iconSize]} className={b('icon')} />
             </button>
         </Popover>
     );
--- a/src/components/HelpMark/HelpMark.tsx
+++ b/src/components/HelpMark/HelpMark.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 
 import {CircleQuestion} from '@gravity-ui/icons';
 
+import type {IconProps} from '../Icon';
 import {Icon} from '../Icon';
 import {Popover} from '../Popover';
 import type {PopoverProps} from '../Popover';
@@ -11,15 +12,15 @@ import {block} from '../utils/cn';
 import './HelpMark.scss';
 
 const b = block('help-mark');
-const ICON_SIZE = 16;
 
 export interface HelpMarkProps extends QAProps, React.ButtonHTMLAttributes<HTMLButtonElement> {
+    iconProps?: Omit<IconProps, 'data'>;
     popoverProps?: Omit<PopoverProps, 'children'>;
     children?: React.ReactNode;
 }
 
 export const HelpMark = React.forwardRef<HTMLButtonElement, HelpMarkProps>(function HelpMark(
-    {children, qa, className, popoverProps, ...restProps},
+    {children, qa, className, popoverProps, iconProps, ...restProps},
     ref,
 ) {
     return (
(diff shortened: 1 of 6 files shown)


src/components/HelpMark/HelpMark.tsx
```diff
@@ -35,7 +36,7 @@ export const HelpMark = React.forwardRef<HTMLButtonElement, HelpMarkProps>(funct
                 className={b(null, className)}
                 data-qa={qa}
             >
-                <Icon data={CircleQuestion} size={ICON_SIZE} className={b('icon')} />
+                <Icon data={CircleQuestion} className={b('icon')} {...iconProps} />
```
kvestus, line 39: For compatibility, let's keep the previous size if not provided in iconProps

kvestus (changes requested):

Raubzeug (changes requested):

src/components/HelpMark/__stories__/HelpMark.stories.tsx
```diff
@@ -28,4 +35,29 @@ export default {
 
 type Story = StoryObj<typeof HelpMark>;
 
-export const Default: Story = {args: {children: 'Some content'}};
+export const Default: Story = {};
+
+export const Size: Story = {
```
ogonkov, line 40: Better use mdx, and include this different variations via `<Canvas/>` component, this will utilize default Storybook behavior with story code switcher

For example, i changed Select stories this way

https://github.com/gravity-ui/uikit/blob/e2b91bf0c6e20f9e2ca7a14539342cc69885ad07/src/components/Select/__stories__/Showcase.mdx?plain=1#L4-L6

uuxxx:
removed iconProps, left only size prop with size grid. added mdx for size variations

src/components/HelpMark/HelpMark.tsx
```diff
@@ -8,18 +8,22 @@ import type {PopoverProps} from '../Popover';
 import type {QAProps} from '../types';
 import {block} from '../utils/cn';
 
+import {ICON_SIZE_MAP} from './consts';
```
ogonkov, line 11: I think we usually call this file `constants`

src/components/HelpMark/consts.ts
```diff
@@ -0,0 +1,8 @@
+import type {IconSize} from './HelpMark';
+
+export const ICON_SIZE_MAP: Record<IconSize, number> = {
```
ogonkov, line 3: I think that it would be better to have `IconSize` to be inherited from this map, to have type updated after we change map

```
type IconSize = keyof typeof ICON_SIZE_MAP;
```

uuxxx:
@ogonkov fixed

src/components/HelpMark/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const ICON_SIZE_MAP = {
```
Raubzeug, line 1: I think this map should be same as https://github.com/gravity-ui/uikit/blob/main/src/components/Button/constants.ts#L25 , it is already approoved by designers.

Raubzeug (changes requested):

src/components/HelpMark/constants.ts
uuxxx: why does it have m and l both equal to 16? looks like a bug

Raubzeug (approved):

src/components/HelpMark/__stories__/HelpMark.stories.tsx
```diff
@@ -24,8 +34,44 @@ export default {
...
+        </React.Fragment>
+    );
+};
+
+export const Default = {};
+
+export const Size = {
+    name: 'Size',
+    decorators: [Title],
+    tags: ['!dev'],
```
amje, line 66: Let's keep this story in the list. We have story `Size` for all components

src/components/HelpMark/__stories__/HelpMark.stories.scss
```diff
@@ -0,0 +1,5 @@
+.help-mark-stories {
+    &__showcase {
+        align-items: center;
```
amje, line 3: Default alignment (start) should be fine

amje (changes requested):

src/components/HelpMark/constants.ts
```diff
@@ -0,0 +1,7 @@
...
+    xs: 12,
```
Raubzeug, line 2: Feedback from designer: lets remove xs size, it's too small and it's hard to point cursor to it.

Raubzeug (changes requested):

Raubzeug (approved):

src/components/HelpMark/__stories__/HelpMark.stories.scss
uuxxx: default alignment is stretch and it doesn't look good in this case (smaller buttons stretch to the parent container's size). Now fixed it by using `<ShowcaseItem/>`

amje (approved):

src/components/HelpMark/HelpMark.tsx
```diff
@@ -8,18 +8,22 @@ import type {PopoverProps} from '../Popover';
...
 import {block} from '../utils/cn';
 
+import {ICON_SIZE_MAP} from './constants';
+
 import './HelpMark.scss';
 
 const b = block('help-mark');
-const ICON_SIZE = 16;
+
+type IconSize = keyof typeof ICON_SIZE_MAP;
```
kvestus, line 17: Let's add the ability to set a custom size as number

src/components/HelpMark/consts.ts
kvestus: Let's add the ability to set a custom size as number. I created issue for custom size and map doesn't have needed size)

src/components/HelpMark/constants.ts
kvestus: I need it)

Raubzeug: @kvestus  Hi! I'll return on Monday, need to discuss it with designer!

Raubzeug: @kvestus it seems that xs size should be added in separate PR, cause our designers asks to add virtual padding for this size (to extend hover area). Will you agree with this?

kvestus: @Raubzeug Hi! Yes, thank you!

PR #2260, opened 2025-05-05
fix(Label): fixed hover state for Label with type "copy"

loktionov94:
UXRFC-657

## Summary by Sourcery

Modify hover state and action types for Label component to improve interaction behavior

Bug Fixes:
- Fixed hover state styling for Label component to correctly handle different button types
- Ensured proper hover interactions for copy and close button addons

Enhancements:
- Updated CSS selector to exclude copy button from default hover styles
- Added action modifiers for button addons to differentiate interaction types


goshander:
@loktionov94 the behavior of the copy icon with additional icons is broken, see showcase page

https://preview.gravity-ui.com/uikit/2260/?path=/story/components-data-display-label--showcase-story

loktionov94:
@goshander current behaviour in preview for copy lables is not broken. In discussions in the design ticket, there is an agreement about full hover on the copy label. Current story doesn`t contain `copyText` in cases with copy label. I`ll fix this storybook story

goshander (approved):

PR #2278, opened 2025-05-23
feat(Select): add fixed top zone on mobile

iapolya:
**Before:**

https://github.com/user-attachments/assets/e24860fc-7309-4c1c-98ee-f0c16a4d4b15

**After:**

https://github.com/user-attachments/assets/490f8d9a-d520-4927-98df-51aeb9d89c2f

If you customize the renderPopup for the mobile version, the following drawbacks arise:
- High class nesting (deep hierarchy) requiring display properties to be overridden
- Breaks the swipe-on-scroll behavior

## Summary by Sourcery

Introduce a fixed top zone on mobile for both Sheet and Select components by adding dedicated props and separating top content from scrollable areas.

New Features:
- Add `topContent` prop to `Sheet` and `SheetContent` to render non-scrolling header content.
- Add `renderTop` prop to `Select` and `SelectPopup` to enable a customizable fixed top section above the filter.
- Introduce new Storybook stories (`WithVirtualizedListAndCustomPopup`, `WithTopContent`) and update `DefaultShowcase` to demonstrate the fixed top zone.

Enhancements:
- Refactor sheet layout to isolate a top container and compute its height for proper scroll area adjustment.
- Replace bottom margins with padding in Sheet story styles to maintain consistent spacing.
- Adjust Select popup rendering logic to integrate `renderTop` and preserve swipe-on-scroll behavior on mobile.

Documentation:
- Update Storybook MDX to include the new top content stories

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -65,6 +65,7 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function 
         renderSelectedOption,
         renderEmptyOptions,
         renderPopup = DEFAULT_RENDER_POPUP,
+        renderTop,
         getOptionHeight,
         getOptionGroupHeight,
         filterOption,
@@ -311,6 +312,18 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function 
         return <EmptyOptions filter={filter} renderEmptyOptions={renderEmptyOptions} />;
     };
 
+    const _renderPopupFilter = () => {
+        if (renderTop && mobile) {
+            return () => null;
+        }
+
+        if (renderTop) {
+            return () => renderTop({renderFilter: _renderFilter});
+        }
+
+        return _renderFilter;
+    };
+
     return (
         <div
             ref={controlWrapRef}
@@ -372,8 +385,12 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function 
                           }
                         : undefined
                 }
+                topContent={renderTop?.({renderFilter: _renderFilter})}
             >
-                {renderPopup({renderFilter: _renderFilter, renderList: _renderList})}
+                {renderPopup({
+                    renderFilter: _renderPopupFilter(),
+                    renderList: _renderList,
+                })}
             </SelectPopup>
             <OuterAdditionalContent
                 errorMessage={isErrorMsgVisible ? errorMessage : null}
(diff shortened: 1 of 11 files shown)


src/components/Select/Select.tsx
```diff
@@ -311,6 +312,18 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function
         return <EmptyOptions filter={filter} renderEmptyOptions={renderEmptyOptions} />;
     };
 
+    const _renderPopupFilter = () => {
```
korvin89, line 315: Select has `renderPopup` property, is it not enough?

korvin89 (changes requested):

src/components/Select/Select.tsx
iapolya: https://github.com/gravity-ui/uikit/blob/feat/select-zone/src/components/Sheet/SheetContent.tsx#L195

When using renderSelect, all content (both filters and the list) is placed inside .sheet-scroll-container.

Drawbacks:
- We have to override styles for deeply nested classes
- Logic for events like [onTouchStart](https://github.com/gravity-ui/uikit/blob/feat/select-zone/src/components/Sheet/SheetContent.tsx#L197) breaks

korvin89: Please show me what do you wanna do? I'll try to make it in sandbox

iapolya: @korvin89, filter and scrollable content 

1. https://preview.gravity-ui.com/uikit/?path=/docs/components-inputs-select--showcase&globals=platform:mobile
Story **With Custom Popup** + add height to .g-list

2. **With Virtualized List And Custom Popup**
https://preview.gravity-ui.com/uikit/2278/?path=/docs/components-inputs-select--showcase

korvin89:
@iapolya I realized the problem, I apologize that it took some time.

I suggest the following solution:
- add the `renderContent` prop to the `Sheet` component (or something like that), for example [as in Select](https://github.com/gravity-ui/uikit/blob/60b9655a75a5d1884eeb6c69761e8d0f44654331/src/components/Select/types.ts#L61-L64)
- use this new prop in `Select`

@mournfulCoroner What do you think about it?

iapolya:
> @iapolya I realized the problem, I apologize that it took some time.
> 
> I suggest the following solution:
> 
> * add the `renderContent` prop to the `Sheet` component (or something like that), for example [as in Select](https://github.com/gravity-ui/uikit/blob/60b9655a75a5d1884eeb6c69761e8d0f44654331/src/components/Select/types.ts#L61-L64)
> * use this new prop in `Select`
> 
> @mournfulCoroner What do you think about it?

PR #2301, opened 2025-06-13
feat(Table): add `rowActionsIcon` prop

vijay2904:
Added `rowActionsIcon` prop to Table with Actions.

Added Test cases to check different types of inputs to `rowActionsIcon` prop.

## Summary by Sourcery

Add a `rowActionsIcon` prop to the table-with-actions HOC to allow overriding the default actions icon, update Storybook examples to demonstrate its use, and document the new prop in READMEs.

New Features:
- Add `rowActionsIcon` prop to `withTableActions` HOC to customize the row actions icon.

Enhancements:
- Extend Storybook stories to show `rowActionsIcon` with an SVG element, string emoji, and React component.

Documentation:
- Document `rowActionsIcon` prop in both English and Russian README files.


Raubzeug:
@vijay2904 Hi! 👋

Thank you so much for taking the time to contribute to the project and for the pull request! 

I noticed that your `rowActionsIcon` property is aimed at customizing the actions icon, which is a great idea! However, we already have a `renderRowActions` property that allows full customization of the entire actions section for each row, including the icon itself.

Since `renderRowActions` provides the flexibility to render any custom content (including custom icons), I think it might cover the same use case as your proposed `rowActionsIcon`. This helps us keep the API surface smaller and more consistent.

What do you think? Can `renderRowActions` cover your needs? If there's a genuine gap that your approach fills better, I'm totally open to reconsidering!

Raubzeug:
@vijay2904 it seems test are still broken :(

amje:
@vijay2904 It seems you've updated snapshot from a non-linux machine. Use this command:
```sh
npm run playwright:docker:update -- -g @Table
```
to update snapshot from the docker

vijay2904:
@Raubzeug @amje Finally, all the tests are successful. Thanks for your support : )

PR #2304, opened 2025-06-14
feat(CopyToClipboard): add fallback for copyText

korvin89:
## Summary by Sourcery

Provide a fallback mechanism for copying text using document.execCommand when the Clipboard API is unavailable or fails, and update tests accordingly.

New Features:
- Implement copyTextFallback to copy text via a hidden input and execCommand('copy') when navigator.clipboard.writeText is not available.
- Expand copyText function to use the new fallback and enhance error handling for missing APIs.

Enhancements:
- Improve error messages to distinguish between missing navigator.clipboard and document support.

Tests:
- Mock document.execCommand in tests to simulate fallback behavior.
- Add a test case for successful copy using the fallback path.
- Rename existing test to specifically cover clipboard.writeText failure scenario.


src/components/CopyToClipboard/copyText.ts
```diff
@@ -1,11 +1,43 @@
 export function copyText(text: string) {
     if (navigator?.clipboard?.writeText) {
-        try {
-            return navigator.clipboard.writeText(text);
-        } catch (error) {
-            return Promise.reject(error);
+        return navigator.clipboard.writeText(text);
+    }
+
+    if (typeof document !== 'undefined') {
```
ogonkov, line 10: What about navigator usage above? I guess navigator usage will break ssr aswell

src/components/CopyToClipboard/copyText.ts
```diff
@@ -1,11 +1,43 @@
 export function copyText(text: string) {
     if (navigator?.clipboard?.writeText) {
-        try {
```
ogonkov, line 3: This try..catch was added in case of NotAllowed exception, can we return try catch?

src/components/CopyToClipboard/copyText.ts
```diff
@@ -1,11 +1,43 @@
...
-    return Promise.reject(new Error('Native copy is not available'));
+    if (typeof document !== 'undefined') {
+        return copyTextFallback(text);
+    }
+
+    return Promise.reject(new Error('Neither navigator.clipboard nor document is available'));
+}
+
+async function copyTextFallback(text: string) {
+    const activeElement = document.activeElement as HTMLElement;
```
ogonkov: may be check for instance instead of casting?

```
const activeElement = document.activeElement;

// ...

if (activeElement instanceof HTMLElement) {
  activeElement.focus();
}
```

src/components/CopyToClipboard/copyText.ts
```diff
@@ -1,11 +1,43 @@
...
+        textarea.style.position = 'absolute';
+        textarea.style.left = '1000%';
+        // Preserve spaces and line breaks
+        textarea.style.whiteSpace = 'pre';
+        // Do not inherit user-select (it may be `none`)
+        textarea.style.userSelect = 'text';
+        textarea.value = text;
+        document.body.append(textarea);
+        textarea.select();
+        document.execCommand('copy');
```
ogonkov: We should check returned value here to make sure that command was executed, otherwise i guess we need to throw error

PR #2310, opened 2025-06-20
feat(SegmentedRadioGroup): allow `children` composition

ogonkov:
gravity-ui/rfc#28

## Summary by Sourcery

Allow direct composition of SegmentedRadioGroup.Option children by introducing context providers and refactoring the radio group hook and option component accordingly

New Features:
- Support rendering SegmentedRadioGroup.Option as children instead of only via the options prop
- Add a story to demonstrate children composition with wrappers like Tooltip

Enhancements:
- Introduce SegmentedRadioGroupContext to share group name, value, disabled state, ref, and change handler
- Refactor useRadioGroup to return contextProps for stable and dynamic state
- Update SegmentedRadioGroupOption to consume context for selection logic and merge refs via useForkRef

Documentation:
- Add ChildrenMap story showcasing option composition

Chores:
- Add context and type definitions for useRadioGroup and useFormResetHandler parameters


.nvmrc
```diff
@@ -1 +1 @@
-20
+20.19
```
amje: Why?

src/components/SegmentedRadioGroup/SegmentedRadioGroup.tsx
```diff
@@ -38,33 +42,27 @@ export const SegmentedRadioGroup = React.forwardRef(function SegmentedRadioGroup
...
-            style={style}
-            className={b({size, width}, className)}
-            data-qa={qa}
-        >
-            {optionsProps.map((optionProps) => (
-                <Option {...optionProps} key={optionProps.value} />
-            ))}
-        </div>
+        <SegmentedRadioGroupContextStable.Provider value={contextProps.stable}>
+            <SegmentedRadioGroupContextValue.Provider value={contextProps.value}>
```
amje: What about using single context?

src/components/SegmentedRadioGroup/SegmentedRadioGroupOption.tsx
```diff
@@ -24,18 +30,34 @@ type SegmentedRadioGroupOptionComponentType = <T extends string = string>(
 export const SegmentedRadioGroupOption = React.forwardRef(function SegmentedRadioGroupOption<
     T extends string,
 >(props: SegmentedRadioGroupOptionProps<T>, ref: React.ForwardedRef<HTMLLabelElement>) {
-    const {disabled = false, content, children, title} = props;
-    const {checked, inputProps} = useRadio(props);
+    const {
+        name,
+        disabled: contextDisabled,
+        ref: contextRef,
```
amje: What is the point of storing ref inside the context? It will always be the last option, won't it?

src/components/SegmentedRadioGroup/SegmentedRadioGroupOption.tsx
```diff
@@ -24,18 +30,34 @@ type SegmentedRadioGroupOptionComponentType = <T extends string = string>(
...
+        onChange,
+    } = React.useContext(SegmentedRadioGroupContextStable);
+    const {currentValue} = React.useContext(SegmentedRadioGroupContextValue);
+    const {disabled: propsDisabled, content, children, title, value} = props;
+    const disabled = propsDisabled || contextDisabled;
```
amje: Consider using this naming `disabledProp`, `disabledContext` instead. This meets our code style

src/hooks/private/useFormResetHandler/types.ts
```diff
@@ -0,0 +1,5 @@
+export type UseFormResetHandlerParams =
+    | HTMLInputElement
+    | HTMLTextAreaElement
+    | HTMLSelectElement
+    | HTMLLabelElement;
```
amje: Why we would want to reset the label element?

.nvmrc
ogonkov: stylelint-order expect this version of nodejs

src/components/SegmentedRadioGroup/SegmentedRadioGroup.tsx
ogonkov: May be, they both used in same component

.nvmrc
amje: The reason why there is only major part is that nvm would not install a specific version of node, if user has greater version. That's for deduplication. Even though it's innacurate

src/components/SegmentedRadioGroup/SegmentedRadioGroupOption.tsx
ogonkov: I guess you right, i already asked @zamkovskaya about ref usage in useRadio/useRadioGroup, they both enforce form reset logic by this ref, and it is looks like duplication

src/hooks/private/useFormResetHandler/types.ts
ogonkov: This is what actually happens in current code, i just fixed typings here

src/hooks/private/useRadioGroup/useRadioGroup.ts
```diff
@@ -60,6 +63,14 @@ export function useRadioGroup<ValueType extends string = string>(
         }
     };
 
+    const contextProps = {
```
amje: We should memoize it for not updating the context on every render

src/components/SegmentedRadioGroup/__tests__/SegmentedRadioGroup.test.tsx
```diff
@@ -243,4 +255,43 @@ describe('SegmentedRadioGroup', () => {
             },
         );
     });
+
+    describe.each([
+        {name: 'options', options},
+        {name: 'children', children},
+    ])('with $name', ({options, children}) => {
+        describe('in form', () => {
+            test('should support form reset', async () => {
```
amje: There is no reason to test form reset in both input format

src/components/SegmentedRadioGroup/__stories__/SegmentedRadioGroup.stories.tsx
```diff
@@ -78,6 +79,19 @@ export const Disabled: Story = {
     },
 };
 
+export const ChildrenMap: Story = {
```
amje: Let's call the story `WithTooltip` and wrap every option in tooltip. The story should demonstrate how it looks, not how it works inside

src/components/SegmentedRadioGroup/SegmentedRadioGroupContext.ts
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react';
+
+import type {UseRadioGroupContextProps} from '../../hooks/private/useRadioGroup/types';
+
+export const SegmentedRadioGroupContext = React.createContext<UseRadioGroupContextProps>({
```
amje: It would be better to make this context a part of `useRadioGroup` hook, and apply it in both `SegmentedRadioGroup` and `RadioGroup`. And maybe it should be named just `RadioGroupContext`

src/components/SegmentedRadioGroup/SegmentedRadioGroupContext.ts
```diff
@@ -0,0 +1,13 @@
...
+    name: '',
+    currentValue: '',
+    disabled: false,
+    ref: () => {},
+    onChange: () => {},
+});
```
amje: Let's make the default value of the context to be null. And throw on incorrect usage: https://github.com/gravity-ui/uikit/blob/main/src/components/tabs/hooks/useTab.ts#L14

amje (changes requested):

src/components/SegmentedRadioGroup/SegmentedRadioGroup.tsx
```diff
@@ -38,33 +39,25 @@ export const SegmentedRadioGroup = React.forwardRef(function SegmentedRadioGroup
...
+        <SegmentedRadioGroupContext.Provider value={contextProps}>
+            <div
+                {...containerProps}
+                ref={ref}
+                style={style}
+                className={b({size, width}, className)}
+                data-qa={qa}
+            >
+                {children ||
+                    options?.map((optionProps) => (
```
amje: We should use `optionsProps` from the hook. There is mapping happening inside the hook

src/hooks/private/useRadioGroup/types.ts
```diff
@@ -0,0 +1,10 @@
+import type {UseFormResetHandlerParams} from '../useFormResetHandler/types';
+
+export type UseRadioGroupContextProps = {
```
amje: The props are for context not hook
```suggestion
export type RadioGroupContextProps = {
```

src/components/SegmentedRadioGroup/__stories__/SegmentedRadioGroup.stories.tsx
ogonkov: May be use different floating components, like Popover?

amje: Not sure if it's correct to wrap with Popover. Let's wrap all options in Tooltips and that's all

amje (approved):

zamkovskaya (approved):

PR #2332, opened 2025-07-04
fix: fix reduced motion behavior

amje:
## Summary by Sourcery

Respect user’s prefers-reduced-motion setting across components by centralizing transition behavior in a new hook and disabling animations when reduced motion is preferred

New Features:
- Introduce useFloatingTransition hook to consolidate Floating UI transition logic
- Add useMatchMedia hook to listen to media query matches
- Provide README documentation for new useFloatingTransition and useMatchMedia hooks

Enhancements:
- Refactor Popup and Modal components to use useFloatingTransition instead of custom transition logic
- Update SheetContent and useAnimateHeight hook to respect prefers-reduced-motion and disable animations accordingly

Documentation:
- Add README files for useFloatingTransition and useMatchMedia hooks

Tests:
- Mock window.matchMedia in test setup to enable reduced motion handling in tests

Chores:
- Adjust global reduced motion styles to use 0ms for transitions and animations

### DIFF
--- a/src/hooks/private/useMatchMedia/useMatchMedia.ts
+++ b/src/hooks/private/useMatchMedia/useMatchMedia.ts
@@ -1,0 +1,24 @@
+import * as React from 'react';
+
+export interface UseMatchMediaProps {
+    query: string;
+}
+
+export function useMatchMedia({query}: UseMatchMediaProps) {
+    const media = React.useMemo(
+        () => (typeof window === 'undefined' ? null : window.matchMedia(query)),
+        [query],
+    );
+    const [matches, setMatches] = React.useState(Boolean(media?.matches));
+
+    React.useEffect(() => {
+        const handleChange = (event: MediaQueryListEvent) => {
+            setMatches(event.matches);
+        };
+
+        media?.addEventListener('change', handleChange);
+        return () => media?.removeEventListener('change', handleChange);
+    }, [media]);
+
+    return matches;
+}
--- a/src/hooks/private/useMatchMedia/useMatchMedia.ts
+++ b/src/hooks/private/useMatchMedia/useMatchMedia.ts
@@ -1,24 +1,24 @@
 import * as React from 'react';
 
 export interface UseMatchMediaProps {
-    query: string;
+    media: string;
 }
 
-export function useMatchMedia({query}: UseMatchMediaProps) {
-    const media = React.useMemo(
-        () => (typeof window === 'undefined' ? null : window.matchMedia(query)),
-        [query],
+export function useMatchMedia({media}: UseMatchMediaProps) {
+    const mql = React.useMemo(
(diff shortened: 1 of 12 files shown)


styles/styles.scss
```diff
@@ -42,16 +42,18 @@
...
     *::before,
     *::after {
         /* stylelint-disable-next-line declaration-no-important */
         scroll-behavior: auto !important;
         /* stylelint-disable-next-line declaration-no-important */
-        transition-duration: 0.001ms !important;
+        transition-duration: 0ms !important;
         /* stylelint-disable-next-line declaration-no-important */
-        animation-duration: 0.001ms !important;
+        animation-duration: 0ms !important;
```
ogonkov: As far as i know with 0 duration we unable to listen for transition/animation end events, it's not quite clear why we reduce duration to 0

src/hooks/private/useMatchMedia/useMatchMedia.ts
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react';
+
+export interface UseMatchMediaProps {
+    media: string;
+}
+
+export function useMatchMedia({media}: UseMatchMediaProps) {
+    const mql = React.useMemo(
```
ogonkov, line 8: Why we need to memoize this object, couldn't we keep it in useEffect?

styles/styles.scss
amje: Safari with the `Reduced motion` setting enabled has odd behavior: hover states are stuck, scroll position resets, modals not shown.

src/hooks/private/useMatchMedia/useMatchMedia.ts
amje: We need it to for the initial state

styles/styles.scss
korvin89: I suppose this will break the logic in those places that are tied to the beginning/end of the animation?

amje: @korvin89 @ogonkov Fixed such places (Modal, Sheet,)

ogonkov: But that still break code that external users could use in their apps, isn't it?

amje: > But that still break code that external users could use in their apps, isn't it?

Potentialy, but such code never worked in Safari anyway. That's why we should remove this code eventually. It applies styles very aggressively

src/hooks/private/useMatchMedia/useMatchMedia.ts
ogonkov: Why can't we use false as initial value?

ogonkov (approved):

styles/styles.scss
ogonkov: @amje  With .01 instead of .001 Safari seems working ok

PR #2334, opened 2025-07-07
feat(Accordion): add accordion component

KostyaAvtushko:


src/components/Accordion/AccordionContext.tsx
```diff
@@ -0,0 +1,117 @@
...
+    React.useEffect(() => {
+        if (controlled) {
+            setItems((prev) => {
+                if (!_isEqual(prev, value)) {
+                    onUpdate(value);
+                }
+
+                return value;
+            });
+        }
```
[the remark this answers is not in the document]
KostyaAvtushko: Should I remove onUpdate here? May be its useless

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,86 @@
...
+            value={value as AccordionValue<Multiple>}
+            controlled={'value' in props}
+            ariaLevel={ariaLevel}
+        >
+            <div
+                className={accordionBlock({size, view}, className)}
+                data-qa={qa}
+                ref={ref}
+                role="region"
+                aria-label="Accordion"
```
ogonkov: That should be in i18n, and i guess users could want to override this value

src/components/Accordion/AccordionItem/AccordionItem.scss
```diff
@@ -0,0 +1,77 @@
+@use '../../variables';
+@use '../../../../styles/mixins';
+
+$block: '.#{variables.$ns}accordion-item';
+
+@mixin accordion-item-trigger($size) {
+    .g-disclosure__trigger {
```
ogonkov: It's better to pass own classname for this element, overriding styles of other components could lead to problems with styles maintaining in future

KostyaAvtushko: Thanks! I'll remove all g-disclosure mentions from the accordion styles

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,86 @@
...
+    multiple: Multiple;
+    className: string;
+    arrowPosition: AccordionArrowPosition;
+    defaultValue: AccordionValue<Multiple>;
+    value: AccordionValue<Multiple>;
+    onUpdate: (value: AccordionValue<Multiple>) => void;
+    children: React.ReactElement<AccordionItemProps>[] | React.ReactElement<AccordionItemProps>;
+    ariaLevel: number;
+}> &
+    QAProps;
```
ogonkov: Can we move types to `Accordion/types.ts` file?

src/components/Accordion/AccordionSummary/AccordionSummary.scss
```diff
@@ -0,0 +1,40 @@
...
+    #{$block}__trigger {
+        @if $size == 's' or $size == 'm' {
+            @include mixins.text-subheader-1();
+        }
+        @if $size == 'l' or $size == 'xl' {
+            @include mixins.text-subheader-2();
+        }
+        display: flex;
+        justify-content: space-between;
+        width: 100%;
```
ogonkov: This is the common styles for trigger, i guess it would be better to set them separately from size modifiers to single rule

src/components/Accordion/__tests__/Accordion.test.tsx
korvin89, line 1: Suggestions for unit tests:
- remove all tests that are covered by screenshot testing:
  - `render accordion by default`
  - `render with given "%s" size`
  - `render accordion items with summary prop`
- remove class names usage, wherever possible, it is better to use a text search or a `qa` attribute

src/components/Accordion/types.ts
```diff
@@ -0,0 +1,26 @@
...
+
+type AccordionSize = 's' | 'm' | 'l' | 'xl';
+type AccordionView = 'solid' | 'top-bottom';
+type AccordionArrowPosition = 'start' | 'end';
+
+export type AccordionValue<Multiple extends boolean> = Multiple extends true
+    ? string[]
+    : string | undefined;
+
+export type AccordionProps<Multiple extends boolean = false> = Partial<{
```
korvin89: I suggest to remove the `Partial` here and using the `?:` notation to specify optional fields, this is how the types for the remaining components are described

src/components/Disclosure/Disclosure.tsx
```diff
@@ -40,6 +41,8 @@ export interface DisclosureProps extends QAProps {
     keepMounted?: boolean;
     /** Callback fired when the expand/collapse state is changed  */
     onUpdate?: (expanded: boolean) => void;
+    /** Callback fires on keyboard events when summary is focused */
+    onKeyDown?: (e: React.KeyboardEvent<HTMLButtonElement>) => void;
```
korvin89: Maybe `onSummaryKeyDown`?

src/components/Accordion/AccordionContext.tsx
```diff
@@ -0,0 +1,114 @@
...
+        size,
+        view,
+        multiple,
+        defaultValue,
+        onUpdate,
+        value,
+        controlled,
+    } = props;
+
+    const [items, setItems] = React.useState<AccordionValue<Multiple>>(
```
korvin89: Lets use [useControlledState](https://github.com/gravity-ui/uikit/blob/a16a44755a472bf4be4e57a6bad35db6a8e025f0/src/hooks/useControlledState/useControlledState.ts#L3) hook here

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,64 @@
...
+    return (
+        <AccordionProvider
+            size={size}
+            view={view}
+            multiple={multiple}
+            arrowPosition={arrowPosition}
+            onUpdate={onUpdate}
+            defaultValue={defaultValue as AccordionValue<Multiple>}
+            value={value as AccordionValue<Multiple>}
+            controlled={'value' in props}
```
korvin89: I think u don't need `controlled` property in case of using `useControlledState`

src/components/Accordion/AccordionContext.tsx
```diff
@@ -0,0 +1,114 @@
+import * as React from 'react';
+
+import _isEqual from 'lodash/isEqual';
```
korvin89: Lets remove underscore, it confuses typescript

src/components/Accordion/utils.ts
```diff
@@ -0,0 +1,35 @@
+import type {AccordionValue} from './types';
+
+export function updateAccordionItemsApi<Multiple extends boolean>(
```
korvin89: It looks like a complicated and unnecessary abstraction. That is, in fact, from these utilities you only need to be able to collect the next `value`, it is better to do everything else in place (in `AccordionContext.tsx`), without hiding the logic of calling callbacks in utilitarian methods. Now you have a method that collects a new `value`, calls the callback and returns this `value` after calling the callback

src/components/Accordion/AccordionItem/AccordionItem.scss
```diff
@@ -0,0 +1,59 @@
...
+        }
+        @if $size == 'xl' {
+            padding: var(--g-spacing-0) var(--g-spacing-5) var(--g-spacing-5) var(--g-spacing-5);
+        }
+    }
+}
+
+#{$block} {
+    &_disabled {
+        .g-disclosure__trigger {
```
korvin89: Lets expand `Disclosure` public interface to avoid class names usage

src/components/Accordion/AccordionItem/AccordionItem.tsx
```diff
@@ -0,0 +1,152 @@
...
+import {AccordionSummary, isAccordionSummaryComponent} from '../AccordionSummary/AccordionSummary';
+import {
+    accordionDetailsBlock,
+    accordionItemBlock,
+    accordionSummaryTriggerBlock,
+} from '../constants';
+
+import './AccordionItem.scss';
+
+export type AccordionItemProps = Partial<{
```
korvin89: Lets use `?:` notation instead of `Partial`

src/components/Accordion/AccordionItem/AccordionItem.tsx
```diff
@@ -0,0 +1,152 @@
...
+            }
+        },
+        [isControlled, isDefault, onUpdate, updateItems, value, id],
+    );
+    return {id, isExpanded, handleUpdate};
+}
+
+function prepareChildren(children: React.ReactNode, qa?: string) {
+    const items = React.Children.toArray(children);
+    let summary, details;
```
korvin89: Let's explicitly specify the expected types of variables here

src/components/Accordion/AccordionContext.tsx
```diff
@@ -0,0 +1,114 @@
...
+        >
+    > {
+    multiple: Multiple;
+    value: AccordionValue<Multiple>;
+    defaultValue: AccordionValue<Multiple>;
+    onUpdate: (value: AccordionValue<Multiple>) => void;
+    controlled: boolean;
+}
+
+const AccordionAttributesContext = React.createContext<
```
korvin89: Why is it called `Attributes`? It seems you can just leave the `AccordionContext`

src/components/Accordion/AccordionContext.tsx
```diff
@@ -0,0 +1,114 @@
...
+    );
+}
+
+export function useAccordionAttributes() {
+    const state = React.useContext(AccordionAttributesContext);
+
+    return state;
+}
+
+export function useAccordionItems() {
```
korvin89: Why do I need this hook? It's a little bit strange that you define the context from above, but in fact you only use it in `AccordionItem.tsx'. Plus, it's not clear why you should introduce a second context.

src/components/Accordion/AccordionItem/AccordionItem.tsx
```diff
@@ -0,0 +1,152 @@
...
+    value,
+    onUpdate,
+}: {
+    expanded?: boolean;
+    defaultExpanded?: boolean;
+    value?: string;
+    onUpdate?: (next: boolean) => void;
+}) {
+    const id = useUniqId();
+    const {items, updateItems} = useAccordionItems();
```
korvin89: Using `useAccordionItems` makes it harder to read the code:
- `items` are only needed to define the `IsExpanded`, this can be done outside the `AccordionItem`
- it is better to define `updateItems` either in the general context or in the `Accordion`

src/components/Accordion/AccordionItem/AccordionItem.tsx
```diff
@@ -0,0 +1,152 @@
...
+        if (isControlled) return expanded;
+        if (isDefault) return false;
+        if (Array.isArray(items)) return items.includes(value ?? id);
+        return items === (value ?? id);
+    }, [isControlled, isDefault, expanded, items, value, id]);
+
+    const handleUpdate = React.useCallback(
+        (next: boolean) => {
+            onUpdate?.(next);
+            if (!isControlled && !isDefault) {
```
korvin89: And how should it work for a controlled state?

src/components/Accordion/AccordionSummary/AccordionSummary.scss
```diff
@@ -0,0 +1,43 @@
+@use '../../variables';
+@use '../../../../styles/mixins';
+
+$block: '.#{variables.$ns}accordion-summary';
+
+@mixin accordion-item-trigger-header($size) {
+    #{$block}__trigger {
```
korvin89: Why wrap another selector with a class in a mixin? This makes it difficult to read the styles. In a good way, this should be described in the `&__trigger` block

src/components/Accordion/AccordionSummary/AccordionSummary.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+    const handleKeyDown = (e: React.KeyboardEvent<HTMLButtonElement>) => {
+        if (
+            e.key === KeyCode.ARROW_DOWN ||
+            e.key === KeyCode.ARROW_UP ||
+            e.key === KeyCode.HOME ||
+            e.key === KeyCode.END
+        ) {
+            e.preventDefault();
+            const accordionContainer = e.currentTarget.closest('.g-accordion');
```
korvin89: But here (and below) it seems to be a good place to use context - you can put links to all the necessary refs in it and avoid using the native dom api

src/components/Accordion/AccordionSummary/AccordionSummary.tsx
```diff
@@ -0,0 +1,88 @@
...
+            if (!accordionContainer) return;
```
korvin89: It's strange that eslint missed this, we always write the entire construction with curly braces

src/components/Accordion/AccordionContext.tsx
KostyaAvtushko: I copied the idea of two contexts, and hooks, from DisclosureContext. Thanks for noticing, I'll fix that

src/components/Accordion/AccordionContext.tsx
```diff
@@ -0,0 +1,148 @@
...
+    );
+
+    const summaryRefsRef = React.useRef<Map<string, AccordionSummaryRef>>(new Map());
+
+    React.useEffect(() => {
+        if (!Array.isArray(items) && multiple) {
+            setItems((items ? [items] : []) as AccordionValue<Multiple>);
+            return;
+        }
+    }, [multiple]);
```
korvin89, line 88: Lets add missing dependencies here (`items` and `setItems`)

PR #2340, opened 2025-07-12
feat(CopyToClipboard): reexport copyText function

teleginzhenya:


src/components/CopyToClipboard/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from './CopyToClipboard';
 export type {CopyToClipboardStatus, CopyToClipboardProps} from './types';
+export {copyText} from './copyText';
```
korvin89: Can you name it something more specific? Smth like `copyToClipboard`, `copyTextToClipboard` etc

teleginzhenya:
```suggestion
export {copyText as copyTextToClipboard} from './copyText';
```

ogonkov (approved):

PR #2343, opened 2025-07-14
chore(Select): add popup with actions story

qradle-yndx:
## Summary by Sourcery

Add a new CustomPopup story for the Select component that demonstrates a filterable, multi-select popup with apply and cancel buttons

New Features:
- Introduce CustomPopup story in Select.stories.tsx to showcase custom action buttons

Enhancements:
- Add CustomPopupShowcase component as a demo for the new story
- Implement .popup-actions SCSS styling for consistent button layout
- Configure the CustomPopup story to disable storybook controls

### DIFF
--- a/src/components/Select/__stories__/CustomPopupShowcase.tsx
+++ b/src/components/Select/__stories__/CustomPopupShowcase.tsx
@@ -1,0 +1,80 @@
+import * as React from 'react';
+
+import {Button} from '../../Button';
+import {Flex} from '../../layout';
+import i18n from '../i18n';
+import {Select} from '../index';
+
+const options = [
+    {value: 'val1', content: 'Value 1'},
+    {value: 'val2', content: 'Value 2'},
+    {value: 'val3', content: 'Value 3'},
+    {value: 'val4', content: 'Value 4'},
+];
+
+const noop = () => {};
+
+const SelectWithCustomPopup = ({
+    onApply,
+    onCancel,
+}: {
+    onApply?: () => void;
+    onCancel?: () => void;
+}) => {
+    const [value, setValue] = React.useState<string[]>([]);
+    const [open, onOpenChange] = React.useState(false);
+
+    const handleCancel = () => {
+        onCancel?.();
+        onOpenChange(false);
+    };
+
+    const handleApply = () => {
+        onApply?.();
+        onOpenChange(false);
+    };
+
+    return (
+        <Select
+            title="Popup with action buttons"
+            value={value}
+            options={options}
+            filterable
+            hasClear
+            className="select-example select-width-300"
+            multiple={true}
+            onUpdate={setValue}
+            onOpenChange={onOpenChange}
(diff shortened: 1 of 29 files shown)


src/components/Select/i18n/ru.json
```diff
@@ -1,5 +1,7 @@
 {
   "label_clear": "Очистить",
   "label_show-error-info": "Показать попап с информацей об ошибке",
-  "label_filter": "Фильтр"
+  "label_filter": "Фильтр",
+  "action_apply": "Применить",
+  "action_cancel": "Отмена"
```
korvin89: Let's do not add lines to the i18n component that are needed in the story, you can hardcode them in English

src/components/Select/__stories__/CustomPopupShowcase.tsx
```diff
@@ -0,0 +1,80 @@
...
+                                    {i18n('action_apply')}
+                                </Button>
+                            )}
+                        </div>
+                    </React.Fragment>
+                );
+            }}
+        />
+    );
+};
```
korvin89: Right now, your stories aren't doing the most important thing - the buttons aren't working. I suppose the idea was for the submission to be on the button, plus for the cancellation to be rolled back to the previous state?

src/components/Select/__stories__/CustomPopupShowcase.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+export const CustomPopupShowcase = () => {
```
korvin89: This showcase should have a clearer name, since the goal was to show exactly how to make Cancel/Apply buttons in Select

src/components/Select/__stories__/SelectShowcase.scss
```diff
@@ -74,3 +74,11 @@ $blockSelectClear: '.#{variables.$ns}select-clear';
         }
     }
 }
+
+.popup-actions {
```
korvin89, line 78: I suggest not making styles as a separate file, but hardcode them into jsx, so it will be much easier to copy the source code from the example

qradle-yndx: moved to style attritube

src/components/Select/i18n/ru.json
qradle-yndx: fixed

src/components/Select/__stories__/CustomPopupShowcase.tsx
qradle-yndx: not sure about the solution, made some updates, please check

src/components/Select/__stories__/CustomPopupShowcase.tsx
qradle-yndx: Renamed showcase to 'With Action Buttons'. I'll set any name with pleasure if you have better one

PR #2344, opened 2025-07-16
feat(Tooltip): add dismissOnScroll option with default true value

chrstnv:


src/components/Tooltip/Tooltip.tsx
```diff
@@ -60,13 +60,16 @@ export interface TooltipProps
     openDelay?: number;
     /** Delay in ms before close */
     closeDelay?: number;
+    /** Whether to dismiss the floating element upon scrolling an overflow ancestor. */
+    dismissOnScroll?: boolean;
```
amje: Not sure if we need to configure this. It's not a generic popup, but specific use case, tooltips.

amje (changes requested):

src/components/Tooltip/Tooltip.tsx
chrstnv: I made this parameter optional to maintain backward compatibility, since I wasn't certain that every consumer actually wants this specific behavior :)

amje: It's not backward compatible anyway, it's ON by default in this PR. If a consumer wants other behavior he can use the generic Popup component where every aspect of floating can be configured. Tooltip and ActionTooltip are not such componets. If we believe this option should be enabled in these components then it should be set like `outsidePress` setting. Of course we can add this prop later based on user feedback, but not right away.

chrstnv: Fixed!

PR #2345, opened 2025-07-17
feat: selectively handle reduced motion setting

amje:
Instead of applying very short transitions and animations for all elements, we selectively tweak the styles and code to remove the "motion" animations. The `transition-duration: 1s` was especially problematic because it enabled transitions of all properties for all elements. This caused scroll jumps for floating elements when focusing.

## Summary by Sourcery

Respect users' prefers-reduced-motion setting by selectively disabling or simplifying animations and transitions across hooks and components, and refactor the useFloatingTransition hook to handle transition callbacks internally without relying on CSS events.

New Features:
- Add media queries to disable or simplify animations in Toast, Sheet, Popup, Modal, Skeleton, Spin, and global mixins when prefers-reduced-motion is enabled
- Enable useAnimateHeight hook to respect the reduced-motion preference

Enhancements:
- Refactor useFloatingTransition to remove external transition event handlers and matchMedia dependency, using internal timers and cleanup for transition callbacks
- Update Popup and Modal components to remove onTransitionEnd props and use the updated transition status and isPositioned flag
- Add reduced-motion-specific keyframes and mixins for toast animations

Documentation:
- Update useFloatingTransition README to reflect the renamed CSS transition props

### DIFF
--- a/src/components/Modal/Modal.scss
+++ b/src/components/Modal/Modal.scss
@@ -76,4 +76,10 @@ $modal-transform-transition: transform $modal-transition-duration ease-out;
     &[data-floating-ui-status='close'] #{$block}__content {
         transition: $modal-height-transition, $modal-transform-transition;
     }
+
+    @media (prefers-reduced-motion: reduce) {
+        &[data-floating-ui-status] #{$block}__content {
+            transition: none;
+        }
+    }
 }
--- a/src/components/Sheet/Sheet.scss
+++ b/src/components/Sheet/Sheet.scss
@@ -39,6 +39,14 @@ $block: '.#{variables.$ns}sheet';
         &_with-transition {
             transition: transform $transition-duration ease;
         }
+
+        @media (prefers-reduced-motion: reduce) {
+            opacity: 0;
+
+            &_with-transition {
+                transition: opacity $transition-duration ease;
+            }
+        }
     }
 
     &__sheet-swipe-area {
@@ -81,6 +89,10 @@ $block: '.#{variables.$ns}sheet';
 
         &_without-scroll {
             overflow: hidden;
+        }
+
+        @media (prefers-reduced-motion: reduce) {
+            transition-duration: 1ms;
         }
     }
 
--- a/src/components/Skeleton/Skeleton.scss
+++ b/src/components/Skeleton/Skeleton.scss
@@ -27,6 +27,10 @@ $block: '.#{variables.$ns}skeleton';
                 var(--g-color-base-generic)
             );
             animation: #{variables.$ns}skeleton 1.2s ease-out infinite;
+
+            @media (prefers-reduced-motion: reduce) {
+                animation: none;
+            }
         }
     }
 
--- a/src/components/Spin/Spin.scss
+++ b/src/components/Spin/Spin.scss
@@ -44,6 +44,10 @@ $block: '.#{variables.$ns}spin';
             height: 36px;
         }
     }
+
+    @media (prefers-reduced-motion: reduce) {
+        animation: none;
+    }
 }
 
 @keyframes #{variables.$ns}spin {
--- a/src/components/Spin/Spin.scss
+++ b/src/components/Spin/Spin.scss
@@ -7,6 +7,10 @@ $block: '.#{variables.$ns}spin';
     backface-visibility: hidden;
     animation: #{variables.$ns}spin 1s linear infinite;
 
+    @media (prefers-reduced-motion: reduce) {
+        animation: none;
+    }
+
     &__inner {
         box-sizing: border-box;
         height: 100%;
@@ -43,10 +47,6 @@ $block: '.#{variables.$ns}spin';
             width: 36px;
             height: 36px;
         }
-    }
-
-    @media (prefers-reduced-motion: reduce) {
-        animation: none;
     }
 }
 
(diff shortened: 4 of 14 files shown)


src/components/Modal/Modal.scss
```diff
@@ -76,4 +76,10 @@ $modal-transform-transition: transform $modal-transition-duration ease-out;
     &[data-floating-ui-status='close'] #{$block}__content {
         transition: $modal-height-transition, $modal-transform-transition;
     }
```
ogonkov, line 78: Can we use same selectors block, to reduce selectors duplication?

```suggestion
    &[data-floating-ui-status='close'] #{$block}__content {
        transition: $modal-height-transition, $modal-transform-transition;
        
        @media (prefers-reduced-motion: reduce) {
            transition: none;
        }
    }
```

src/components/Sheet/Sheet.scss
```diff
@@ -39,6 +39,14 @@ $block: '.#{variables.$ns}sheet';
         &_with-transition {
             transition: transform $transition-duration ease;
         }
+
+        @media (prefers-reduced-motion: reduce) {
+            opacity: 0;
+
+            &_with-transition {
+                transition: opacity $transition-duration ease;
```
ogonkov, line 47: Why we animate opacity instead?

src/components/Sheet/Sheet.scss
```diff
@@ -82,6 +90,10 @@ $block: '.#{variables.$ns}sheet';
         &_without-scroll {
             overflow: hidden;
         }
+
+        @media (prefers-reduced-motion: reduce) {
+            transition-duration: 1ms;
+        }
```
ogonkov, line 96: I guess that block should start at line 89, because we override `&__sheet-scroll-container` styles

src/components/Skeleton/Skeleton.scss
```diff
@@ -27,6 +27,10 @@ $block: '.#{variables.$ns}skeleton';
                 var(--g-color-base-generic)
             );
             animation: #{variables.$ns}skeleton 1.2s ease-out infinite;
+
+            @media (prefers-reduced-motion: reduce) {
+                animation: none;
```
ogonkov, line 32: I think we need to set minimal animation duration instead, to not break some code that depends on it

src/hooks/private/useFloatingTransition/useFloatingTransition.ts
```diff
@@ -55,6 +56,12 @@ export function useFloatingTransition({
     React.useEffect(() => {
         if (status === 'open' && previousStatus === 'initial') {
             onTransitionIn?.();
+
+            if (!enabled) {
+                requestAnimationFrame(() => {
```
ogonkov: You must clear this effect

ogonkov (changes requested):

src/hooks/private/useFloatingTransition/useFloatingTransition.ts
amje: What do you mean by clearing the effect, cancelling the frame?

ogonkov: Yes

src/components/Skeleton/Skeleton.scss
amje: Cannot imagine the code that would depend on an infinite animation

src/components/Sheet/Sheet.scss
amje: To match the veil animation

src/components/Skeleton/Skeleton.scss
ogonkov: Oh, you right

src/components/Modal/Modal.scss
amje: Then this code must be repeated 4 times

src/components/Sheet/Sheet.scss
amje: `&_without-scroll` is also a mod for `sheet-scroll-container`. So there is no difference in order, both rules are applied to the same block, and not interfere with each other

src/hooks/private/useFloatingTransition/useFloatingTransition.ts
amje: In what scenario this needed? I `enabled` toggles, none of these ifs will be true, so no code executed

ogonkov: To clear rAF on component unmount

amje: You suggest not to notify the user about transition completion if component was unmounted. I have the opposite opinion, if `onTransitionIn` called, then `onTransitionInComplete` must be called as well no matter what

ogonkov: But component is unmounted, there is nobody to notify

amje: Is this even possible? That the next frame there will be no component?

ogonkov: I just follow react docs guidance here

In strict mode this effect would be called twice, and i think this rAF would be called twice as result

src/components/Spin/Spin.scss
```diff
@@ -44,6 +44,10 @@ $block: '.#{variables.$ns}spin';
             height: 36px;
         }
     }
+
+    @media (prefers-reduced-motion: reduce) {
+        animation: none;
+    }
```
ogonkov: Move it to line 10 pls, before `inner` element styles

amje:
@sourcery-ai dismiss

amje:
@sourcery-ai review

src/components/Spin/Spin.scss
amje: done

ogonkov (approved):

mournfulCoroner (approved):

PR #2350, opened 2025-07-25
feat: move Drawer from navigation to uikit

igor0226:
## Summary by Sourcery

Move the Drawer component from navigation into the UIKit package, introducing a fully-featured sliding panel with animations, resize support, and improved transition tracking, alongside updates to testing fixtures and documentation.

New Features:
- Add a Drawer component with slide-in panel supporting four directions, hide veil, and disable portal options
- Introduce resizable functionality for Drawer with size, minSize, maxSize, and resize callbacks
- Expose isTransitionInProgress flag in useFloatingTransition hook for animation state tracking
- Extend Playwright mount fixture to support height option and add a defaultDelay fixture for consistent test timing

Enhancements:
- Export Drawer from the components index
- Implement useDrawerFloating and useResizableDrawerItem utilities to manage Drawer behavior

Documentation:
- Add English and Russian READMEs and Storybook stories for the new Drawer component

Tests:
- Add unit tests and Playwright stories/helpers for Drawer component

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [7.27.0](https://github.com/gravity-ui/uikit/compare/v7.26.2...v7.27.0) (2025-12-04)
+
+
+### Features
+
+* added a hook for generating a unique color with oklch color space ([#2288](https://github.com/gravity-ui/uikit/issues/2288)) ([96568a8](https://github.com/gravity-ui/uikit/commit/96568a80795de6816abab7f588afd8ef662c4d26))
+
 ## [7.26.2](https://github.com/gravity-ui/uikit/compare/v7.26.1...v7.26.2) (2025-12-01)
 
 
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -65,6 +65,8 @@
 /src/components/Tooltip @amje
 /src/components/layout @IsaevAlexandr
 
+/src/components/lab/ColorPicker @Estasie
+
 /src/hooks/useActionHandlers @ogonkov
 /src/hooks/useFileInput @korvin89
 /src/hooks/useForkRef @ValeraS
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@gravity-ui/uikit",
-  "version": "7.26.2",
+  "version": "7.27.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@gravity-ui/uikit",
-      "version": "7.26.2",
+      "version": "7.27.0",
       "license": "MIT",
       "dependencies": {
         "@bem-react/classname": "^1.7.0",
@@ -15,6 +15,7 @@
         "@gravity-ui/icons": "^2.16.0",
         "@hello-pangea/dnd": "^18.0.1",
         "@tanstack/react-virtual": "^3.13.12",
+        "@uiw/react-color": "^2.9.2",
         "blueimp-md5": "^2.19.0",
         "lodash": "^4.17.21",
         "rc-slider": "^11.1.9",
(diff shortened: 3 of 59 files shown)


src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,326 @@
...
+        React.useEffect(() => {
+            setInitialRender(true);
+        }, [direction]);
+
+        const resizerElement = resizable ? (
+            <div className={b('resizer', {direction})} {...resizerHandlers}>
+                <div className={b('resizer-handle', {direction})} />
+            </div>
+        ) : null;
+
```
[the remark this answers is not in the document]
igor0226: To mitigate this issue I've added a simple ternary construction with `visible && resizable` check

igor0226:
@sourcery-ai review

igor0226:
@sourcery-ai review

igor0226:
@sourcery-ai review

src/components/Drawer/README-ru.md
```diff
@@ -0,0 +1,93 @@
...
+| showInitialAnimation  | Включает анимацию при первом открытии, если Drawer рендерится в открытом состоянии |       `boolean`       |        `false`        |
+| onOpenChange          | Обработчик, вызываемый при изменении состояния видимости                           |      `Function`       |                       |
+| onTransitionEnter     | Обработчик начала анимации открытия                                                |      `Function`       |                       |
+| onTransitionExit      | Обработчик начала анимации закрытия                                                |      `Function`       |                       |
+| onTransitionEntered   | Обработчик завершения анимации открытия                                            |      `Function`       |                       |
+| onTransitionExited    | Обработчик завершения анимации закрытия                                            |      `Function`       |                       |
+| onResizeEnd           | Колбэк, вызываемый по окончании изменения размера                                  |      `Function`       |                       |
```
amje: `Обработчик завершения изменения размера` и для остальных в таком же стиле

src/components/Drawer/README-ru.md
```diff
@@ -0,0 +1,93 @@
...
+| onResize              | Колбэк, вызываемый во время изменения размера                                      |      `Function`       |                       |
+| onResizeStart         | Колбэк, вызываемый в начале изменения размера                                      |      `Function`       |                       |
+| maxSize               | Максимальная ширина контента в пикселях                                            |       `number`        |                       |
+| minSize               | Минимальная ширина контента в пикселях                                             |       `number`        |                       |
+| size                  | Ширина контента в пикселях                                                         |       `number`        |                       |
+
+## CSS API
+
+| Имя                                          | Описание                                      |          Значение по умолчанию          |
```
amje: CSS API is not for reading values so default column is not suitable here

src/components/Drawer/__snapshots__/Drawer.visual.test.tsx-snapshots/Drawer-render-story-DisablePortal-dark-chromium-linux.png
amje, line 1: This is a complex story with controls. Better to write dedicated tests for the component and not use story for that.

src/components/Drawer/__stories__/DisablePortal.scss
```diff
@@ -0,0 +1,60 @@
+.disable-portal {
+    --g-drawer-item-position: absolute;
+
+    width: 100dvw;
+    height: 100dvh;
```
amje: Why we needed such size?

amje:
Снимок экрана 2025-08-26 в 18 36 30

Content is not covering all available height. Also the scroll is visible through the resizer.

src/components/Drawer/components/Drawer.tsx
```diff
@@ -0,0 +1,232 @@
...
+    /**
+     * Callback called when the drawer is being resized.
+     */
+    onResize?: OnResizeHandler;
+    /**
+     * Callback called at the end of resizing.
+     */
+    onResizeEnd?: OnResizeHandler;
+    /** Optional additional class names to style the background veil element. */
+    veilClassName?: string;
```
amje: Not sure why we need to pass this. What styles are we expected to get from the user here?

src/components/Drawer/components/Drawer.tsx
```diff
@@ -0,0 +1,232 @@
...
+    /** Optional flag to hide the background darkening */
+    hideVeil?: boolean;
+    /**
+     * Option that enables first opening animation if the Drawer is being rendered with open state.
+     * @default false
+     */
+    showInitialAnimation?: boolean;
```
amje: Why we disable this in the frist place?

src/components/Drawer/components/Drawer.tsx
```diff
@@ -0,0 +1,232 @@
...
+                    data-floating-ui-status={status}
+                    data-transiting={isTransitionInProgress}
+                    lockScroll={!disableBodyScrollLock}
+                >
+                    <div
+                        ref={veilRef}
+                        className={b('veil', {hidden: hideVeil}, veilClassName)}
+                        role="presentation"
+                        onClick={handleVeilClick}
+                    />
```
amje: Why it's not like in the Modal, where veil is the root node? Seems it would be cleaner

src/components/Drawer/components/DrawerItem.tsx
```diff
@@ -0,0 +1,144 @@
...
+const b = block('drawer');
+
+import './Drawer.scss';
+
+export interface DrawerItemProps {
+    /** Content to be displayed within the drawer item. */
+    children?: React.ReactNode;
+
+    /** Determines whether the drawer item is visible or hidden. */
+    visible?: boolean;
```
amje:
```suggestion
    open?: boolean;
```

src/components/Drawer/components/DrawerItem.tsx
```diff
@@ -0,0 +1,144 @@
...
+    className?: string;
+
+    /** Determines whether the drawer item can be resized */
+    resizable?: boolean;
+
+    /**
+     * The width of the resizable drawer item.
+     * If not provided, the width will be stored internally.
+     */
+    width?: number;
```
amje:
```suggestion
    size?: number;
```

src/components/Drawer/components/DrawerItem.tsx
```diff
@@ -0,0 +1,144 @@
...
+            style = {},
+            disablePortal,
+            ...rest
+        } = props;
+
+        const [isInitialRender, setInitialRender] = React.useState(true);
+        const itemRef = React.useRef<HTMLDivElement>(null);
+        const handleRef = useForkRef(ref, itemRef);
+
+        const cssDirection = direction === 'left' ? undefined : direction;
```
amje: We can use only `direction`. Don't need 2nd variable

src/components/Drawer/components/DrawerItem.tsx
```diff
@@ -0,0 +1,144 @@
...
+            if (resizable) {
+                if (['left', 'right'].includes(direction)) {
+                    css.width = `${resizedWidth}px`;
+                } else {
+                    css.height = `${resizedWidth}px`;
+                }
+            }
+
+            return css;
+        }, [direction, resizable, resizedWidth, style]);
```
amje: Memoization not working here, style is default new object every render. Moreover it's not needed at all. Nothing to optimize

src/components/Drawer/components/DrawerItem.tsx
```diff
@@ -0,0 +1,144 @@
...
+            maxResizeWidth,
+            onResizeStart,
+            onResizeEnd,
+            onResize,
+            style = {},
+            disablePortal,
+            ...rest
+        } = props;
+
+        const [isInitialRender, setInitialRender] = React.useState(true);
```
amje: State doesn't update anywhere

src/components/Drawer/components/Drawer.scss
```diff
@@ -0,0 +1,199 @@
...
+            inset-inline-end: 0;
+        }
+
+        &_direction_bottom {
+            inset-block-start: auto;
+            inset-inline-end: 0;
+        }
+
+        &_disable-portal {
+            --_--item-position: static;
```
amje: `disablePortal` should behave like in `Modal` component. It shouldn't render in place

src/components/Drawer/hooks/useResizeHandlers.ts
```diff
@@ -0,0 +1,60 @@
...
+        },
+        [onStart, direction, handleMove, handleUp],
+    );
+
+    React.useEffect(() => {
+        return () => {
+            window.removeEventListener('pointermove', handleMove);
+            window.removeEventListener('pointerup', handleUp);
+        };
+    }, [handleMove, handleUp]);
```
amje, line 58: It's better to remove all listeners on "unmount effect" (without deps). And store `move` and `up` linsteners in ref. Thus the component will not notify if it's been unmounted while resizing

src/components/Drawer/components/DrawerItem.tsx
igor0226: it is being updated inside a useEffect hook, 20 lines below

src/components/Drawer/components/Drawer.tsx
igor0226: I thought it could be handy in case user wants to load a page with a drawer opened be default

src/components/Drawer/components/Drawer.tsx
igor0226: It absolutely would be but in this particular component we need to change the veil's opacity independently from the item's opacity. In the Modal component both opacity values are being changed simultaneously, which is expected behavior.

src/hooks/private/useFloatingTransition/useFloatingTransition.ts
```diff
@@ -27,6 +28,7 @@ export function useFloatingTransition({
     onTransitionOut,
     onTransitionOutComplete,
 }: UseFloatingTransitionProps): UseFloatingTransitionResult {
+    const [isTransitionInProgress, setIsTransitionInProgress] = React.useState(false);
```
korvin89: You can subscribe on this event `onTransitionIn`, `onTransitionInComplete`, `onTransitionOut`, `onTransitionOutComplete`, there is n need in such interface extention

src/components/Drawer/README.md
```diff
@@ -0,0 +1,92 @@
...
+| onResizeStart         | Callback called at the start of resizing                                                    |      `Function`       |                 |
+| maxSize               | Max width of the drawer in pixels                                                           |       `number`        |                 |
+| minSize               | Min width of the drawer in pixels                                                           |       `number`        |                 |
+| size                  | Width of the drawer in pixels                                                               |       `number`        |                 |
+
+## CSS API
+
+| Name                                         | Description                                                 |
+| :------------------------------------------- | :---------------------------------------------------------- |
+| DrawerItem                                   |                                                             |
```
korvin89: Let's divide it into several tables for each element, so it will be easier to read

src/components/Drawer/__tests__/helpersPlaywright.ts
```diff
@@ -0,0 +1,5 @@
+import {composeStories} from '@storybook/react';
+
+import * as DefaultDrawerStories from '../__stories__/Drawer.stories';
+
+export const DrawerStories = composeStories(DefaultDrawerStories);
```
korvin89: Let's separate the stories and the tests so that they don't depend on each other. The minimum information that the test verifies should be added to the test cases. This will make the tests more readable and the stories more flexible to change

playwright/core/defaultDelayFixture.ts
```diff
@@ -0,0 +1,9 @@
+import type {PlaywrightFixture} from './types';
+
+const DEFAULT_MOUNT_TEST_DELAY_MS = 1000;
+
+export const defaultDelayFixture: PlaywrightFixture<() => Promise<void>> = async ({page}, use) => {
+    const defaultDelay = async () => await page.waitForTimeout(DEFAULT_MOUNT_TEST_DELAY_MS);
```
korvin89: You can do without delay. It is better to rely on the `data-floating-ui-status` data attribute - this solution will be more stable, since the current solution is tied to limited timings

src/components/Drawer/__stories__/HideVeil.scss
korvin89, line 1: Why is there a fairly large set of additional styles used for each story?

src/components/Drawer/README.md
```diff
@@ -0,0 +1,92 @@
...
+| disableBodyScrollLock | Disables locking scroll while open                                                          |       `boolean`       |     `false`     |
+| disableEscapeKeyDown  | Disables triggering close on `Esc`                                                          |       `boolean`       |     `false`     |
+| disableOutsideClick   | Disables triggering close on outside clicks                                                 |       `boolean`       |     `false`     |
+| disablePortal         | Disables using `Portal`                                                                     |       `boolean`       |     `false`     |
+| keepMounted           | `Drawer` will not be removed from the DOM upon hiding                                       |       `boolean`       |     `false`     |
+| resizable             | Enables resizing of the drawer via pointer                                                  |       `boolean`       |     `false`     |
+| open                  | Manages `Drawer` visibility                                                                 |       `boolean`       |     `false`     |
```
korvin89: The component should be able to work with a non-controllable state. For this, the `defaultOpen` setting is needed

src/components/Drawer/README.md
```diff
@@ -0,0 +1,92 @@
...
+| direction             | Specifies the direction from which the drawer should slide in                               |       `string`        |     `left`      |
+| contentClassName      | `class` HTML attribute for the content node                                                 |       `string`        |                 |
+| restoreFocusRef       | Element the focus will be restored to                                                       |   `React.RefObject`   |                 |
+| children              | Any React content                                                                           |   `React.ReactNode`   |                 |
+| container             | DOM element to which component is mounted via `Portal`                                      |     `HTMLElement`     | `document.body` |
+| hideVeil              | Optional flag to hide the background darkening                                              |       `boolean`       |                 |
+| autoFocus             | While open, the focus will be set to the first interactive element in the content           |       `boolean`       |     `true`      |
```
korvin89: This prop is not in the code, moreover, it is most likely not needed in this component

src/components/Drawer/README.md
```diff
@@ -0,0 +1,92 @@
...
+| qa                    | Test attribute (`data-qa`)                                                                  |       `string`        |                 |
+| style                 | `style` HTML attribute for the root node                                                    | `React.CSSProperties` |                 |
+| aria-label            | `aria-label` HTML attribute to describe `Drawer`                                            |       `string`        |                 |
+| aria-labelledby       | ID of the visible `Drawer` caption element                                                  |       `string`        |                 |
+| direction             | Specifies the direction from which the drawer should slide in                               |       `string`        |     `left`      |
+| contentClassName      | `class` HTML attribute for the content node                                                 |       `string`        |                 |
+| restoreFocusRef       | Element the focus will be restored to                                                       |   `React.RefObject`   |                 |
```
korvin89: This prop is not used in the code

src/components/Drawer/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Properties
+
+| Name                  | Description                                                                                 |         Type          |     Default     |
+| :-------------------- | :------------------------------------------------------------------------------------------ | :-------------------: | :-------------: |
+| className             | `class` HTML attribute for the root node                                                    |       `string`        |                 |
+| qa                    | Test attribute (`data-qa`)                                                                  |       `string`        |                 |
+| style                 | `style` HTML attribute for the root node                                                    | `React.CSSProperties` |                 |
+| aria-label            | `aria-label` HTML attribute to describe `Drawer`                                            |       `string`        |                 |
+| aria-labelledby       | ID of the visible `Drawer` caption element                                                  |       `string`        |                 |
```
korvin89: Aria attributes are not used in the code

src/components/Drawer/README.md
```diff
@@ -0,0 +1,92 @@
...
+| contentClassName      | `class` HTML attribute for the content node                                                 |       `string`        |                 |
+| restoreFocusRef       | Element the focus will be restored to                                                       |   `React.RefObject`   |                 |
+| children              | Any React content                                                                           |   `React.ReactNode`   |                 |
+| container             | DOM element to which component is mounted via `Portal`                                      |     `HTMLElement`     | `document.body` |
+| hideVeil              | Optional flag to hide the background darkening                                              |       `boolean`       |                 |
+| autoFocus             | While open, the focus will be set to the first interactive element in the content           |       `boolean`       |     `true`      |
+| focusTrap             | Enables focus trapping behavior                                                             |       `boolean`       |     `true`      |
```
korvin89: This prop is not used in the code

src/components/Drawer/components/Drawer.tsx
```diff
@@ -0,0 +1,227 @@
...
+import i18n from '../i18n';
+
+import {DrawerItem} from './DrawerItem';
+import type {DrawerItemProps} from './DrawerItem';
+
+import './Drawer.scss';
+
+const b = block('drawer');
+
+export type DrawerChild = React.ReactElement<DrawerItemProps>;
```
korvin89: Why do we need this type?

src/components/Drawer/components/DrawerItem.tsx
```diff
@@ -0,0 +1,132 @@
...
+            onResizeEnd,
+            onResize,
+        });
+
+        const isVerticalDirection = ['left', 'right'].includes(direction);
+        const isHorizontalDirection = !isVerticalDirection;
+
+        React.useEffect(() => {
+            setIsInitialRender(true);
+        }, [direction]);
```
korvin89: Why `direction` is here?

src/components/Drawer/README.md
igor0226: Now I see - I copied these props from `Modal`. It seems that neither `restoreFocusRef` nor `focusTrap` are in use in Modal component. Considering that, may be it makes sense to get rid of them in Modal too?

src/components/Drawer/components/Drawer.scss
```diff
@@ -0,0 +1,211 @@
+@use '../../variables';
+
+$block: '.#{variables.$ns}drawer';
+
+#{$block} {
+    pointer-events: none;
+    outline: none;
+
+    --_--item-shadow-default: 0 1px 5px 0 var(--g-color-sfx-shadow);
+    --_--item-position: fixed;
```
amje: Is it needed to override the position style? Modal doesn't have one. If so it should be a prop to pass to the `floating-ui` hook anyway, not CSS API

src/components/Drawer/components/Drawer.scss
```diff
@@ -0,0 +1,211 @@
...
+        &_direction_bottom {
+            width: 28px;
+            height: 2px;
+        }
+    }
+
+    &__resizer {
+        position: absolute;
+        inset-block-start: 0;
+        z-index: var(--g-drawer-item-resizer-z-index, var(--_--resizer-z-index));
```
amje: This should not be configurable. Set it to the same value as in Modal

src/components/Drawer/components/Drawer.scss
```diff
@@ -0,0 +1,211 @@
...
+    --_--resizer-handle-color: var(--g-color-line-generic);
+    --_--resizer-handle-color-hover: var(--g-color-line-generic-hover);
+    --_--resizer-z-index: 100;
+
+    &_hide-veil {
+        --_--item-shadow: var(--g-drawer-item-shadow, var(--_--item-shadow-default));
+    }
+
+    &__item {
+        position: var(--g-drawer-item-position, var(--_--item-position));
```
amje: `item` name doesn't need to be included in the names. Drawer is the item itself

src/components/Drawer/components/Drawer.scss
```diff
@@ -0,0 +1,211 @@
...
+        z-index: var(--g-drawer-item-z-index, 110);
+        inset-inline-start: var(--g-drawer-item-left-offset, 0);
+        inset-block: var(--g-drawer-item-top-offset, 0) 0;
```
amje: What are these variables for? We have direction prop which overrides them eventually. Not sure if we want it to be configurable by user at all. Let's keep it simple for now and remove them

src/components/Drawer/components/Drawer.scss
```diff
@@ -0,0 +1,211 @@
...
+
+    &[data-transiting='false'] &__item {
+        // Hack for some browsers so text won't be blurred after animation
+        filter: blur(0px);
+        transform: translateZ(0);
+    }
+
+    /* left */
+    &[data-floating-ui-status='initial'] &__item,
+    &[data-floating-ui-status='close'] &__item {
```
amje: Let's be explicit and always set modifier for direction

src/components/Drawer/components/Drawer.scss
```diff
@@ -0,0 +1,211 @@
...
+        &_direction_bottom {
+            inset-inline-start: 0;
+            inset-block-start: 0;
+        }
+
+        &_direction_top,
+        &_direction_bottom {
+            cursor: row-resize;
+            width: 100%;
+            height: var(--g-drawer-item-resizer-width, var(--_--resizer-width));
```
amje: If name used for both width and height, then it should be called like `*-size`, otherwise it's confusing

src/components/Drawer/components/Drawer.tsx
amje: Ok, but why it's disabled on initial opening? It should ON I guess

src/components/Drawer/components/Drawer.tsx
```diff
@@ -0,0 +1,216 @@
...
+    /**
+     * Option that enables first opening animation if the Drawer is being rendered with open state.
+     * @default false
+     */
+    showInitialAnimation?: boolean;
+}
+
+export const Drawer = ({
+    open,
+    defaultOpen = false,
```
amje: Uncontrolled state not usable for open state. What's the purpose of having it?

src/components/Drawer/components/Drawer.tsx
```diff
@@ -0,0 +1,216 @@
...
+                            className={contentClassName}
+                            resizable={resizable}
+                            size={size}
+                            onResize={onResize}
+                            onResizeStart={onResizeStart}
+                            onResizeEnd={onResizeEnd}
+                            minResizeWidth={minSize}
+                            maxResizeWidth={maxSize}
+                            {...getFloatingProps()}
+                            {...restProps}
```
amje: We should wrap it into `filterDOMProps`. See Modal [implementaion](https://github.com/gravity-ui/uikit/blob/main/src/components/Modal/Modal.tsx#L282)

src/components/Drawer/components/Drawer.tsx
```diff
@@ -0,0 +1,216 @@
...
+                            'skip-animation': !showInitialAnimation && isInitialRender,
+                        },
+                        className,
+                    )}
+                    data-qa={qa}
+                    data-floating-ui-status={status}
+                    data-transiting={isTransitionInProgress}
+                    lockScroll={!disableBodyScrollLock}
+                >
+                    <div className={b('veil', {hidden: hideVeil})} />
```
amje: Why veil is separate from content? FloatingOverlay can be used as a veil. Let's do the same as we did in Modal component. These two components should have almost exact implementation

src/components/Drawer/components/Drawer.tsx
```diff
@@ -0,0 +1,216 @@
...
+    const {getFloatingProps} = useInteractions([dismiss, role]);
+    const handleFloatingRef = useForkRef<HTMLDivElement>(refs.setFloating, floatingRef);
+
+    const portal =
+        isMounted || keepMounted ? (
+            <Portal container={container} disablePortal={disablePortal}>
+                <FloatingOverlay
+                    style={style}
+                    ref={overlayRef}
+                    aria-modal="true"
```
amje: There is no need for that. Floating UI handles screenreaders navitation by itself. Moreover if disablePortal is true, this node is always presented in the DOM, and will be considered as always open modal.

src/components/Drawer/components/DrawerItem.tsx
```diff
@@ -0,0 +1,135 @@
...
+            onResizeStart,
+            onResizeEnd,
+            onResize,
+            style = {},
+            ...restProps
+        } = props;
+
+        const [isInitialRender, setIsInitialRender] = React.useState(true);
+        const itemRef = React.useRef<HTMLDivElement>(null);
+        const handleRef = useForkRef(ref, itemRef);
```
amje: `itemRef` is not used anywhere in the file.

src/components/Drawer/components/DrawerItem.tsx
```diff
@@ -0,0 +1,135 @@
...
+            onResizeStart,
+            onResizeEnd,
+            onResize,
+        });
+
+        const isVerticalDirection = ['left', 'right'].includes(direction);
+        const isHorizontalDirection = !isVerticalDirection;
+
+        React.useEffect(() => {
+            setIsInitialRender(true);
```
amje: That's not valid, effect is happening after render. So its actually after initial render.

src/components/Drawer/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const DRAWER_ANIMATION_DURATION_MS = 300;
+export const DRAWER_ITEM_MIN_RESIZE_WIDTH = 200;
+export const DRAWER_ITEM_MAX_RESIZE_WIDTH = 800;
+export const DRAWER_ITEM_INITIAL_RESIZE_WIDTH = 400;
```
amje: RESIZE_WIDTH => SIZE

src/components/Drawer/components/Drawer.scss
```diff
@@ -0,0 +1,211 @@
...
+        transform: translate(0, 0);
+    }
+
+    &:not(&_skip-animation)[data-floating-ui-status='open'] &__item,
+    &:not(&_skip-animation)[data-floating-ui-status='close'] &__item {
+        transition: transform 300ms;
+    }
+
+    &[data-transiting='false'] &__item {
+        // Hack for some browsers so text won't be blurred after animation
```
amje: Could you please explain what problem are we fixing here? Is this relevant for our other components with transitions?

src/components/Drawer/hooks/useResizableDrawerItem.ts
```diff
@@ -0,0 +1,97 @@
...
+    onResizeStart?: OnResizeHandler;
+    onResizeEnd?: OnResizeHandler;
+    onResize?: OnResizeHandler;
+}
+
+export function useResizableDrawerItem(params: UseResizableDrawerItemParams) {
+    const {
+        direction = 'left',
+        size,
+        minResizeWidth = DRAWER_ITEM_MIN_RESIZE_WIDTH,
```
amje: width => size

src/components/Drawer/components/DrawerItem.tsx
igor0226: It is used for handling `floatingRef` prop

src/components/Drawer/components/Drawer.scss
igor0226: Initially, I copied this from the navigation package
Later, I did some research and it seems like we don't need this hack anymore

src/components/Drawer/components/Drawer.tsx
igor0226: These props are already wrapped inside of `DrawerItem`

src/components/Drawer/components/Drawer.tsx
igor0226: ultimately, fixed this one

src/components/Drawer/components/Drawer.tsx
igor0226: Unfortunately, in case of Drawer we cannot use FloatingOverlay as a veil, because we need to change the veil's opacity independently from the Item (content)

src/components/Drawer/components/Drawer.tsx
korvin89: Why it is not usable? What if user don't wanna handle the state? Select and Menu [use](https://github.com/search?q=repo%3Agravity-ui%2Fuikit%20defaultOpen&type=code) this pattern for example

amje: Select and Menu both have triggers, Drawer doesn't. After closing Select or Menu you can easily reopen it without controlling the state. For Drawer it won't work: if you close it, the only way to reopen is to set open={true}

src/components/Drawer/components/Drawer.tsx
```diff
@@ -0,0 +1,214 @@
...
+    /**
+     * Specifies the default open state of the component.
+     * @default false
+     */
+    defaultOpen?: boolean;
+    /**
+     * Specifies the direction from which the drawer should slide in, `left` by default.
+     * @default left
+     */
+    direction?: DrawerDirection;
```
amje: Let's rename it to `placement`. I think it's more accurate name for that. `direction` confuses a bit, the name is used for content alignment across the library

src/components/Drawer/components/Drawer.tsx
```diff
@@ -0,0 +1,214 @@
...
+    /**
+     * Callback called when the drawer is being resized.
+     */
+    onResize?: OnResizeHandler;
+    /**
+     * Callback called at the end of resizing.
+     */
+    onResizeEnd?: OnResizeHandler;
+    /** Optional flag to hide the background darkening */
+    hideVeil?: boolean;
```
amje: Are we really need a prop for this? The same is achievable via `--g-drawer-veil-background-color: transparent`. But I think the CSS API is also unnecessary. Veil color is controllable by theme token `--g-color-sfx-veil` for all the components.

src/components/Drawer/components/Drawer.tsx
```diff
@@ -0,0 +1,214 @@
...
+                            direction={direction}
+                            className={contentClassName}
+                            resizable={resizable}
+                            size={size}
+                            onResize={onResize}
+                            onResizeStart={onResizeStart}
+                            onResizeEnd={onResizeEnd}
+                            minSize={minSize}
+                            maxSize={maxSize}
+                            {...getFloatingProps()}
```
amje, line 190: This props are ignored inside DrawerItem.

src/components/Drawer/components/Drawer.tsx
amje: > need to change the veil's opacity independently from the Item

What's the restriction? Why we can't use the same animations/transitions for both Modal and Drawer?

src/components/Drawer/components/Drawer.scss
```diff
@@ -0,0 +1,187 @@
...
+$block: '.#{variables.$ns}drawer';
+$drawer-resizer-width: 8px;
+
+#{$block} {
+    pointer-events: none;
+    outline: none;
+
+    &__item {
+        position: absolute;
+        z-index: 110;
```
amje: Let's keep it high enough. It should be equal or maybe slightly less than Modal's zIndex

src/components/Drawer/components/DrawerItem.tsx
amje: Where? In this file it is the only place of using itemRef

src/components/Drawer/components/DrawerItem.tsx
```diff
@@ -0,0 +1,130 @@
...
+        });
+
+        const isVerticalDirection = ['left', 'right'].includes(direction);
+        const isHorizontalDirection = !isVerticalDirection;
+
+        const resizerElement = resizable ? (
+            <div
+                className={b('resizer', {direction})}
+                onPointerDown={open && resizable ? onResizerPointerDown : undefined}
+                role="presentation"
```
amje: What's the purpose of this? There is not semantics on element

src/components/Drawer/components/DrawerItem.tsx
```diff
@@ -0,0 +1,130 @@
...
+            <div
+                ref={handleRef}
+                className={b(
+                    'item',
+                    {
+                        direction,
+                        resizable,
+                        hidden: !open,
+                    },
+                    [className],
```
amje: Unnecessary array wrap

src/components/Drawer/components/DrawerItem.tsx
```diff
@@ -0,0 +1,130 @@
...
+            onResizeEnd,
+            onResize,
+        });
+
+        const isVerticalDirection = ['left', 'right'].includes(direction);
+        const isHorizontalDirection = !isVerticalDirection;
+
+        const resizerElement = resizable ? (
+            <div
+                className={b('resizer', {direction})}
```
amje: You don't need to use `direction` mod for all elements. Keep it only for root element and use CSS cascading

src/components/Drawer/hooks/useDrawerFloating.ts
```diff
@@ -0,0 +1,56 @@
...
+
+type UseDrawerFloatingParams = {
+    onTransitionIn?: () => void;
+    onTransitionInComplete?: () => void;
+    onTransitionOut?: () => void;
+    onTransitionOutComplete?: () => void;
+    context: FloatingContext;
+};
+
+export function useDrawerFloating({
```
amje: `useDrawerFloatingTransition` should work better. The hook is not about floating logic

src/components/Drawer/__tests__/Drawer.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+        expect(screen.getByTestId(qa)).toHaveClass('g-custom-drawer');
+    });
+    test('should keep component mounted if keepMounted is passed', () => {
+        render(
+            <Drawer qa={qa} className={b()} keepMounted>
+                <div>{PLACEHOLDER_TEXT}</div>
+            </Drawer>,
+        );
+
+        expect(screen.getByTestId(qa)).toHaveClass('g-custom-drawer');
```
amje: There is an assertion `toBeInTheDocument()`.

src/components/Drawer/__stories__/mock.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
+
+export function PlaceholderText() {
+    return (
+        <React.Fragment>
+            Lorem ipsum dolor sit amet consectetur, adipisicing elit. Deserunt iste dolores tenetur
```
amje: There is no need to keep those mocks. We have `faker` library, to generate test content, [including](https://fakerjs.dev/api/lorem.html) lorem sentences

src/components/Drawer/__snapshots__/Drawer.visual.test.tsx-snapshots/Drawer-render-with-hideVeil-prop-dark-chromium-linux.png
amje, line 1: Content is clipped on the screenshot

src/components/Drawer/components/Drawer.tsx
igor0226: If I'm not getting anything wrong, we have the following structure. Therefore, I cannot change FloatingOverlay's opacity in the way it won't affect the DrawerItem's opacity. We want to have two parallel actions to happen on Drawer close:
1) the veil's opacity is changing 1 -> 0
2) the DrawerItem's is sliding to the passed direction (or placement, as you propose). The DrawerItem's opacity is not changing here

```
<FloatingOverlay...>
    <div className={b('veil', {hidden: hideVeil})} />
    <FloatingFocusManager...>
       <DrawerItem
    </FloatingFocusManager>
</FloatingOverlay> 
```

src/components/Drawer/components/Drawer.scss
```diff
@@ -0,0 +1,173 @@
...
+    z-index: #{$drawer-z-index};
+    inset: 0;
+
+    background-color: var(--g-drawer-veil-background-color, var(--g-color-sfx-veil));
+
+    pointer-events: initial;
+
+    &__item {
+        position: absolute;
+        z-index: #{$drawer-z-index + 1};
```
amje: There is no point of trying to override parent's zIndex. They are in different stacking contexts. What's your goal here?

src/components/Drawer/__stories__/mock.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+
+import {faker} from '@faker-js/faker/locale/en';
+
+export function PlaceholderText() {
+    faker.lorem.sentences(2);
```
amje: Something went wrong here

PR #2355, opened 2025-07-29
fix(Button): raised button has incorrect background color apperance

js2me:
## Summary by Sourcery

Bug Fixes:
- Set the raised button’s background color via its CSS variable rather than the direct background property


amje:
@js2me Thanks for the fix!

PR #2358, opened 2025-07-30
feat(PlaceholderContainer): add pin to props of action

Leonid21B:
https://st.yandex-team.ru/STARTREK-57195

## Summary by Sourcery

New Features:
- Add className prop to PlaceholderContainerActionProps to enable custom styling of action buttons


Leonid21B:
> @Leonid21B could you give an example of concrete case? What problem would these changes solve?

For example in our interface we need add styles for button in component
telegram-cloud-photo-size-2-5460671005988091325-y

PR #2362, opened 2025-07-31
fix(Toaster): add vertical gap for wrapped actions

### LINKED ISSUE #2360: No vertical gap between buttons in the toaster.
### Describe the bug

When button labels are long enough to wrap to multiple lines, the vertical spacing between buttons collapses, making them appear visually "stuck" together without proper separation.

Would be great to add consistent vertical spacing between buttons regardless of text length.

### Reproduction

https://stackblitz.com/edit/gravity-ui-uikit-example-b1ne8pzn?file=src%2FApp.tsx,src%2Findex.tsx

### Steps to reproduce

Click `Show Toast` button

Toast will appear.

### Environment

MacOS, Sequoia 15.5.
Chromium

### Validations

ogonkov:
Fixes #2360

### DIFF
--- a/src/components/Toaster/Toast/Toast.scss
+++ b/src/components/Toaster/Toast/Toast.scss
@@ -105,10 +105,9 @@ $block: '.#{variables.$ns}toast';
 
     &__actions {
         margin-block-start: var(--g-spacing-3);
-    }
-
-    &__action {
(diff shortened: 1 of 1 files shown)


korvin89:
> @korvin89 they already exist, was added in #1893

Nope, there is no screenshot for this case

PR #2375, opened 2025-08-07
feat(NumberInput): add decimalScale prop

### LINKED ISSUE #2321: Add propetry to specifying the number of decimal places
### Description

The `NumberInput` component currently lacks built-in support for controlling decimal precision (number of digits after the decimal point). Adding this feature would improve usability for financial, scientific, and measurement-related inputs where consistent decimal formatting is required.

### Suggested solution

Add a new prop (e.g., `decimalScale` or `precision`) to:
- **Restrict input**: Automatically limit decimal places during typing  
  Example: `decimalScale={2}` → Allows `"12.34"` but prevents `"12.345"`  
- **Format values**: Display numbers with consistent decimal places  
  Example: `decimalScale={2}` → Shows `"5"` as `"5.00"`  
- **Round values**: Adjust existing values to match precision  
  Example: Input `"3.14159"` with `decimalScale={2}` → Becomes `"3.14"`

### Alternative

_No response_

### Additional context

_No response_

### Validations

ErmolaevID:
close #2321 

I'd like to highlight several important points regarding the current implementation:
1. When value is a number that has more decimal places than the `decimalScale` allows, rounding to the nearest integer occurs. For example: `<NumberInput ... value={1.28} decimalScale={1} /> => value === 1.3` Is this the expected behavior? 
2. When `allowDecimal = false` and attempting to insert a floating-point value, the current behavior rounds down (floor) rather than to the nearest integer. This behavior was already present in the library. I've added a test case for this scenario with name `"rounds float number down if not allowDecimal prop"`. For example: `<NumberInput value={1.8}/> => value === 1`. Is this the expected behavior, or should it round to the nearest integer in this case?
3. How should we handle situations where `step` contains more decimal places than allowed by `decimalScale`? Currently, I've implemented behavior where `step` gets rounded to match the number of decimal places specified in `decimalScale`. For example: `<NumberInput step={0.08} decimalScale={1} /> => step === 0.1`. Check: `rounds up step value` and `rounds down step value` tests. However, would it be more appropriate to throw an error for an invalid `step` configuration?

### DIFF
--- a/src/components/NumberInput/NumberInput.tsx
+++ b/src/components/NumberInput/NumberInput.tsx
@@ -2,6 +2,8 @@
 
 import * as React from 'react';
 
+import isNumber from 'lodash/isNumber';
+
 import {KeyCode} from '../../constants';
 import {useControlledState, useForkRef} from '../../hooks';
 import {useFormResetHandler} from '../../hooks/private';
@@ -18,6 +20,7 @@ import {
     getInternalState,
     getParsedValue,
     getPossibleNumberSubstring,
+    truncateExtraDecimalNumbers,
     updateCursorPosition,
 } from './utils';
 
@@ -69,6 +72,8 @@ export interface NumberInputProps
      * @default false
      */
     allowDecimal?: boolean;
+    /** Maximum number of digits allowed after the decimal point  */
+    decimalScale?: number;
     /** The control's value */
     value?: number | null;
     /** The control's default value. Use when the component is not controlled */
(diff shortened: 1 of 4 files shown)


amje (changes requested):
Waiting for review

src/components/NumberInput/NumberInput.tsx
```diff
@@ -69,6 +72,8 @@ export interface NumberInputProps
      * @default false
      */
     allowDecimal?: boolean;
+    /** Maximum number of digits allowed after the decimal point  */
+    decimalScale?: number;
```
amje, line 76: Let's call the prop `decimalDigits` to be more accurate

amje, line 76: Let's reflect this prop in the docs (READMEs)

amje (changes requested):

PR #2380, opened 2025-08-13
feat: use @hello-pangea/dnd

Raubzeug:
`react-beautiful-dnd` is no longer supported and achieved.

We declare support for React19, but `react-beautiful-dnd` doesn't support it and crashes. So I suggest to replace it with `@hello-pangea/dnd`. It's a fork from `react-beautiful-dnd`, which is developing. In particular, the bug with StrictMode is solved, and React19 is supported. Api is the same as `react-beautiful-dnd`.


amje:
@sourcery-ai review

korvin89 (approved):

PR #2384, opened 2025-08-14
fix(DefinitionList): should not overflow parent container

### LINKED ISSUE #2381: Copy button in DefinitionList is causing overflow issues
### Describe the bug

When rendered, copy button pushes parent's width beyond its content

### Reproduction

https://stackblitz.com/edit/gravity-ui-uikit-example-a2epqv7b?file=src%2FApp.tsx

### Steps to reproduce

1. Set viewport so content starts wrapping
2. Page has scroll, but is expected not to have

### Environment

**OS**: MacOS Sequoia 15.5
**Browser**: Edge 138.0

### Validations

Raubzeug:
closes #2381


src/components/DefinitionList/DefinitionList.scss
```diff
@@ -69,9 +73,9 @@ $block: '.#{variables.$ns}definition-list';
         position: relative;
         display: inline-flex;
         align-items: center;
-        padding-inline-end: var(--g-spacing-7);
+        padding-inline-end: var(--_--copy-button);
 
-        margin-inline-end: calc(-1 * var(--g-spacing-7));
+        margin-inline-end: calc(-1 * var(--_--copy-button));
```
amje: I guess these two lines not needed anymore if we adjust positioning of the button

PR #2385, opened 2025-08-14
feat(TextInput): ability to pass popover props for error message

js2me:


korvin89:
@js2me  Hi! We have added contribution guide, you can check it [here](https://github.com/gravity-ui/uikit/tree/main/contribute#how-to-contribute).

Could you provide reproduction of this bug?

js2me:
@korvin89 Hello, this is not bug, this is enhancement of the TextInput component

korvin89:
> @korvin89 Hello, this is not bug, this is enhancement of the TextInput component

I mean this one https://t.me/gravity_ui_chat/6836

Here is a deal: we don't want to give such api for users, but we want to make your case work. We need to understand - is it a bug in component, or bug in your environment

PR #2386, opened 2025-08-14
feat: add ColorPicker component

Estasie:


package.json
```diff
@@ -131,8 +131,11 @@
     "@gravity-ui/i18n": "^1.8.0",
     "@gravity-ui/icons": "^2.13.0",
     "@tanstack/react-virtual": "^3.13.9",
+    "@uiw/react-color": "^2.7.3",
     "blueimp-md5": "^2.19.0",
+    "i": "^0.3.7",
     "lodash": "^4.17.21",
+    "npm": "^11.5.2",
```
amje: Are these deps added by mistake?

src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,198 @@
...
+            />
+
+            <Popup
+                open={open}
+                className={b('popup')}
+                placement={'bottom-end'}
+                anchorElement={anchor}
+                onOpenChange={handleOpenChange}
+            >
+                <Flex direction={'column'} gap={2}>
```
amje: Flex component should not be used inside the other components, it's for user-only component as it follows spacing variables which can be overwriten and we want our components to be fixed spacing

src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,198 @@
...
+                hsva={hsva}
+                onClick={() => setState((prev) => ({...prev, open: true}))}
+                onColorChange={updateHsva}
+                ref={setAnchor}
+            />
+
+            <Popup
+                open={open}
+                className={b('popup')}
+                placement={'bottom-end'}
```
amje: Only `bottom-end` value makes component look shifted when it rendered on the left side of the screen. Let's use `'bottom-start' | 'bottom-end'` so the component could flip from side to side.

src/components/ColorPicker/components/ColorDisplay/ColorDisplay.tsx
```diff
@@ -0,0 +1,68 @@
...
+                onColorChange?.(updatedHsva);
+            } catch {
+                setInputValue(getTextValueByMode(hsva, Modes.Hex, false));
+            }
+        }, [inputValue, hsva.a, onColorChange, hsva]);
+
+        return (
+            <Card view={'outlined'} className={b('picker-wrapper')} ref={ref}>
+                <Flex alignItems={'center'} gap={2}>
+                    <Button size={'s'} className={b('underlay')} onClick={onClick}>
```
amje: Button has complex styles and instead of overriding them it's easier to use native button, reset and apply only needed styles. It will be much cleaner

package.json
```diff
@@ -131,6 +131,7 @@
     "@gravity-ui/i18n": "^1.8.0",
     "@gravity-ui/icons": "^2.13.0",
     "@tanstack/react-virtual": "^3.13.9",
+    "@uiw/react-color": "^2.7.3",
```
amje: Keyboard support has been added in 2.8.0. Let's update

src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+import {ColorDisplay, ColorPointer, HexInput, RgbInputs} from './components';
+import {DEFAULT_COLOR, b} from './constants';
+import {Modes} from './types';
+import {useColorPickerState} from './useColorPickerState';
+import {convertSelectedModeColorToHsva, getTextValueByMode} from './utils';
+
+export interface ColorPickerProps {
+    withAlpha?: boolean;
+    mode?: Modes;
```
amje: `mode` seems not very usefull if it's only affect inputs in popup. I assume we can remove it without affecting ux/dx

src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,177 @@
...
+    defaultColor?: HsvaColor;
+    onChange?: (color: HsvaColor) => void;
+    onClose?: () => void;
```
amje: The component should have similar API to other inputs. Let's change API to the following:
```ts
interface ColorPickerProps {
    // Same size as in the `TextInput`
    size?: 's' | 'm' | 'l' | 'xl';
    // if controlled value
    value?: string;
    // if uncontrolled value
    defaultValue?: string;
    // onChange naming is reserved for native event handlers
    onUpdate?: (value: string) => void;
    // if controlled open state
    open?: boolean;
    // if uncontrolled open state
    defaultOpen?: boolean;
    onOpenChange?: (open: boolean) => void;
    // it should be false by default (common case)
    withAlpha?: boolean;
}
```

For controlled/uncontrolled state there is a hook `useContolledState` (see other components as an example)

src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,177 @@
...
+import {DEFAULT_COLOR, b} from './constants';
+import {Modes} from './types';
+import {useColorPickerState} from './useColorPickerState';
+import {convertSelectedModeColorToHsva, getTextValueByMode} from './utils';
+
+export interface ColorPickerProps {
+    withAlpha?: boolean;
+    mode?: Modes;
+    defaultColor?: HsvaColor;
+    onChange?: (color: HsvaColor) => void;
```
amje: What about accepting more common format as value, like simple HEX?

Estasie: I changed it to the hex, incoming as string and resolving it only for external usage,  locally I use hsva for better precision

src/components/ColorPicker/ColorPicker.tsx
Estasie: Added more props, also added 'onlyPicker' prop for button only render without value

src/components/ColorPicker/ColorPicker.tsx
Estasie: Removed it

package.json
amje: Keyboard navigation is not working, but controls became focusable. Maybe we need to add some handlers?

src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,226 @@
...
+     */
+    onOpenChange?: (open: boolean) => void;
+    /*
+     *Enables alpha channel support for HEXA/RGBA mode and transparency slider.
+     */
+    withAlpha?: boolean;
+    /*
+     * Render only picker button without value
+     */
+    onlyPicker?: boolean;
```
amje: Let's call it `compact: boolean`. We have similar prop in `Pagination` component

src/components/lab/ColorPicker/components/RgbInputs/RgbInputs.tsx
```diff
@@ -0,0 +1,56 @@
...
+type ChannelConfig = {
+    label: string;
+    pin: InputControlPin;
+};
+
+const createChannelInput = ({label, pin}: ChannelConfig) => ({
+    renderInput: (props: any) => (
+        <TextInput
+            value={String(props.value)}
+            onChange={props.onChange}
```
amje, line 38: I think we should pass rest props to the `controlProps` as component expect them to be on the input element. Same true for `HexInput` component

package.json
Estasie: Now keyboard navigation should work fine

src/components/ColorPicker/constants.ts
```diff
@@ -0,0 +1,6 @@
+import {block} from '../utils/cn';
+
+import './ColorPicker.scss';
+
+export const b = block('color-picker');
+export const DEFAULT_COLOR = '#aeaeae';
```
amje: Let's use `#000000` like native color picker does

src/components/ColorPicker/index.ts
```diff
@@ -0,0 +1,2 @@
+export {ColorPicker, type ColorPickerProps} from './ColorPicker';
+export type {Modes as ColorPickerModes} from './types';
```
amje: Isn't it a private API now?

src/components/ColorPicker/ColorPicker.scss
```diff
@@ -0,0 +1,207 @@
...
+                --_--base-picker-wrapper-width: 130px;
+            }
+
+            &_xl {
+                --__--picker-wrapper-padding: 0px var(--g-spacing-1);
+                --_--base-picker-wrapper-width: 140px;
+            }
+        }
+
+        &_onlyPicker {
```
amje: Let's keep `kebab-case` through SCSS-files

src/components/ColorPicker/ColorPicker.scss
```diff
@@ -0,0 +1,207 @@
...
+        display: inline-block;
+        position: relative;
+    }
+
+    &__color-swatch {
+        box-sizing: border-box;
+
+        /* Color swatch specific styles */
+        min-width: var(--g-button-height, var(--_--height));
+        height: var(--g-button-height, var(--_--height));
```
amje: Not sure if we want Button's CSS API to work in this component

Estasie: removed button styles and replaced it with local ones

PR #2387, opened 2025-08-18
feat: add useLang hook, alllow pass lang via ThemeProvider

ValeraS:


src/components/theme/ThemeProvider.tsx
```diff
@@ -30,6 +32,7 @@ export interface ThemeProviderProps extends React.PropsWithChildren<{}> {
     scoped?: boolean;
     rootClassName?: string;
     layout?: Omit<PrivateLayoutProviderProps, 'children'>;
+    langOptions?: Partial<LangOptions>;
```
amje: I would keep `lang` and `fallbackLang` on the top level to not be repetitive (`langOptions.lang*`)

src/components/theme/__stories__/Theme.stories.tsx
```diff
@@ -25,25 +26,48 @@ export default meta;
...
+            'dialog.body': 'Dialog Body',
+            'current direction': 'current direction',
+        },
+        ru: {
+            one: 'Один',
+            two: 'Два',
+            three: 'Три',
+            'dialog.header': 'Заголовок диалога',
+            'dialog.body': 'Тело диалога',
+            'current direction': 'текущее направление',
```
amje, line 46: Let's also add translations for the tooltip

amje (approved):

src/components/theme/ThemeProvider.tsx
```diff
@@ -18,11 +18,13 @@ import {
...
 import {updateBodyClassName, updateBodyDirection} from './dom-helpers';
 import type {Direction, RealTheme, Theme, ThemeContextProps} from './types';
+import type {LangOptions} from './useLang';
+import {LangContext, defaultLangOptions} from './useLang';
 import {useSystemTheme} from './useSystemTheme';
 
 const b = block(ROOT_CLASSNAME);
 
-export interface ThemeProviderProps extends React.PropsWithChildren<{}> {
+export interface ThemeProviderProps extends React.PropsWithChildren<{}>, Partial<LangOptions> {
```
ogonkov, line 27: I believe ThemeProvider is no longer accurate

src/components/theme/useLang.tsx
```diff
@@ -0,0 +1,27 @@
...
+    fallbackLang: StringWithSuggest<Lang>;
+}
+
+export const defaultLangOptions: LangOptions = {lang: 'en', fallbackLang: 'en'};
+
+export const LangContext = React.createContext<LangOptions | undefined>(undefined);
+
+export function useLang(): LangOptions {
+    const config = useSyncExternalStore(
+        (onChange) => subscribeConfigure(onChange),
```
ogonkov: React docs recommend to memoize callbacks for this hook, i'm not sure if same applies to this package

src/components/theme/ThemeProvider.tsx
ValeraS: Yes, we have decided to rename it in the next major release.

src/components/theme/useLang.tsx
ValeraS: I don't see such recommendations in the React docs, but yes, wrappers are not needed here.

ogonkov: I was reffering to this

> If a different subscribe function is passed during a re-render, React will re-subscribe to the store using the newly passed subscribe function. You can prevent this by declaring subscribe outside the component.

ogonkov (approved):

PR #2391, opened 2025-09-02
refactor(Select): optimize selected options handling and improve performance with memoization

Antamansid:
What is the problem?

We have a large log panel on our project, where using the Select component it is possible to select a log file.

Unfortunately, it turned out that there can be an extremely large number of these logs (there were 251_756 files at the booth where I noticed the problem - and this, I suppose, is not the limit). With so many items, the page just froze. 

It seems like the changes are not critical and should not significantly affect the logic of the component as a whole.
I tested them on the project - yes, everything stopped lagging.
I have prepared a stand for reproduction: [click](https://stackblitz.com/edit/gravity-ui-uikit-example-mjamgcvo?file=src%2FApp.tsx)

I didn't specify 250_000 points there, with the current number, my browser was already noticeably lagging.


src/components/Select/components/SelectList/SelectList.tsx
```diff
@@ -5,8 +5,8 @@ import * as React from 'react';
 import {List} from '../../../List';
 import {SelectQa, selectListBlock} from '../../constants';
 import type {SelectOption, SelectProps} from '../../types';
-import {getOptionsHeight, getPopupItemHeight, scrollToItem} from '../../utils';
```
korvin89: Looks like unnecessary changing

src/components/Select/utils.tsx
```diff
@@ -120,12 +120,12 @@ export const getSelectedOptionsContent = (
         (opt) => !isSelectGroupTitle(opt),
     ) as SelectOption[];
 
-    const selectedOptions = value.reduce((acc, val) => {
-        const selectedOption = flattenSimpleOptions.find((opt) => opt.value === val);
+    const optionsMap = new Map<string, SelectOption>();
```
korvin89: Lets do it in one instruction, kinda
```ts
const optionsMap = new Map<string, SelectOption>(flattenSimpleOptions.map((opt) => [opt.value, opt]);
```

PR #2432, opened 2025-09-12
fix(CopyToClipboard): correctly use fallback

### LINKED ISSUE #2421: Fix copy to clipboard functionality on Colors and Typography demos
### Describe the bug

Nothing is copied to the clipboard on clicking:

Image

### Reproduction

https://preview.gravity-ui.com/uikit/?path=/story/colors--texts

### Steps to reproduce

_No response_

### Environment

MacOS 15.5
Microsoft Edge 138

### Validations

Yanseses:
https://github.com/gravity-ui/uikit/issues/2421


dgaponov (approved):

src/components/CopyToClipboard/copyText.ts
```diff
@@ -1,7 +1,7 @@
 export function copyText(text: string) {
     if (typeof navigator !== 'undefined' && navigator.clipboard?.writeText) {
         try {
-            return navigator.clipboard.writeText(text);
+            navigator.clipboard.writeText(text);
```
ogonkov: How this solves the problem?

amje: `writeText` returns a Promise, so if it fails normally (security issue) we'll never get into the catch block and run fallback

amje: I guess this should we async/await function to work correctly indeed

PR #2434, opened 2025-09-16
docs(Loader/Spin): update stories layout

### LINKED ISSUE #2423: Update story layout for Loader Size story and Spin Size story. Use ClipboardButton Size story as a reference
### Description

Use `Showcase*` components for layout

### Additional context

_No response_

### Validations

- [x] Check that there isn't [already an issue](https://github.com/gravity-ui/uikit/issues) that pursue the same goal to avoid creating a duplicate.

baevm:
Updated stories for Loader and Spin components.
closes #2423


src/components/Spin/__stories__/Spin.stories.tsx
```diff
@@ -1,27 +1,36 @@
...
-        <span style={{margin: '16px'}} />
-        xl: <Spin {...args} size="xl" />
-    </div>
-);
-export const Size = SizeTemplate.bind({});
+export const Default: Story = {};
+
+export const Size: Story = {
+    render: (args) => (
+        <Showcase>
```
ogonkov: Move Showcase to `decorators` please

baevm: If i got it right, like this? moved in commit 71e8e5a64ab2498f5435568f274b6195caa2b0b7

ogonkov: Yes, but move `decorators` to story obj please, in case we would want to add some story, that wont need showcase wrapper

baevm: fixed in fb82fa9a6e5573ae64b27dd99a22a1b8daaca264

ogonkov (approved):

Issue #2437, opened 2025-09-18
Add new size for Switch and loading state for Switch


lentsd:
### Description

New S size and Loading state

### Suggested solution

Add new size and loading prop

### Validations

- [x] Check that there isn't [already an issue](https://github.com/gravity-ui/uikit/issues) that request the same feature to avoid creating a duplicate.

PR #2438, opened 2025-09-18
feat(Switch): add 's' size option and loading state support

lentsd:
Link to issue https://github.com/gravity-ui/uikit/issues/2437

### DIFF
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -4,7 +4,6 @@ import * as React from 'react';
 
 import {useCheckbox} from '../../hooks/private';
 import {ControlLabel} from '../ControlLabel';
-import type {ControlLabelSize} from '../ControlLabel';
 import type {ControlProps, DOMProps, QAProps} from '../types';
 import {block} from '../utils/cn';
 
@@ -13,7 +12,7 @@ import {CheckboxTickIcon} from './CheckboxTickIcon';
 
 import './Checkbox.scss';
 
-export type CheckboxSize = ControlLabelSize;
+export type CheckboxSize = 'm' | 'l' | 'xl';
 
 export interface CheckboxProps extends ControlProps, DOMProps, QAProps {
     size?: CheckboxSize;
--- a/src/components/ControlLabel/ControlLabel.scss
+++ b/src/components/ControlLabel/ControlLabel.scss
@@ -17,6 +17,11 @@ $block: '.#{variables.$ns}control-label';
         pointer-events: none;
     }
 
+    &_size_s {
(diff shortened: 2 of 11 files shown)


src/components/Radio/Radio.tsx
```diff
@@ -4,15 +4,14 @@ import * as React from 'react';
...
-import type {ControlLabelSize} from '../ControlLabel';
 import type {ControlProps, DOMProps, QAProps} from '../types';
 import {block} from '../utils/cn';
 
 import './Radio.scss';
 
 const b = block('radio');
 
-export type RadioSize = ControlLabelSize;
+export type RadioSize = 'm' | 'l' | 'xl';
```
korvin89, line 14: Why break the current type relationship?

src/components/Checkbox/Checkbox.tsx
```diff
@@ -13,7 +12,7 @@ import {CheckboxTickIcon} from './CheckboxTickIcon';
 
 import './Checkbox.scss';
 
-export type CheckboxSize = ControlLabelSize;
+export type CheckboxSize = 'm' | 'l' | 'xl';
```
korvin89, line 15: Why break the current type relationship?

lentsd: Because Radio, Checkbox, and Switch don't have all the size options that ControlLabel has. In this case, we see incorrect behavior in the storybook. For example, the current broken storybook for Switch XL size: https://preview.gravity-ui.com/uikit/?path=/story/components-inputs-switch--default&args=size:xl

korvin89: Checkbox and Radio have the same set of sizes

lentsd: I add S size for control label in this pull request, so now they don't have same set of sizes

src/components/ControlLabel/ControlLabel.scss
```diff
@@ -17,6 +17,11 @@ $block: '.#{variables.$ns}control-label';
         pointer-events: none;
     }
 
+    &_size_s {
+        font-size: var(--g-text-body-1-font-size);
+        line-height: 16px;
```
korvin89: Let’s use the same line height here as in the M size, otherwise it looks inconsistent.

src/components/Checkbox/Checkbox.tsx
korvin89: Oh, yeeep, my bad

src/components/ControlLabel/ControlLabel.scss
lentsd: done 👌🏼

PR #2443, opened 2025-09-26
feat: add `useCollapseChildren` hook

amje:


src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -47,103 +49,33 @@ export const Breadcrumbs = React.forwardRef(function Breadcrumbs(
...
-        setCalculated(false);
-    }, [items.length]);
-    useResizeObserver({
-        ref: listRef,
-        onResize: handleResize,
+    const {
+        calculated,
+        recalculate,
+        visibleCount: visibleItemsCount,
+    } = useCollapseChildren({
```
ValeraS, line 56: We need to limit the width of the current element. Let's set the `current` data attribute for the current element or get ref to the current element  and use the getChildWidth function to limit the maximum width.

```suggestion
    } = useCollapseChildren({
        getChildWidth(child) {
            const width = child.getBoundingClientRect().width;
            const maxWidth = child.dataset['current'] ? 200 : Infinity;
            return Math.min(maxWidth, width);
        },

```

amje: fixed

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -2,7 +2,8 @@
 
 import * as React from 'react';
 
-import {useForkRef, useLayoutEffect, useResizeObserver} from '../../hooks';
+import {useForkRef, useResizeObserver} from '../../hooks';
+import {useCollapseChildren} from '../../hooks/useCollapseChildren';
```
ValeraS:
```suggestion
import {useCollapseChildren, useForkRef, useResizeObserver} from '../../hooks';
```

src/hooks/useCollapseChildren/index.ts
```diff
@@ -0,0 +1 @@
+export * from './useCollapseChildren';
```
ValeraS, line 1: need to add re-export in `src/hooks/index.ts`

ValeraS (approved):

src/hooks/useCollapseChildren/index.ts
amje: done

src/components/Breadcrumbs/Breadcrumbs.tsx
amje: done

PR #2447, opened 2025-10-05
refactor: migration to lodash-es from lodash (ES)

js2me:


korvin89:
@js2me could fix unit tests?

korvin89:
@js2me We discussed the current pull request with the team and concluded that it introduces breaking changes to the library, so we’ll postpone these changes until the next major release. I’ll add an appropriate tag to this pull request so that we don’t lose track of it in the future.

amje:
We decided to migrate to the es-toolkit instead #2707

PR #2448, opened 2025-10-06
chore(HelpMark): add controls layout showcase story

amje:


Raubzeug (changes requested):
A bit not aligned when standalone. What do you think, maybe it's expected behaviour?

Screenshot 2025-10-06 at 17 46 02

amje:
> A bit not aligned when standalone. What do you think, maybe it's expected behaviour?
> 
> Screenshot 2025-10-06 at 17 46 02

It is what it is now. Hoping we'll fix it in #2342

PR #2449, opened 2025-10-08
feat(Hotkey): remove mac esc icon

MefistAldemisov:
Changed mac `⎋` hotkey character to `esc` key name, which seems to be more convenient


d3m1d0v:
I suggest changing type to "feat" and adding a scope – `Hotkey`. For example:

"feat(Hotkey): update mac esc display name"

PR #2456, opened 2025-10-16
chore: update deps

teleginzhenya:


src/components/controls/TextArea/__snapshots__/TextArea.visual.test.tsx-snapshots/TextArea-smoke-empty-light-chromium-linux.png
teleginzhenya, line 1: Something happened in Chromium after update to `mcr.microsoft.com/playwright:v1.56.0-jammy`
Now Chromium adds weird padding when textarea has disabled attribute


Screenshot 2025-10-16 at 10 31 55
Screenshot 2025-10-16 at 10 31 47

PR #2472, opened 2025-10-29
fix: replace usages of `React.useId` with `useUniqId`

persovt2:
…and PinInput stories


src/components/Accordion/AccordionSummary/AccordionSummary.tsx
```diff
@@ -2,6 +2,8 @@
 
 import * as React from 'react';
 
+import {useUniqId} from 'src/hooks';
```
amje: You should specify relative path. We don't have aliases set up

amje (approved):

PR #2487, opened 2025-11-06
feat: add FileDropZone component

arsen-afaunov:


src/hooks/lab/useDropZone/useDropZone.ts
```diff
@@ -0,0 +1,181 @@
...
+
+        return () => {
+            dropZoneElement?.removeEventListener('dragenter', handleDragEnterNative);
+            dropZoneElement?.removeEventListener('dragover', handleDragOverNative);
+            dropZoneElement?.removeEventListener('dragleave', handleDragLeaveNative);
+            dropZoneElement?.removeEventListener('drop', handleDropNative);
+        };
+    }, [handleDragEnterNative, handleDragLeaveNative, handleDragOverNative, handleDropNative, ref]);
+
+    return {
```
arsen-afaunov, line 215: Add keyboard support.

src/hooks/lab/useDropZone/useDropZone.ts
```diff
@@ -0,0 +1,181 @@
...
+    onDrop: (items: DataTransferItemList) => void;
+}
+
+const DROP_ZONE_BASE_ATTRIBUTES = {
+    'aria-dropeffect': 'copy' as DataTransfer['dropEffect'],
+    tabIndex: 0,
+};
+
+export interface DroppableProps extends Required<typeof DROP_ZONE_BASE_ATTRIBUTES> {
+    onDragEnter: (e: React.DragEvent) => void;
```
arsen-afaunov, line 28: Should use generics? Why?

src/hooks/useDropZone/useDropZone.ts
```diff
@@ -0,0 +1,181 @@
...
+        [handleDragOverNative],
+    );
+
+    const handleDragLeaveNative = React.useCallback(
+        (event: DragEvent) => {
+            if (disabled || !eventItemTypesAcceptable(accept, event)) {
+                return;
+            }
+
+            setIsDraggingOver(false);
```
arsen-afaunov: Fix nested events behavior.

src/hooks/lab/useDropZone/useDropZone.ts
arsen-afaunov: Done.

src/hooks/lab/useDropZone/useDropZone.ts
arsen-afaunov: No gain.

src/hooks/useDropZone/useDropZone.ts
arsen-afaunov: Fixed.

src/components/lab/FileDropZone/README.md
```diff
@@ -0,0 +1,106 @@
+<!--GITHUB_BLOCK-->
```
arsen-afaunov, line 1: Add StoryBook docs.

korvin89:
The state of the component is not reset with such a case:

https://github.com/user-attachments/assets/b7320bb0-f837-49b0-aee1-6e4d4a05237d

src/hooks/useFileInput/useFileInput.ts
```diff
@@ -3,6 +3,7 @@ import * as React from 'react';
 export type UseFileInputProps = {
     onUpdate?: (files: File[]) => void;
     onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
+    multiple?: boolean;
```
korvin89: The `multiple` prop was not initially added, as the hook gives the basic configuration for the native input. That is, if the consumer needs `multiple`, then you need to add it yourself, otherwise it turns out to be a strange proxy that passes the prop through and returns it in the same form without using it in any way in calculations.

src/hooks/useDropZone/useDropZone.ts
```diff
@@ -0,0 +1,191 @@
...
+    onDrop: (items: DataTransferItemList) => void;
+}
+
+const DROP_ZONE_BASE_ATTRIBUTES = {
+    'aria-dropeffect': 'copy' as DataTransfer['dropEffect'],
+    tabIndex: 0,
+    role: 'button',
+};
+
+export interface DroppableProps extends Required<typeof DROP_ZONE_BASE_ATTRIBUTES> {
```
korvin89: We usually name props according to this logic `<ComponentOrHookName> + details`. This is necessary so that all types have their own name space + it improves DX when working with lib types

src/hooks/useDropZone/README.md
```diff
@@ -0,0 +1,26 @@
...
+# useDropZone
+
+<!--/GITHUB_BLOCK-->
+
+```tsx
+import {useDropZone} from '@gravity-ui/uikit';
+```
+
+The `useDropZone` hook provides props for an element to act as a drop zone and also gives access to the dragging-over state.
+Additionally, the hook supports a more imperative approach: it can accept a `ref` for cases where you don't have direct access to the HTML element you want to make a drop zone.
```
korvin89: Can you describe the minimum usage examples?

src/hooks/useDropZone/useDropZone.ts
```diff
@@ -0,0 +1,191 @@
...
+    onDragEnter: (e: React.DragEvent) => void;
+    onDragOver: (e: React.DragEvent) => void;
+    onDragLeave: (e: React.DragEvent) => void;
+    onDrop: (e: React.DragEvent) => void;
+}
+
+export interface UseDropZoneState {
+    isDraggingOver: boolean;
+    getDroppableProps: () => DroppableProps;
```
korvin89: There seems to be no reason to make this as function

src/hooks/lab/useDropZone/useDropZone.ts
```diff
@@ -0,0 +1,191 @@
...
+    onDrop: (e: React.DragEvent) => void;
+}
+
+export interface UseDropZoneState {
+    isDraggingOver: boolean;
+    getDroppableProps: () => DroppableProps;
+}
+
+function typeMatchesPattern(actualMimeType: string, expectedMimeTypePattern: string): boolean {
+    const actualMimeTypeParts = actualMimeType.split('/');
```
korvin89, line 46: Are we assuming that only files will be used?

src/hooks/useDropZone/useDropZone.ts
```diff
@@ -0,0 +1,191 @@
...
+                return;
+            }
+
+            nestingCounterRef.current--;
+
+            if (nestingCounterRef.current !== 0) {
+                return;
+            }
+
+            setIsDraggingOver(false);
```
korvin89: `isDraggingOver` sets to true without a condition, but sets to false by condition, can you describe why how logic works?

src/components/lab/FileDropZone/FileDropZone.Provider.tsx
```diff
@@ -0,0 +1,140 @@
...
+            | 'errorIcon'
+            | 'errorMessage'
+            | 'validationState'
+        >,
+        UseFileInputResult,
+        UseDropZoneState {}
+
+const FileDropZoneContext = React.createContext<FileDropZoneContextValue | null>(null);
+
+export const FileDropZoneProvider = ({
```
korvin89, line 47: We usually give `Provider` postfix to entities that don't have a layout. For example: [TabProvider](https://github.com/gravity-ui/uikit/blob/main/src/components/tabs/TabProvider.tsx#L10), [ToasterProvider](https://github.com/gravity-ui/uikit/blob/d716102ed6138ea1665739fa973234820067d517/src/components/Toaster/Provider/ToasterProvider.tsx#L15), [AccordionProvider](https://github.com/gravity-ui/uikit/blob/d716102ed6138ea1665739fa973234820067d517/src/components/Accordion/AccordionContext.tsx#L53C17-L53C34), [DisclosureProvider](https://github.com/gravity-ui/uikit/blob/d716102ed6138ea1665739fa973234820067d517/src/components/Disclosure/DisclosureContext.tsx#L33C17-L33C35) etc

src/components/lab/FileDropZone/FileDropZone.tsx
```diff
@@ -0,0 +1,68 @@
...
+import './FileDropZone.scss';
+
+export interface FileDropZoneProps extends Pick<BaseInputControlProps, 'validationState'> {
+    accept: UseDropZoneAccept;
+    onAdd: (files: File[]) => void;
+    title?: string;
+    description?: string;
+    buttonText?: string;
+    icon?: IconData | null;
+    errorIcon?: IconData | null;
```
korvin89, line 24: Do we really need more than one icon property?

src/components/lab/FileDropZone/parts/FileDropZone.Button.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+
+import {Button} from '../..';
+import {useFileZoneContext} from '../FileDropZone.Provider';
+import {cnFileDropZone} from '../FileDropZone.classname';
+import i18n from '../i18n';
+
+type FileDropZoneButtonProps = {
+    className?: string;
```
korvin89, line 9: It seems to me that it would be convenient to have the `children` property so that explicitly using the button in the component layout, you can explicitly define the text in the button itself and not in its parent

`buttonText` is convenient to use as a parent зкщзукен if you don't explicitly use `FileDropZoneButton`

src/components/lab/FileDropZone/parts/FileDropZone.Description.tsx
```diff
@@ -0,0 +1,20 @@
+import {Text} from '../..';
+import {useFileZoneContext} from '../FileDropZone.Provider';
+
+type FileDropZoneDescriptionProps = {
+    className?: string;
```
korvin89, line 5: The same with the `children` property as in `FileDropZoneButton`

src/components/lab/FileDropZone/parts/FileDropZone.Title.tsx
```diff
@@ -0,0 +1,24 @@
+import {Text} from '../..';
+import {useFileZoneContext} from '../FileDropZone.Provider';
+import i18n from '../i18n';
+
+type FileDropZoneTitleProps = {
+    className?: string;
```
korvin89, line 6: The same with the `children` property as in `FileDropZoneButton`

src/components/FileDropZone/FileDropZone.tsx
```diff
@@ -0,0 +1,68 @@
...
+import {FileDropZoneButton} from './parts/FileDropZone.Button';
+import {FileDropZoneDescription} from './parts/FileDropZone.Description';
+import {FileDropZoneIcon} from './parts/FileDropZone.Icon';
+import {FileDropZoneTitle} from './parts/FileDropZone.Title';
+
+import './FileDropZone.scss';
+
+export interface FileDropZoneProps extends Pick<BaseInputControlProps, 'validationState'> {
+    accept: UseDropZoneAccept;
+    onAdd: (files: File[]) => void;
```
korvin89: We usually use the name `onUpdate` for such callbacks, this is a familiar pattern for users of our library

src/components/lab/FileDropZone/README.md
arsen-afaunov: Done.

src/hooks/useDropZone/useDropZone.ts
arsen-afaunov: I think the function will scale better in case parameterization is needed: you can just add optional arguments instead of overloading the type and turning it into a union. In my view, the function is a better fit here.

src/hooks/lab/useDropZone/useDropZone.ts
arsen-afaunov: I didn’t quite understand the question. MIME types allow specifying arbitrary data. For instance, in the Storybook example, `text/plain` is used, which enables you to drag and drop HTML elements. Moreover, this way you can add custom MIME types like `application/x-my-custom-type` to make sure that only permitted elements are drag-n-dropped —this can be useful, for example, for a Kanban board app.

src/components/lab/FileDropZone/parts/FileDropZone.Button.tsx
arsen-afaunov: The motivation was as follows. First, I wanted to eliminate redundancy in the component's interface and make the parent's props the single source of truth in this regard. This way, the cognitive load on the component user is reduced, as they no longer need to keep two sets of props with the same meaning in mind. Second, this approach allows for the encapsulation of the logic related to text handling, ensuring consistency between the displayed texts and the component's internal states—for example, the user of the component won't need to handle error messages or grammatical pluralization themselves if the `multiple` prop is provided.

More generally, the compound component pattern here was intended solely to allow arbitrary layouts of elements, while fully hiding their internal implementation and their relationships.

src/components/lab/FileDropZone/FileDropZone.tsx
arsen-afaunov: It seemed to me that since the base icon and the “error” icon are semantically different, it makes sense to allow them to be customized separately. If we don't want to give the component user this flexibility in order to maintain visual consistency, that's not a problem—we can just remove this prop.

src/hooks/useDropZone/useDropZone.ts
arsen-afaunov: I didn’t quite understand this either. Both states are set conditionally...

src/components/lab/FileDropZone/FileDropZone.Provider.tsx
arsen-afaunov: Okay. I’ll see what can be done.

src/hooks/useFileInput/useFileInput.ts
arsen-afaunov: Done.

src/hooks/useDropZone/useDropZone.ts
arsen-afaunov: Done.

src/hooks/useDropZone/README.md
arsen-afaunov: Done.

src/components/FileDropZone/FileDropZone.tsx
arsen-afaunov: Done.

src/components/lab/FileDropZone/FileDropZone.Provider.tsx
arsen-afaunov: Moved all markup-related code outside the component.

Issue #2422, opened 2025-09-12
Fix initial render of AvatarStack stories in StrictMode


amje:
### Describe the bug

There is an error on initial render of those stories in dev mode.

### Reproduction

https://preview.gravity-ui.com/uikit/?path=/story/components-data-display-avatarstack--default

### Environment

MacOS 15.5
Microsoft Edge 138

### Validations

- [x] Check that there isn't [already an issue](https://github.com/gravity-ui/uikit/issues) that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #2497, opened 2025-11-15
fix: fix initial render of AvatarStack stories in StrictMode

astandrik:
Giving keys to avatars array items seems to fix the issue

Closes https://github.com/gravity-ui/uikit/issues/2422

### DIFF
--- a/src/components/AvatarStack/__stories__/AvatarStack.stories.tsx
+++ b/src/components/AvatarStack/__stories__/AvatarStack.stories.tsx
@@ -26,16 +26,19 @@ function getChildren({
     avatarSize: AvatarSize;
     randomAvatar: boolean;
 }) {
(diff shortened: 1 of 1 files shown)


src/components/AvatarStack/__stories__/AvatarStack.stories.tsx
```diff
@@ -26,18 +26,16 @@ function getChildren({
     avatarSize: AvatarSize;
     randomAvatar: boolean;
 }) {
-    return faker.helpers.uniqueArray(
```
ogonkov, line 29: Can we keep uniqueArray method?

astandrik: What for? (yes we can)

Issue #2335, opened 2025-07-08
Add color prop to Icon component


olegKusov:
### Description

Main idea. we want to simplify setting color of Icon. Currently we need to use inline style or className.

### Suggested solution

color prop to Icon component.

### Validations

- [x] Check that there isn't [already an issue](https://github.com/gravity-ui/uikit/issues) that request the same feature to avoid creating a duplicate.

PR #2499, opened 2025-11-15
feat(Icon): add `color` prop

astandrik:
Closes https://github.com/gravity-ui/uikit/issues/2335


src/components/Icon/Icon.tsx
```diff
@@ -29,14 +29,26 @@ export interface IconProps extends QAProps, DOMProps {
     size?: number | string;
     fill?: string;
     stroke?: string;
+    color?: CSSProperties['color'] | string;
```
amje: What's the point of duplication style functionality? Let's instead add feature from the `Text` component, which have `color` property accepting text color name (`typeof TEXT_COLORS[number]`).

astandrik: applied your proposal

src/components/Icon/Icon.tsx
```diff
@@ -22,7 +24,7 @@ interface IconComposition {
     prefix?: string;
 }
 
-export interface IconProps extends QAProps, DOMProps {
+export interface IconProps extends QAProps, DOMProps, ColorTextBaseProps {
```
amje: Let's set explicitly the color prop instead of inherit so we wouldn't add unwanted props by accident

PR #2503, opened 2025-11-17
fix: add isomorphic version of `useLayoutEffect` hook to fix SSR warnings

monochromer:
We use server side rendering and `useLayoutEffect` spam into our logs with message:

```
useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. 
This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. 
To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. 
See https://reactjs.org/link/uselayouteffect-ssr for common fixes.
```

### DIFF
--- a/src/components/TreeList/__stories__/stories/WithDndListStory.tsx
+++ b/src/components/TreeList/__stories__/stories/WithDndListStory.tsx
@@ -10,6 +10,7 @@ import type {
     OnDragEndResponder,
 } from '@hello-pangea/dnd';
 
+import {useLayoutEffect} from '../../../../hooks';
 import {Icon} from '../../../Icon';
 import {ListContainerView, ListItemView, useList} from '../../../useList';
 import type {ListItemViewProps} from '../../../useList';
@@ -54,7 +55,7 @@ export const WithDndListStory = (storyProps: WithDndListStoryProps) => {
         getItemId: ({id}) => id,
     });
 
-    React.useLayoutEffect(() => {
+    useLayoutEffect(() => {
         containerRef?.current?.focus();
     }, []);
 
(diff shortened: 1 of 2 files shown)


src/hooks/private/useIsomorphicLayoutEffect/useIsomorphicLayoutEffect.ts
```diff
@@ -0,0 +1,4 @@
+import * as React from 'react';
+
+export const useIsomorphicLayoutEffect =
+    typeof window !== 'undefined' ? React.useLayoutEffect : React.useEffect;
```
ValeraS:
```suggestion
export const useLayoutEffect = typeof window === 'undefined' ? () => {} : React.useLayoutEffect;
```

Using the name `useLayoutEffect` allows to lint the hook's dependencies. There is also no need to call useEffect as it has no effect on the server.

src/components/theme/ThemeProvider.tsx
```diff
@@ -66,7 +67,7 @@ export function ThemeProvider({
 
     const prevRootClassName = React.useRef('');
 
-    React.useLayoutEffect(() => {
+    useIsomorphicLayoutEffect(() => {
```
ValeraS: There are other places where `useLayoutEffect` is used.

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -2,6 +2,8 @@
 
 import * as React from 'react';
 
+import {useLayoutEffect} from 'src/hooks/private';
```
ValeraS:
```suggestion
import {useLayoutEffect} from '../../hooks/private';
```
path should be relative

src/hooks/useLayoutEffect/useLayoutEffect.ts
```diff
@@ -0,0 +1,3 @@
+import * as React from 'react';
```
ValeraS, line 1: Let's move the file to public hooks and rename it to `useLayoutEffect.ts`.

ValeraS (approved):

PR #2508, opened 2025-11-18
fix(a11y): add accessibility for dropdown submenu via tab

### LINKED ISSUE #2507: DropdownMenu: accessibility fix
If a DropdownMenu contains a submenu, it currently opens on hover when the user moves the mouse over the item that triggers the submenu. However, the submenu is not accessible from the keyboard because it has no keyboard event handlers to open it.

As a result, this makes the submenu inaccessible for screen reader users — the submenu simply cannot be reached without a mouse.

I propose adding the necessary event handlers so that the submenu can also be opened via click (which works with Enter on the focused element and through standard screen reader interaction patterns).

### Reproduction

https://stackblitz.com/edit/gravity-ui-uikit-example-jpx7ukl6?file=src%2FApp.tsx

### Steps to reproduce

Focus the button that opens submenu using tab. Than try to open submenu from keyboard, using Enter, Space etc.

Expected: opened menu
Receiver: not opened menu :(

### Environment

MacOS, yandex browser

### Validations

PosikBoy:
Closes #2507 

If a DropdownMenu contains a submenu, it currently opens on hover when the user moves the mouse over the item that triggers the submenu. However, the submenu is not accessible from the keyboard because it has no keyboard event handlers to open it.

As a result, this makes the submenu inaccessible for screen reader users — the submenu simply cannot be reached without a mouse.

I propose adding the necessary event handlers so that the submenu can also be opened via click (which works with Enter on the focused element and through standard screen reader interaction patterns).

### DIFF
--- a/src/components/DropdownMenu/__tests__/DropdownMenu.test.tsx
+++ b/src/components/DropdownMenu/__tests__/DropdownMenu.test.tsx
@@ -1,0 +1,317 @@
+import userEvent from '@testing-library/user-event';
+
+import {fireEvent, render, screen, waitFor} from '../../../../test-utils/utils';
+import {DropdownMenu} from '../DropdownMenu';
+
+describe('DropdownMenu keyboard navigation', () => {
+    const mockAction = jest.fn();
+    const mockSubAction = jest.fn();
+
+    const menuItems = [
+        {
+            text: 'Item without submenu',
+            action: mockAction,
+        },
+        {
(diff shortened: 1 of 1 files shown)


PosikBoy:
> @PosikBoy Hi! Could add test for this case?

Added tests for component, including based test cases and my new test case with tab using.

amje (approved):

korvin89 (approved):

Issue #2514, opened 2025-11-24
Accordion view in Storybook


KostyaAvtushko:
### Describe the bug

After update from 7.26.0 to 7.26.1 Accordion stories started looking strange without padding. Seems like it happened after some package update

### Reproduction

https://preview.gravity-ui.com/uikit/?path=/story/components-data-display-accordion--playground

### Environment

macos, yandex browser

### Validations

- [x] Check that there isn't [already an issue](https://github.com/gravity-ui/uikit/issues) that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #2516, opened 2025-11-25
chore: fix Accordion storybook

KostyaAvtushko:
After react-docgen-typescript there was in an issue with duplicated items inside Accordion. The package looks a little bit outdated if you check the versions of its deps and it does not look like a package that has good support by maintainers. So, I think we can try to switch to react-docgen(it also supports TypeScript as far as I understand) which is supported by reactjs team, and does not cause problems with Accordion's story

UPD: May the best solution is to get rid of displayName=DisclosureSummary


KostyaAvtushko:
related to https://github.com/gravity-ui/uikit/issues/2514

KostyaAvtushko:
> This breaks plenty of Storybook controls. See Button stories for example

So, seems like fixed version of `react-docgen-typescript` can be a solution for now (I dont see any alternative solutions at the moment). May be you have some thoughts?

Issue #2459, opened 2025-10-22
Fix types to work with React 19


amje:
### Description

Currenty some types in components are incompatible with React 19. For instance, there are many places with `React.RefObject<SomeType>` which allow it to pass into `ref` prop. But in React 19 the interface has changed from this:
```ts
interface RefObject<T> {
    readonly current: T | null;
}
```

to this:
```ts
interface RefObject<T> {
    current: T;
}
```

So it doesn't allow to pass our type to `ref` prop anymore, because it requires `null` to be presented in generic.

Possible fix: every `RefObject<T>` rewrite as `RefObject<T | null>`

### Validations

- [x] Check that there isn't [already an issue](https://github.com/gravity-ui/uikit/issues) that pursue the same goal to avoid creating a duplicate.

amje:
@e-yushkiavichus Ok. Awesome

PR #2522, opened 2025-12-02
fix: update types for React 19 compatibility

e-yushkiavichus:
[#2459](https://github.com/gravity-ui/uikit/issues/2459)


src/components/Table/Table.tsx
```diff
@@ -422,7 +422,11 @@ export class Table<I extends TableDataItem = Record<string, string>> extends Rea
                         return (
                             <th
                                 key={id}
-                                ref={this.state.columnHeaderRefs[index]}
+                                ref={
+                                    this.state.columnHeaderRefs[
+                                        index
+                                    ] as React.Ref<HTMLTableCellElement>
```
ogonkov, line 428: Why we need to cast type here?

e-yushkiavichus: `columnHeaderRefs` is typed as `React.RefObject<HTMLTableCellElement | null>[]` (to align with React 19, where `RefObject<T>` no longer allows `null` in `current` without explicit typing). When accessing `this.state.columnHeaderRefs[index]`, we get `RefObject<HTMLTableCellElement | null>`, while the `ref` prop on `<th>` expects `Ref<HTMLTableCellElement>` (which is `RefCallback<T> | RefObject<T> | null`).

TypeScript treats `RefObject<HTMLTableCellElement | null>` as structurally incompatible with `RefObject<HTMLTableCellElement>` (even though they're functionally equivalent at runtime, since both have `current: HTMLTableCellElement | null`). The assertion bridges this structural type mismatch.

src/components/Modal/Modal.tsx
```diff
@@ -180,7 +180,10 @@ function ModalComponent({
         onOpenChange: handleOpenChange,
     });
 
-    const handleFloatingRef = useForkRef<HTMLDivElement>(refs.setFloating, floatingRef);
+    const handleFloatingRef = useForkRef<HTMLDivElement>(
+        refs.setFloating,
+        floatingRef as React.Ref<HTMLDivElement>,
```
amje, line 188: Could you please add the following comment in every place where you have added such type casting?
```js
// TODO: Remove casting in React 19 (https://github.com/gravity-ui/uikit/issues/2537)
```

amje (changes requested):

src/components/Modal/Modal.tsx
e-yushkiavichus: Done!

Issue #2521, opened 2025-12-01
Raised with action cards not matching figma design


despectra:
### Describe the bug

According to the Gravity UI design system figma markup:
 
- The shadow of this type of card should look more airy for card size = L
- It should also have the 1px stroke of color `line.generic`

<img width="1182" height="1005" alt="Image" src="https://github.com/user-attachments/assets/aeea836c-eff1-467f-802e-69779b936ba9" />

Hovewer, the actual component with this setup doesn't match the design:

<img width="848" height="690" alt="Image" src="https://github.com/user-attachments/assets/12cbffba-faca-4c91-87f9-909f41c2a9f2" />

https://github.com/user-attachments/assets/0976aa85-3dc1-4193-99ee-8fa43db4871b

### Reproduction

https://codesandbox.io/p/devbox/elegant-panini-nfs4mx

### Steps to reproduce

- Open the reproduction link
- Hover the L-sized card with type="action" and view="raised"

### Environment

- MacOS Tahoe 26.1
- Yandex Browser 25.10.0.2546 (64-bit)

### Validations

- [x] Check that there isn't [already an issue](https://github.com/gravity-ui/uikit/issues) that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #2526, opened 2025-12-06
fix: raised card design

tungulin:
- #2521

Added border for Card variant raised, fixed shadows for Card variant raised size l. According to the design of [Figma](https://www.figma.com/design/yk6gP9xvXV5Li99hLuXkgF/Gravity-UI-Design-System--Community-?node-id=53349-612868&t=HTykaa36KSIO9sEN-4)


src/components/Card/Card.scss
```diff
@@ -84,6 +87,12 @@ $block: '.#{variables.$ns}card';
                 content: '';
             }
         }
+
+        @at-root &#{$block}_size_l {
```
amje, line 94: Let's also move `--_--box-shadow: 0px 1px 5px var(--g-color-sfx-shadow);` declaration from above under `@at-root &#{$block}_size_m` section here to be symmetrical with `&_type_container` styles below

amje (changes requested):

src/components/Card/Card.scss
tungulin: fixed

PR #2540, opened 2025-12-23
fix(Dialog): remove unnecessary overflow on body

amje:


src/components/Dialog/DialogBody/DialogBody.scss
```diff
@@ -5,11 +5,14 @@ $block: '.#{variables.$ns}dialog-body';
...
     flex: 1 1 auto;
-    overflow: auto;
     transition: height 0.35s ease-in-out;
 
     &_has-borders {
         border-block-start: 1px solid var(--g-color-line-generic);
         border-block-end: 1px solid var(--g-color-line-generic);
     }
+
+    #{variables.$block}_has-scroll & {
```
Alutery, line 13: Is this necessary?

amje: What do u mean exactly? The `overflow` is only needed when this mod is on

Alutery: I just see a build error because of this

amje: I see. I've commited over unsubmitted changes. Fixed!

PR #2550, opened 2026-01-17
fix: replace barrel imports for useLayoutEffect

DarkGenius:
## Summary

  Fixed `TypeError: useLayoutEffect is not a function` error that occurs when `@gravity-ui/uikit` is used with Next.js `optimizePackageImports` option.

  ## Problem

Error found when updating packages in `gravity-ui/landing` ([PR](https://github.com/gravity-ui/landing/pull/547)).

screenshot

When Next.js is configured with:
```js
experimental: {
  optimizePackageImports: ['@gravity-ui/uikit'],
}
```
the barrel imports (`import { useLayoutEffect } from '../..'`) in hooks are not correctly resolved, causing `useLayoutEffect` to be undefined at runtime.

This breaks components like `DropdownMenu`, `Select`, and others that use `useListNavigation` internally.

### DIFF
--- a/src/hooks/useListNavigation/useListNavigation.ts
+++ b/src/hooks/useListNavigation/useListNavigation.ts
@@ -1,7 +1,7 @@
 import * as React from 'react';
 
-import {useLayoutEffect} from '../..';
 import {KeyCode} from '../../constants';
+import {useLayoutEffect} from '../useLayoutEffect';
 
 import {moveBack} from './moveBack';
 import {moveForward} from './moveForward';


src/hooks/private/useCheckbox/useCheckbox.ts
```diff
@@ -1,8 +1,9 @@
 import * as React from 'react';
 
-import {useControlledState, useForkRef, useLayoutEffect} from '../..';
+import {useControlledState, useForkRef} from '../..';
```
ValeraS: Why does `useLayoutEffect` break the build, but these imports don't?

src/hooks/useListNavigation/useListNavigation.ts
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
 
-import {useLayoutEffect} from '../..';
```
ValeraS, line 3: I think the problem is only with `useListNavigation`, because this import creates circular dependencies, and the other imports work fine.

ValeraS (changes requested):
Do you still need to use `experimental.optimizePackageImports` for `@gravity/uikit`? I believe the tree-shaking problem has been resolved (#2324) and this option is no longer needed. Could you check this, please?

src/hooks/useListNavigation/useListNavigation.ts
DarkGenius: Agree. I've removed the changes in other files.

imsitnikov:
> > Do you still need to use `experimental.optimizePackageImports` for `@gravity/uikit`? I believe the tree-shaking problem has been resolved (#2324) and this option is no longer needed. Could you check this, please?
> 
> @imsitnikov Could you check this, please?

@dgaponov do you remember the reason of adding `optimizePackageImports` for uikit?

Issue #2402, opened 2025-09-03
New animation for Disclosure


KostyaAvtushko:
### Description

The recently added Accordion component, which internally uses Disclosure, has an animation issue. The current opacity animations during opening/closing cause an intermediate jump/glitch inside the Accordion.

It seems we need to create a new height-based animation for Disclosure and implement it in the Accordion component as well.

### Validations

- [x] Check that there isn't [already an issue](https://github.com/gravity-ui/uikit/issues) that pursue the same goal to avoid creating a duplicate.

PR #2551, opened 2026-01-18
feat(Disclosure): add height animation

KostyaAvtushko:
Related to https://github.com/gravity-ui/uikit/issues/2402

## Summary by Sourcery

Animate Disclosure content height with CSS transitions and update associated behavior and tests.

New Features:
- Add height-based expand/collapse animation for Disclosure content using CSS transitions.

Enhancements:
- Refine Disclosure DOM structure and visibility handling to support smooth animations while preserving accessibility semantics.
- Simplify Accordion tests to assert expanded state via ARIA attributes instead of Disclosure content CSS classes.
- Remove Accordion-specific overrides for legacy Disclosure exit/enter animations now that Disclosure uses height transitions.

Tests:
- Extend Disclosure tests to cover accessibility tree visibility during expand/collapse and keepMounted behavior.
- Update Accordion tests to assert behavior via ARIA attributes and new Disclosure implementation rather than animation classes.


src/components/Disclosure/Disclosure.scss
```diff
@@ -41,38 +43,21 @@ $gap: 8px;
...
-    }
-    &__content_visible {
-        display: block;
-    }
-    &__content#{$block}_exit_active {
         display: block;
-        animation-name: #{variables.$ns}disclosure-collapsed;
-        animation-duration: 0.1s;
+        max-height: 0;
+        overflow: hidden;
```
Raubzeug, line 50: Could we avoid keeping `overflow: hidden` after the disclosure is fully expanded? Unlike the previous implementation, the expanded content can now clip focus outlines, shadows, and non-portaled overlays. Could overflow clipping be limited to the entering and exiting phases?

src/components/Disclosure/DisclosureDetails/DisclosureDetails.tsx
```diff
@@ -14,31 +12,100 @@ export interface DisclosureDetailsProps extends QAProps {
...
         >
-            <div
-                ref={containerRef}
-                id={ariaControls}
-                role="region"
-                aria-labelledby={ariaLabelledby}
-                className={b('content', {visible: expanded}, className)}
-                data-qa={qa || DisclosureQa.DETAILS}
-            >
+            <div ref={innerRef} className={className}>
```
Raubzeug: Is moving `className` from the region element to the inner measurement wrapper intentional? This changes the DOM styling contract, and vertical margins applied through `className` are not included in `offsetHeight`, so they may be clipped by the outer `overflow: hidden` container. Could we preserve `className` on the public root or measure a height that includes the complete rendered box?

src/components/Disclosure/DisclosureDetails/DisclosureDetails.tsx
```diff
@@ -14,31 +12,100 @@ export interface DisclosureDetailsProps extends QAProps {
...
+        <div
+            {...hiddenAttributes}
+            aria-hidden={expanded ? undefined : true}
+            id={ariaControls}
+            role="region"
+            aria-labelledby={ariaLabelledby}
+            className={b('content', {visible})}
+            data-qa={qa || DisclosureQa.DETAILS}
+            style={style as React.CSSProperties}
+            onTransitionEnd={handleTransitionEnd}
```
Raubzeug: Should this animation respect `prefers-reduced-motion`? If the transition is disabled by that preference or by application CSS, `transitionend` will not fire and `keepMounted={false}` content will never reach the `hidden` state and unmount. Could we add an explicit no-transition completion path?

src/components/Disclosure/DisclosureDetails/DisclosureDetails.tsx
```diff
@@ -14,31 +12,100 @@ export interface DisclosureDetailsProps extends QAProps {
...
+
+            if (keepMounted) {
+                return prev === 'hidden' ? 'hidden' : 'exiting';
+            }
+
+            return prev === 'hidden' ? 'hidden' : 'exiting';
+        });
+    }, [expanded, keepMounted]);
+
+    React.useEffect(() => {
```
Raubzeug: Could we simplify the visibility choreography here? The first effect moves the state to `entering`, while the second immediately moves it to `visible`, so `entering` does not actually survive until `transitionend`. Would a single transition effect or reducer with an explicit animation-frame boundary make the state machine easier to reason about?

Raubzeug (changes requested):
Thanks for working on this. I found several issues that I think should be addressed before merging; please see the inline comments.

Two additional questions do not have a clean changed-line anchor:

- Could we add a browser-level test for the actual height transition, including a content resize and unmount after a real `transitionend`? The current tests manually dispatch the event, while most Accordion assertions check only `aria-expanded`, so the measured-height and real CSS transition paths are not exercised.

- Could we verify the documented `<Disclosure.Details>` composition? `prepareChildren` wraps every non-Summary child in another `DisclosureDetails`, so an explicit Details element creates nested regions with duplicate IDs and, after this change, two height observers and transitions. Should explicit Details be detected and reused instead?

src/components/Accordion/AccordionItem/AccordionItem.tsx
```diff
@@ -82,7 +82,9 @@ export const AccordionItem = React.forwardRef<HTMLDivElement, AccordionItemProps
                 )}
             >
                 {preparedSummary}
-                <Disclosure.Details className={accordionDetailsBlock}>{details}</Disclosure.Details>
+                <Disclosure.Details>
```
Raubzeug: Moving `accordionDetailsBlock` to a child fixes Accordion, but it exposes a generic regression for existing `<Disclosure.Details className="...">` consumers. In the collapsed `keepMounted` state, the public region is only given `height: 0`, so its own padding, borders, and margins can still render or occupy space. Before this PR, the collapsed region used `display: none`, so those styles disappeared. Could we make the transition collapse the complete public root box without requiring every consumer to add an inner wrapper?

src/components/Disclosure/__tests__/Disclosure.visual.test.tsx
```diff
@@ -1,13 +1,82 @@
...
+                    );
+                }
+            });
+        });
+
+        const summary = host.getByRole('button', {name: 'Toggle'});
+        await summary.click();
+
+        const details = host.getByRole('region');
+        await expect.poll(() => host.getAttribute('data-transition-count')).toBe('1');
```
Raubzeug, line 38: This resize happens only after the opening transition has completed and the inline height has been cleared, so it verifies ordinary `height: auto` layout rather than adaptation of the animated height. During `entering`, the target is measured only once in `onEntering`; if the content grows before the transition completes, it remains clipped at the stale pixel height and then jumps when `onEntered` clears the inline height. Could we move the resize into the entering phase and either update the target height or explicitly test the intended behavior?

Raubzeug: Could we restore the resize retargeting that existed in `faf9108`? The merge commit removed `waitForStableTransition`, and the final test now asserts that the container keeps `initialTargetHeight` after the content has already grown. When the transition finishes, clearing the inline height jumps directly to `resizedHeight`, so the issue raised in this thread is still present rather than fixed. Please update the active height target during the entering phase and assert the new target before completion.

Raubzeug (approved):

PR #2553, opened 2026-01-19
fix(List): add aria attributes

### LINKED ISSUE #2552: List: Add accessibility in filtering list
### Description

I am using a filtered list controlled by an input field.
However, the input does not have the required ARIA attributes, so screen readers cannot announce the currently active item in the list.
Although users can navigate through the list items using arrow keys, the active option is not properly exposed to assistive technologies.

### Suggested solution

Add the necessary ARIA attributes to the input to correctly describe its relationship with the list and the active item:

role="combobox" — identifies the input as a combobox control

aria-activedescendant — points to the ID of the currently active list item

This ensures that screen readers can correctly announce the active item while navigating the list with the keyboard. 

### Alternative

_No response_

### Additional context

_No response_

### Validations

PosikBoy:
Closes #2552

### DIFF
--- a/src/components/List/components/ListItem.tsx
+++ b/src/components/List/components/ListItem.tsx
@@ -14,6 +14,8 @@ import {getElementId} from '../utils';
 
 const b = block('list');
 
+const ROLES_WITH_ARIA_SELECTED = new Set(['option', 'gridcell', 'row', 'tab']);
+
 export const defaultRenderItem = <T extends unknown>(item: T) => String(item);
 
 function getStyle(provided?: DraggableProvided, style?: React.CSSProperties) {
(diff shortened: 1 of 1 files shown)


src/components/List/components/VariableSizeListElementType.tsx
```diff
@@ -0,0 +1,8 @@
+import * as React from 'react';
```
amje, line 3: Missing `'use client'` directive

src/components/List/List.tsx
```diff
@@ -558,6 +570,8 @@ export class List<T = unknown> extends React.Component<ListProps<T>, ListState<T
     private handleKeyMove(event: React.KeyboardEvent, step: number, defaultItemIndex = 0) {
         const {activeItem = defaultItemIndex} = this.state;
 
+        // console.log(activeItem);
+
```
amje: Let's delete this

src/components/List/components/ListItem.tsx
```diff
@@ -59,7 +60,7 @@ export class ListItem<T = unknown> extends React.Component<ListItemProps<T>> {
             // eslint-disable-next-line jsx-a11y/click-events-have-key-events
             <div
                 role={role}
-                aria-selected={selected}
+                aria-selected={role === 'option' ? selected : undefined}
```
korvin89: Let's list all the roles for which this attribute is available here https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-selected#associated_roles

PR #2555, opened 2026-01-20
feat(Card): added aria-checked state and changed role

### LINKED ISSUE #2554: Card: add aria-checked and appropriate role to card with selection type
### Description

For now, card with selection type has no state, that shows to screen reader that card is selected.

### Suggested solution

I suggest adding role radio and aria-checked attribute to card with type selection

### Alternative

_No response_

### Additional context

_No response_

### Validations

PosikBoy:
Closes #2554


src/components/Card/Card.tsx
```diff
@@ -37,6 +37,11 @@ export interface CardProps extends Omit<BoxProps<'div'>, 'as' | 'onClick'> {
     size?: CardSize;
 }
 
+const roles = {
+    selection: 'radio',
+    action: 'button',
```
ogonkov, line 42: I'm not sure that role button could have checked state

PosikBoy: You are right: role="button" supports aria-pressed. 
So I used aria-checked only for elements with role="radio".

`aria-checked={isTypeSelection ? selected : undefined}`

PR #2561, opened 2026-01-23
fix(Dialog): moved close button to the top of dialog in dom, add "as" prop to DialogHeader

### LINKED ISSUE #2560: Dialog: «Close» button located at the end of the dialog window and caption is not marked as heading
### Describe the bug

1) The close button is positioned at the end of the DOM tree. When a screen reader user navigates into the dialog, they must tab through numerous elements before reaching it, creating a poor user experience.
I suggest moving the close button to the top of the dialog window.

2) Heading in dialog not marked as heading
Suggest mart it as h2 

### Reproduction

https://stackblitz.com/edit/gravity-ui-uikit-example-camd3mlr?file=src%2FApp.tsx

### Steps to reproduce

_No response_

### Environment

macos, yandexBrowaer

### Validations

PosikBoy:
Closes #2560


src/components/Dialog/DialogHeader/DialogHeader.tsx
```diff
@@ -20,9 +20,9 @@ export function DialogHeader(props: DialogHeaderProps) {
     return (
         <div className={b(null, className)}>
             {insertBefore}
-            <div className={b('caption')} id={id}>
+            <h2 className={b('caption')} id={id}>
```
Alutery: Maybe use Text component and add `as` field to DialogHeaderProps? And return div as default?

Alutery: Thanks!
Please update the PR title to reflect the last changes

src/components/Dialog/DialogHeader/DialogHeader.tsx
```diff
@@ -11,18 +12,19 @@ export interface DialogHeaderProps {
...
 
 export function DialogHeader(props: DialogHeaderProps) {
-    const {caption = '', insertBefore, insertAfter, className, id} = props;
+    const {caption = '', as = 'div', insertBefore, insertAfter, className, id} = props;
 
     return (
         <div className={b(null, className)}>
             {insertBefore}
-            <div className={b('caption')} id={id}>
+            <Text as={as} className={b('caption')} id={id}>
```
amje: Let's use `variant` prop `subheader-3` instead of setting font styles in CSS

PosikBoy:
> @PosikBoy Is it best practice to put close button first in tab order?

Yes, according to Yandex Inclusion team

src/components/Dialog/DialogHeader/DialogHeader.tsx
```diff
@@ -11,18 +12,19 @@ export interface DialogHeaderProps {
...
 
 export function DialogHeader(props: DialogHeaderProps) {
-    const {caption = '', insertBefore, insertAfter, className, id} = props;
+    const {caption = '', as = 'div', insertBefore, insertAfter, className, id} = props;
 
     return (
         <div className={b(null, className)}>
             {insertBefore}
-            <div className={b('caption')} id={id}>
+            <Text as={as} variant="subheader-3" className={b('caption')} id={id}>
```
Alutery, line 25: it looks like the `.caption` in DialogHeader.scss is no longer needed

PosikBoy: I suggest leaving it so that consumers have the opportunity to redefine the styles.

Alutery: even if you leave it in HTML? I'm confused that the mixin duplicates the variant props

PosikBoy: removed reduant styles

PR #2572, opened 2026-02-09
fix(MenuItem): unstyled a tag replaced with Link component

ukqfhf:
Summary
This PR replaces an unstyled <a> tag with the Link component in the MenuItem component, so the links can be distinguished from regular MenuItems.

Why
In the current implementation, links cannot be distinguished from MenuItems that do not contain links.

### DIFF
--- a/src/components/Menu/MenuItem.tsx
+++ b/src/components/Menu/MenuItem.tsx
@@ -3,6 +3,7 @@
 import * as React from 'react';
 
 import {useActionHandlers} from '../../hooks';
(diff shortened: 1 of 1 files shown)


amje:
@ukqfhf That's by design not a Link. By default, link item and non-link item look the same, it's up to the user adding visiual difference to them

ukqfhf:
How about providing a render prop — as many modern UI kits do (e.g., Base UI) — so that if it’s supplied, the li simply invokes it?

PR #2573, opened 2026-02-09
chore: testing infra moved to playwright-tools

SwinX:
Removed definitions of custom fixtures for mount and screenshot Updated screenshots for all components

- [x] remove webkit
- [x] remove fix for popup styles

### DIFF
--- a/src/components/Popup/Popup.scss
+++ b/src/components/Popup/Popup.scss
@@ -46,8 +46,8 @@ $transition-distance: 10px;
         border-end-end-radius: inherit;
     }
 
-    @at-root [data-floating-ui-status='open'] > &:not(&_disable-transition),
-        [data-floating-ui-status='close'] > &:not(&_disable-transition) {
+    @at-root [data-floating-ui-status='open'] > &,
+        [data-floating-ui-status='close'] > & {
         transition-duration: $transition-duration;
     }
 


src/components/Accordion/__snapshots__/Accordion.visual.test.tsx-snapshots/Accordion-render-showcase-dark-chromium-linux.png
ogonkov, line 1: Please, can you move screenshots instead of deleting them in history?

SwinX: Seems this will not make a lot of sence.

Screenshots naming changed cause uikit moved to new screenshot fixture. So the suffix in name changed from light|dark to 1|2. This is because we're not specifically distinguish screenshots, but use standard playwright mechanics: to add index to each screenshot in test.

We may try to add possibility to customize this suffix. However practically this seems not really needed feature.

Also I updated screenshots anyway
 
And the last thing is that why shall anyone try review commit history for screenshots? For code this totally makes sense. But why for screenshots?

ogonkov: Well, i don't propose to change naming scheme, i'm asking to preserve history with making `git mv` for each renamed screenshot

This will help to examine if they change or not with this migration

SwinX: I understand. What I'm trying to say that since this was not manual rename moving would be pretty hard to do. There is no consistent pattern in rename. And There is around 1000 files.

But my main point is that preserving history for snapshot makes little sense. Why would anybody try to review it?

src/components/Popup/Popup.scss
```diff
@@ -46,8 +46,8 @@ $transition-distance: 10px;
         border-end-end-radius: inherit;
     }
 
-    @at-root [data-floating-ui-status='open'] > &,
-        [data-floating-ui-status='close'] > & {
+    @at-root [data-floating-ui-status='open'] > &:not(&_disable-transition),
+        [data-floating-ui-status='close'] > &:not(&_disable-transition) {
```
amje: We already disabling transition above

playwright/actions/stabilize-floating-ui.ts
```diff
@@ -0,0 +1,42 @@
+import type {Page} from '@playwright/test';
+
+const STABILIZER_ID = '__floating-stabilizer';
+const FLOATING_SELECTOR = '[data-floating-ui-status]';
+
+/**
+ * Converts floating-ui's GPU-composited transform positioning to CPU-rendered top/left.
```
ogonkov, line 7: Why we need to convert them?

SwinX: This is a golden kludge to stabilize popup-based test

In Gravity UI popups are based on floating-ui library. In case of high DPI screens floating-ui applies transform styles to element. In this case rendering of transformed layer moved to GPU. Font rendering on GPU is not stable and produces slightly different results each time.

The recommended playwright screenshot matcher `toHaveScreenshot` takes [several consecutive images](https://playwright.dev/docs/test-snapshots#generating-screenshots) of the page until it receives 2 exact images. Because of unstable rendering playwright unable to receive 2 exact images so popup-based tests are extremely flacky

This hack is not needed if we will set low DPI screen (deviceScaleFactor < 1.5). After discussion with @amje we concluded that we may miss some important details if we will take smaller images for tests.

Issue #2574, opened 2026-02-10
Small Update for Select Component


getDimaMal:
### Description

In many cases, when we need to render option elements that also depend on the active state, we do "some hacks")

### Suggested solution

And in my opinion, it’s better to solve this by updating the props in renderOption and providing an isItemActive property.

### Validations

- [x] Check that there isn't [already an issue](https://github.com/gravity-ui/uikit/issues) that request the same feature to avoid creating a duplicate.

PR #2575, opened 2026-02-10
feat(Select): add IsItemActive Prop for renderOption

getDimaMal:
ISSUE: https://github.com/gravity-ui/uikit/issues/2574

In many cases, when we need to render option elements that also depend on the active state, we do "some hacks")
And in my opinion, it’s better to solve this by updating the props in renderOption and providing an isItemActive property.

### DIFF
--- a/src/components/Select/types.ts
+++ b/src/components/Select/types.ts
@@ -46,7 +46,7 @@ export type SelectRenderControl<T extends HTMLElement = HTMLElement> = (
 
 export type SelectRenderOptionViewParams = {
     itemHeight: number;
-    isItemActive?: boolean;
+    isItemActive?: boolean; // FIXME: make this field required in the next major
 };
 
 export type SelectRenderOption<T> = (


korvin89:
@getDimaMal Hey! Add test for this case please

src/components/Select/__tests__/Select.popup.test.tsx
```diff
@@ -95,11 +95,14 @@ describe('Select popup', () => {
 
             expect(renderOptionGroup).toHaveBeenCalledWith(
                 expect.anything(),
-                expect.objectContaining({itemHeight: height}),
+                expect.objectContaining({itemHeight: height, isItemActive: expect.any(Boolean)}),
```
korvin89: Let's not touch the old tests, but just add one new one

src/components/Select/__tests__/Select.popup.test.tsx
```diff
@@ -148,6 +148,48 @@ describe('Select popup', () => {
...
+        await user.click(selectControl);
+
+        const activeCall = renderOption.mock.calls.find(
+            (call) => call[1] && (call[1] as {isItemActive?: boolean}).isItemActive,
+        );
+        const inactiveCall = renderOption.mock.calls.find(
+            (call) => call[1] && !(call[1] as {isItemActive?: boolean}).isItemActive,
+        );
+        expect(activeCall).toBeDefined();
+        expect(inactiveCall).toBeDefined();
```
korvin89: Can we check the actual value here? (`true` / `false`)

src/components/Select/types.ts
```diff
@@ -46,6 +46,7 @@ export type SelectRenderControl<T extends HTMLElement = HTMLElement> = (
 
 export type SelectRenderOptionViewParams = {
     itemHeight: number;
+    isItemActive: boolean;
```
korvin89: My apologies for not mentioning this earlier — let’s make this field optional for now. Since the `SelectRenderOptionViewParams` type is public, making the field required would cause type errors for anyone using it directly after they update the library. You can also add the following comment:
```js
// FIXME: make this field required in the next major
```

src/components/Select/__tests__/Select.popup.test.tsx
```diff
@@ -148,6 +148,48 @@ describe('Select popup', () => {
...
+            options: GROUPED_OPTIONS,
+        });
+
+        const user = userEvent.setup();
+        const selectControl = getByTestId(TEST_QA);
+        await user.click(selectControl);
+
+        expect(renderOptionGroup).toHaveBeenCalledWith(
+            expect.anything(),
+            expect.objectContaining({isItemActive: expect.any(Boolean)}),
```
korvin89: I think it's worth doing the same check for the group as for the usual options

korvin89 (approved):

PR #2580, opened 2026-02-12
fix(Label): move key-value separator to left part of element

goshander:


src/components/Label/Label.scss
```diff
@@ -68,14 +68,17 @@ $transition-timing-function: ease-in-out;
...
+
     &__value {
         display: flex;
-        opacity: 0.7;
         overflow: hidden;
     }
 
     &__separator {
-        margin: 0 4px;
+        margin: 0 4px 0 2px;
```
ukqfhf, line 81: Would it be better to make these values token‑based?

goshander: Could you specify which tokens you mean?

ukqfhf: `--g-spacing-half` for 2px and `--g-spacing-1` or `--g-spacing-base` for 4px. At the moment, these values cannot be customized. Of course, it would be better if these values were relative like `em`, in this case, if the font size increases, then the distance from the separator should increase also.

ukqfhf: Even better to specify component specific tokens like `--comp-label-separator-margin` or something at the top element to enable deep customization

PR #2582, opened 2026-02-12
fix(Select): close popup after selecting option via keyboard in single filterable select

getDimaMal:
There was an issue when the Selector had filterable="true". If a user navigated the list with the Up/Down arrow keys and then pressed Enter, the value would be selected but the Popup wouldn’t close until the second key press.

After investigating, I found that in a few places we call ref.current.focus() and then handle keyboard events, which causes the Popup to reopen.

The simplest fix is to stop event propagation for the Enter key event.

Before:
https://github.com/user-attachments/assets/c76efcc1-6471-48d0-813d-883810146047

After:
https://github.com/user-attachments/assets/f8b762bb-66d3-4841-b09c-fb2643a10105

### DIFF
--- a/src/components/Accordion/AccordionItem/AccordionItem.tsx
+++ b/src/components/Accordion/AccordionItem/AccordionItem.tsx
@@ -44,6 +44,7 @@ export const AccordionItem = React.forwardRef<HTMLDivElement, AccordionItemProps
             summary,
             onUpdate,
             qa,
+            className,
         } = props;
 
         const attributes = useAccordion();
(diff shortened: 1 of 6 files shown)


getDimaMal:
This fix works correctly with multiple values as well

https://github.com/user-attachments/assets/530bcf33-35ca-4b96-a699-c6b078f6c2ea

src/components/Select/components/SelectList/SelectList.tsx
```diff
@@ -16,7 +17,7 @@ import './SelectList.scss';
 
 type SelectListProps = {
     mobile: boolean;
-    onOptionClick: (option: FlattenOption) => void;
+    onOptionClick: ItemClickHandler<FlattenOption>;
```
korvin89: Let's process it here in the general handler https://github.com/gravity-ui/uikit/blob/d24d571ab6d2079510e0a05c123881799d9f4ad9/src/components/Select/Select.tsx#L194

PR #2587, opened 2026-02-13
fix: add 'no-wrap' option to ThemeProvider scoped prop

### LINKED ISSUE #2585: Scoped ThemeProvider but without additional `div`
### Description

I need to set context inside another component (microfrontend-specific case), and want to avoid additional `div`.

I suggest that ThemeProvider should have option for `scoped` mode, that allows to avoid both changing `body` classes and rendering just a children alone (like in default mode)

### Suggested solution

Extend `scoped` prop to be

```
interface {
  scoped?: true | 'no-wrap'
}
```

### Alternative

_No response_

### Additional context

_No response_

### Validations

DarkGenius:
Closes #2585

Adds support for 'no-wrap' value in scoped prop:
- scoped?: boolean | 'no-wrap'
- When 'no-wrap', renders only children without div wrapper
- When true/false, maintains current behavior with div wrapper

This allows microfrontend-specific use cases where theme context 
needs to be set without adding additional DOM elements.

### DIFF
--- a/src/components/theme/types.ts
+++ b/src/components/theme/types.ts
@@ -9,5 +9,5 @@ export interface ThemeContextProps {
     theme: Theme;
     themeValue: RealTheme;
     direction: Direction;
-    scoped?: boolean;
+    scoped?: boolean | 'no-wrap';
 }


DarkGenius:
discarded

PR #2594, opened 2026-02-20
fix(FileDropZone): pass `accept` prop to file input and respect `multiple` in drop handler

dvortsovs:


src/components/lab/FileDropZone/FileDropZone.Provider.tsx
```diff
@@ -72,9 +73,9 @@ export const FileDropZoneProvider = ({
                 files.push(file);
             }
 
-            onUpdate(files);
+            onUpdate(multiple ? files : files.slice(0, 1));
```
korvin89: Here slice looks redundant, it seems files[0] is quite enough.

src/components/lab/FileDropZone/FileDropZone.scss
```diff
@@ -33,7 +33,7 @@ $block: '.#{variables.$ns}file-drop-zone';
     }
 
     &:focus:not(:focus-visible) {
-        border-color: none;
+        border-color: unset;
```
korvin89, line 36: What do we fix with this change?

dvortsovs: property "border-color" don't have value "none"
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/border-color

korvin89 (approved):

PR #2602, opened 2026-02-27
feat(ColorPicker): add optional prop disabled state

Estasie:
Added prop for disabled state and improved some logic with passing empty strings as value through props


src/components/lab/ColorPicker/ColorPicker.scss
```diff
@@ -57,6 +57,12 @@ $block: '.#{variables.$ns}color-picker';
                 --_--border-radius: var(--g-border-radius-xl);
             }
         }
+
+        &_disabled {
+            cursor: default;
```
feelsbadmans: It's unnecessary property, because of `pointer-events: none`

src/components/lab/ColorPicker/ColorPicker.tsx
```diff
@@ -87,16 +92,18 @@ export const ColorPicker = ({
 
     const isInternalUpdateRef = React.useRef(false);
 
+    const effectiveColor = color && color.trim() !== '' ? color : DEFAULT_COLOR;
```
feelsbadmans: ```color?.trim() || DEFAULT_COLOR;```

feelsbadmans (approved):

PR #2603, opened 2026-03-03
fix: fix types for React 19

amje:


src/components/layout/Container/Container.tsx
```diff
@@ -38,7 +38,7 @@ export interface ContainerProps extends QAProps {
      * By default takes props via `LayoutContext`
      */
     spaceRow?: Space | MediaPartial<Space>;
-    as?: keyof JSX.IntrinsicElements;
+    as?: keyof React.JSX.IntrinsicElements;
```
ogonkov, line 41: May be we need to deny usage of global JSX in eslint?

PR #2614, opened 2026-03-17
feat(FileDropZone): improve file rejection handling and drag feedback

dvortsovs:
## Summary

This PR improves the unstable `FileDropZone` component and the `useDropZone` hook by making file acceptance/rejection handling explicit, adding invalid drag feedback, and expanding test coverage.

## What Changed

- Updated `FileDropZone` callbacks:
  - `onUpdate` now receives both accepted files and rejected files.
  - Added `onUpdateAccepted` for accepted files only.
  - Added `onUpdateRejected` for rejected files only.
- Added typed file rejection data through `DropZoneFileRejection`, including rejection reasons:
  - `invalid-type` for files that do not match the configured MIME types.
  - `too-many-files` for extra files when `multiple` is `false`.
- Made `accept` optional and default it to an empty list, which allows all file types.
- Added invalid drag visual feedback when dragged files do not match accepted MIME types.
- Added `qa` support for the root `FileDropZone` element and a stable QA attribute for the hidden file input.
- Moved file validation into shared utilities and split `FileDropZone` types, constants, and helpers into dedicated files.
- Refactored `useDropZone` so it focuses on drag-and-drop event handling and exposes native drag event callbacks.
- Updated English and Russian documentation for the new API.
- Added unit and visual tests for `FileDropZone`, `useDropZone`, and file separation utilities.

## Notes

The change updates the public unstable API for `FileDropZone` and `useDropZone`, including the `onUpdate` callback signature and exported hook types.


src/components/lab/FileDropZone/__tests__/FileDropZone.visual.test.tsx
```diff
@@ -9,6 +11,22 @@ test.describe('FileDropZone', {tag: '@FileDropZone'}, () => {
...
 
+    test('render story: <Default> clicked', async ({mount, expectScreenshot}) => {
+        const root = await mount(<FileDropZoneStories.Default />);
+
+        await root.click();
+
+        await expectScreenshot();
+    });
+
+    test('render story: <Default> focused', async ({mount, expectScreenshot}) => {
```
amje, line 22: Screenshots from these two tests look the same

src/components/lab/FileDropZone/__tests__/FileDropZone.test.tsx
```diff
@@ -0,0 +1,304 @@
...
+            const zone = screen.getByTestId('drop-zone');
+            expect(zone.className).toMatch(/disabled/);
+        });
+
+        test('does not call onUpdate on drop when disabled', () => {
+            const onUpdate = jest.fn();
+            render(<FileDropZone {...defaultProps} onUpdate={onUpdate} disabled qa="drop-zone" />);
+            const zone = screen.getByTestId('drop-zone');
+
+            fireEvent.dragEnter(zone, createDropData([createFile('a.png', 'image/png')]));
```
amje, line 135: It's better to use user object for firing events (https://github.com/gravity-ui/uikit/blob/7e725f370139f7562925aa715d0773bf95e4f213/src/components/Select/__tests__/Select.clear.test.tsx#L69)

src/components/lab/FileDropZone/types.ts
```diff
@@ -0,0 +1,33 @@
+import type {BaseInputControlProps} from 'src/components/controls/types';
+import type {FileRejection, UseDropZoneAccept} from 'src/hooks/lab/useDropZone';
+
+import type {IconData} from '../..';
```
amje, line 1: Please do not import from index file, use direct import path

src/components/lab/FileDropZone/types.ts
```diff
@@ -0,0 +1,33 @@
+import type {BaseInputControlProps} from 'src/components/controls/types';
```
amje: We don't have any bundler to resolve such aliases. Please use true relative paths

src/components/lab/FileDropZone/FileDropZone.Provider.tsx
```diff
@@ -73,22 +78,59 @@ export const FileDropZoneProvider = ({
...
+        },
+        [onReject],
+    );
+
+    const {isDraggingOver, isInvalidDrag, getDroppableProps} = useDropZone({
         accept,
         disabled,
-        onDrop: handleDrop,
+        multiple,
+        maxFilesCount,
```
amje: DropZone shouldn't know about files, it should be generic

src/components/lab/FileDropZone/types.ts
```diff
@@ -0,0 +1,33 @@
...
+        onUpdate: (files: File[]) => void;
+        onReject?: (items: DropZoneFileRejection[]) => void;
+        title?: string;
+        description?: string;
+        buttonText?: string;
+        icon?: IconData | null;
+        errorIcon?: IconData | null;
+        className?: string;
+        multiple?: boolean;
+        maxFilesCount?: number;
```
amje: Let's not validate inside the component. We cannot predict what the user might need. Leave it for the user choice. These props are redundant and can be converted to smth generic: `maxFilesCount`, `onReject`

amje (changes requested):

src/components/lab/FileDropZone/__tests__/FileDropZone.visual.test.tsx
dvortsovs: These tests verify different actions: one checks focus via Tab, while the other checks clicking on the zone.

src/components/lab/FileDropZone/__tests__/FileDropZone.test.tsx
dvortsovs: I couldn't find a way to replace DnD events using userEvent. So I used this in places, where it possible

src/components/lab/FileDropZone/FileDropZone.tsx
```diff
@@ -58,10 +60,14 @@ const FileDropZoneContent = ({className, children}: FileDropZoneContainerProps)
     return (
         <div
             {...getDroppableProps()}
+            data-qa={qa}
+            aria-disabled={disabled || undefined}
```
korvin89, line 64: **a11y: disabled zone stays in the tab order.** `getDroppableProps()` always returns `tabIndex: 0` and `role="button"`, so even with `aria-disabled` set here the container is still reachable by Tab and announced as a button. Pressing Enter does nothing (`onKeyDown` is disabled), so focus lands "nowhere". Consider forcing `tabIndex={-1}` on the root when `disabled` is set.

src/components/lab/FileDropZone/FileDropZone.scss
```diff
@@ -9,7 +9,6 @@ $block: '.#{variables.$ns}file-drop-zone';
     box-sizing: border-box;
     padding: 16px;
 
-    cursor: pointer;
     outline: none;
```
korvin89, line 13: **UX: `cursor: pointer` was removed, but the whole zone became clickable in this PR.** This seems backwards — previously only the button was the click target, now the entire container handles clicks. It'd be more consistent to *add* `cursor: pointer` to the whole zone (and `not-allowed`/`default` for the `_disabled` modifier) so users see the area is interactive. Was the removal intentional?

src/components/lab/FileDropZone/utils.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    return reasons;
+}
+
+export function getSeparatedItems(dataTransfer: DataTransfer, options: ValidateOptions) {
+    const items = dataTransfer.items;
+    const accepted: DataTransferItem[] = [];
+    const rejected: FileRejection[] = [];
+
+    for (const item of items) {
```
korvin89, line 55: **Potential cross-browser issue: iterating `DataTransferItemList` with `for...of`.** `DataTransferItemList` is not guaranteed to be iterable (no `Symbol.iterator` per spec), so `for (const item of items)` may throw in some browsers. The tests don't catch this because the mocks use a plain array. Since this code is being rewritten anyway, safer to do `Array.from(dataTransfer.items)` first. Same applies to `getFilesFromItems` in the provider.

src/components/lab/FileDropZone/__stories__/constants.ts
```diff
@@ -0,0 +1 @@
+export const FileDropZoneQA = 'file-drop-zone-qa';
```
korvin89: **Naming collision risk:** this `FileDropZoneQA` (string) coexists with `FileDropZoneQa` (the QA object in `../constants.ts`). They differ only by the case of the last letters, which is very easy to confuse. Suggest renaming this story-only constant to something distinct, e.g. `FILE_DROP_ZONE_STORY_QA`. (Also, the `FileDropZoneQa` object would benefit from `as const` so its values are literal types rather than `string`.)

src/components/lab/FileDropZone/FileDropZone.Provider.tsx
```diff
@@ -5,95 +5,194 @@ import {useActionHandlers, useFileInput} from '../../../hooks';
...
+    const handleFileInputUpdate = React.useCallback(
         (files: File[]) => {
-            onUpdate(files);
+            if (multiple || files.length <= 1) {
+                handleUpdate(files, []);
+                return;
+            }
+
+            handleUpdate(
+                files.slice(0, 1),
```
korvin89: **Effectively unreachable branch.** In single-file mode the hidden `<input>` has no `multiple` attribute, so the OS dialog allows selecting only one file — this `too-many-files` slice path can't be hit through normal input usage (only via a synthetic `fireEvent.change` in tests). It also duplicates logic `getSeparatedItems` already does for the drop path. Consider simplifying.

korvin89 (approved):

PR #2625, opened 2026-03-24
feat(DialogFooter): render contents only if they exist

ArturAbdullin:


src/components/Dialog/DialogFooter/DialogFooter.tsx
```diff
@@ -137,19 +137,26 @@ export function DialogFooter(props: DialogFooterProps) {
...
+    const buttons = (() => {
+        if (renderButtons) {
+            return renderButtons(buttonApply, buttonCancel);
+        }
+
+        if (!textButtonCancel && !textButtonApply) {
+            return null;
+        }
+
+        return [buttonCancel, buttonApply];
```
amje: React elements should have a key if they are placed in an array

PR #2630, opened 2026-03-25
fix(Breadcrumbs): render endContent only once to prevent duplicate events

### LINKED ISSUE #2617: Breadcrumbs: endContent is rendered twice, duplicating interactive elements and data-qa attributes
### Describe the bug

`endContent` in Breadcrumbs is rendered twice in the DOM (measurement pass + visible pass), causing any interactive content placed there to be duplicated. This leads to two dialogs opening simultaneously on a single button click. As a side effect, `data-qa` attributes on elements inside `endContent` are also duplicated, causing tests that query by `data-qa` to find multiple elements and fail.

### Reproduction

https://stackblitz.com/edit/gravity-ui-uikit-example-96cxtdwr?file=src%2FApp.css,src%2FApp.tsx

### Steps to reproduce

1. Click the "Open Dialog" button inside the breadcrumbs
2. Observe that two dialogs open simultaneously — the counter inside the dialog shows Open modals count: 2

### Environment

@gravity-ui/uikit@^7.33.0

### Validations

temrjan:
## Summary

Fixes #2617 — `endContent` is rendered twice in the DOM (once in the visible list, once in the hidden measurer container), causing interactive elements to fire events twice and duplicating `data-qa` attributes.

**Root cause:** The measurer `<div aria-hidden="true" inert>` renders a full copy of `endContent` for width calculation. Even though the container is hidden, React still mounts the component tree and attaches event handlers.

**Fix:** Remove `endContent` from the measurer. It is already tracked via `preservedRefs` in the visible list (line 58), so `useCollapseChildren` correctly accounts for its width during collapse calculation.

## Test plan

- [x] Added test: `renders endContent only once in the DOM` — verifies only one button is found by role
- [x] All 15 existing Breadcrumbs tests pass
- [x] ESLint + pre-commit hooks pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.tsx
+++ b/src/components/Breadcrumbs/Breadcrumbs.tsx
@@ -190,11 +190,6 @@ export const Breadcrumbs = React.forwardRef(function Breadcrumbs(
                     {items.map((child, index) =>
                         renderChild(child, index, index === items.length - 1, false),
                     )}
-                    {props.endContent && (
-                        <li key="end-content" className={b('item')}>
-                            {props.endContent}
-                        </li>
-                    )}
                 </ol>
             </div>
         </ol>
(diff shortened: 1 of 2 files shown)


temrjan:
Hi @ValeraS, just a gentle reminder about this PR. Would appreciate a review when you have a moment, thanks!

ValeraS:
@temrjan Hi, this was fixed here #2647

PR #2639, opened 2026-04-01
fix(useList): replace push spread with direct iteration in flattenItems

4y4a:


src/components/useList/utils/flattenItems.test.ts
```diff
@@ -0,0 +1,206 @@
...
+    });
+
+    test('rootIds should only contain top-level item ids', () => {
+        const result = flattenItems({items: data});
+
+        expect(result.rootIds).toEqual(['0', '1', '2']);
+        expect(result.rootIds).toHaveLength(data.length);
+    });
+
+    test('should handle a group with large number of children without RangeError', () => {
```
4y4a, line 166: 300k children that reproduces the `RangeError` on the original implementation.  V8 limits the number of         
  arguments, so any group with >200k children throws `RangeError: Maximum call stack size exceeded`.   But Yandex Browwer has a lower limit, I catch it at 120k elements.

src/components/useList/utils/flattenItems.ts
```diff
@@ -31,32 +27,27 @@ export function flattenItems<T>({
...
         }
 
-        order.push(id);
+        visibleFlattenIds.push(id);
 
         if (isTreeItemGuard(item) && item.children) {
             // don't include collapsed groups
-            if (!(id in expandedById && !expandedById[id])) {
-                order.push(
-                    ...item.children.reduce<string[]>(
```
4y4a, line 40: `reduce` created a new `[]` accumulator for each group, then spread its contents into the parent
 array, copying all ids twice per level.

4y4a, line 40: `order.push(...acc)` spreads the entire children array as function arguments. The reason for the crash on a huge  amount of elements

PR #2640, opened 2026-04-01
feat(Drawer): allow DrawerItem to be scrollable

igor0226:


src/components/Drawer/components/Drawer.scss
```diff
@@ -25,6 +25,12 @@ $drawer-resizer-width: 8px;
         will-change: transform;
     }
 
+    &__item-content {
+        overflow: auto;
```
amje: That's will be a breaking change. It could break user layouts. What it fixes?

igor0226: It is not necessary though
Users can add overflow properties themselves. However, I though that it would be useful to fix this behavior by default

amje: Its already `overflow: auto` on the `g-drawer` class. It's too much overflows

igor0226: What about the following proposal:
If we consider that my change is breaking, then I can show the similar case in a separate story. There I will place overflow:auto rule for the story child component without changing the basic Drawer's styles and layout (and pass a large component as a child one)

PR #2651, opened 2026-04-20
feat(useFocusWithin): with focus on parent document

bratva:
**Problem**
When useFocusWithin is used inside an iframe and popups are portalled into window.parent.document, clicking inside such a portal fires blur on the iframe element with relatedTarget === null (cross-document focus transitions always produce null). The hook treated this as "focus left the component" and fired onBlurWithin immediately, causing dropdowns and popups to close on the first interaction.

**Solution**
When relatedTarget === null and window !== window.parent, the blur is deferred via setTimeout(0). A focusin listener on window.parent then fires before the timeout: it cancels the deferred blur and fires it immediately with the correct synthetic target. If the parent is cross-origin, the error is swallowed and the deferred timeout fires as a fallback.

**Where**
1. Daterange if popup render on parent document
2. Select if popup render on parent document

Issue: https://github.com/gravity-ui/uikit/issues/2653

### DIFF
--- a/src/hooks/useFocusWithin/useFocusWithin.test.tsx
+++ b/src/hooks/useFocusWithin/useFocusWithin.test.tsx
@@ -1,3 +1,5 @@
+import * as React from 'react';
+
 import userEvent from '@testing-library/user-event';
 import ReactDom from 'react-dom';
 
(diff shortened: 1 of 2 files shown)


ValeraS:
@bratva Thank you for your PR. Issue was fixed in this [release](https://github.com/gravity-ui/uikit/releases/tag/v7.39.0)

Issue #2178, opened 2025-03-24
Sheet of components a strip appears under the swipe


NasgulNexus:
### Describe the bug

I have attached a video, you can play it in the storybook. If you open it in the store https://preview.gravity-ui.com/uikit/?path=/story/components-overlays-sheet--with-menu-showcase and look from the browser, you can see the strip, if you make the mobile view through the browser, the strip disappears, but if you zoom in, it appears. It seems that it should either always be shown or it should not be there

### Reproduction

https://preview.gravity-ui.com/uikit/?path=/story/components-overlays-sheet--with-menu-showcase

### Environment

Yandex browser 
Brave

### Validations

- [x] Check that there isn't [already an issue](https://github.com/gravity-ui/uikit/issues) that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

amje:
This line is a gap of veil between "shee-top" and "scroll-container" elements.
Possible fix: use single background on the parent element

amje:
> Possible fix: use single background on the parent element

@mournfulCoroner Take a look please

PR #2658, opened 2026-04-24
fix(Sheet): remove veil seam

ksvfs:
#2178

## Summary by Sourcery

Enhancements:
- Move the sheet background color and top border radius styling from the header and body sections to the main content element so the sheet renders as a single continuous surface.

### DIFF
--- a/src/components/Sheet/Sheet.scss
+++ b/src/components/Sheet/Sheet.scss
@@ -35,6 +35,12 @@ $block: '.#{variables.$ns}sheet';
         width: 100%;
         max-height: 100%;
         will-change: transform;
+        background-color: var(--g-sheet-background-color, var(--g-color-base-float));
+
+        &:has(> #{$block}__sheet-top) {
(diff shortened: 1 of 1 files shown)


korvin89 (changes requested):
What problem does this solve? The current type description violates BEM notation — you shouldn’t nest the styles of one block inside the styles of another.

ksvfs:
> What problem does this solve? The current type description violates BEM notation — you shouldn’t nest the styles of one block inside the styles of another.

This solves the problem described in issue #2178 that I mentioned in the PR description. The veil was visible through the seam between `sheet-top` and `sheet-scroll-container`.

I did not nest the styles of one block inside the styles of another. The name of the `sheet-top` element is only used as part of a selector that still targets the `sheet` element. Is that still considered a violation? There already are usages of `:has()` in the codebase.

mournfulCoroner (approved):

korvin89 (approved):

PR #2673, opened 2026-05-13
fix(Drawer): fixed auto focus in drawer items

ReFFaT:
I fixed the issue where the FloatingFocusManager used scrollIntoView() when autofocusing on elements in the Drawer, causing the browser to interrupt smooth animations and use its own focus-setting method, resulting in sudden jumps.

## Summary by Sourcery

Adjust Drawer initial focus handling to avoid disruptive scroll-based autofocus and add a story to showcase the behavior.

New Features:
- Add an InitialAutoFocus Drawer story demonstrating automatic focus on an input when the Drawer opens.

Bug Fixes:
- Ensure Drawer initial focus is applied without using scroll-into-view behavior that interrupts animations.

Enhancements:
- Extract a useInitialFocus hook to manage autofocus logic via refs or tabbable indices instead of delegating to the floating focus manager.


src/components/Drawer/hooks/useInitialFocus.ts
```diff
@@ -0,0 +1,35 @@
...
+
+type InitialFocus = number | React.RefObject<HTMLElement>;
+
+interface UseInitialFocusParams {
+    open: boolean | undefined;
+    initialFocus: InitialFocus | undefined;
+    floatingRef: React.RefObject<HTMLElement | null>;
+}
+
+export function useInitialFocus({open, initialFocus, floatingRef}: UseInitialFocusParams) {
```
igor0226, line 13: This function looks really neat. However, I wonder why we need this utility

In the Popup component I can see a simple passing of the `initialFocus` prop

https://github.com/gravity-ui/uikit/blob/c240b056157b4c37616a3d330ac391d79f1efbe5/src/components/Popup/Popup.tsx#L328

igor0226 (approved):

PR #2684, opened 2026-05-20
feat(Skeleton): adjustable skeleton

qradle-yndx:
## Summary by Sourcery

Extend the Skeleton component with adjustable visual variants and child-wrapping behavior, and document and demonstrate these new capabilities in stories and READMEs.

New Features:
- Add a `variant` prop to Skeleton to support square, rounded, circle, and text-specific skeleton shapes with a new text-line style.
- Allow Skeleton to wrap a single child element and act as an overlay that automatically adopts the child's dimensions.

Enhancements:
- Refine Skeleton styling to support variant-specific border radii and text-line layout, and introduce a children overlay mode.
- Expand Skeleton Storybook stories to showcase text variants and child-wrapping use cases for Avatar, Button, and Label components.

Documentation:
- Update English and Russian Skeleton READMEs to describe the new `variant` and `children` props with usage guidance and code examples.


src/components/Skeleton/Skeleton.tsx
```diff
@@ -8,26 +8,50 @@ import './Skeleton.scss';
...
+    style,
+    qa,
+    animation = 'gradient',
+    variant = 'rounded',
+    children,
+}: SkeletonProps) {
+    if (children !== undefined) {
+        return (
+            <div
+                className={b({children: true, animation, variant}, className)}
```
ogonkov: Looks like we can just pass this modifier rather than duplicate whole component

```
<div className={b({children: children !== undefined, animation, variant}, className)} style={style} data-qa={qa}>
  {children}
</div>
```

src/components/Skeleton/Skeleton.tsx
```diff
@@ -8,26 +8,50 @@ import './Skeleton.scss';
 const b = block('skeleton');
 
 export interface SkeletonProps
-    extends Pick<React.HTMLAttributes<HTMLDivElement>, 'className' | 'style'>,
+    extends Pick<React.HTMLAttributes<HTMLElement>, 'className' | 'style'>,
```
ogonkov: Why element type was changed?

src/components/Skeleton/Skeleton.tsx
```diff
@@ -1,4 +1,4 @@
-import type * as React from 'react';
```
ogonkov, line 1: This change was intentional?

src/components/Skeleton/Skeleton.scss
```diff
@@ -11,11 +11,43 @@ $block: '.#{variables.$ns}skeleton';
...
     background-color: var(--g-color-base-generic);
 
+    &_children {
+        display: inline-flex;
+        width: unset;
+        height: unset;
+
+        > * {
+            /* stylelint-disable-next-line declaration-no-important */
+            opacity: 0 !important;
```
ogonkov: Why `!important` is required here? May be we can use visibility: hidden?

src/components/Skeleton/Skeleton.scss
```diff
@@ -11,11 +11,43 @@ $block: '.#{variables.$ns}skeleton';
...
+        }
+    }
+
+    &_variant {
+        &_rounded {
+            border-radius: 5px;
```
ogonkov: We might want to add css-api for border-radius

src/components/Skeleton/__stories__/Skeleton.stories.tsx
```diff
@@ -29,4 +43,144 @@ AnimationsNone.storyName = 'Animations/None';
...
+                        </Flex>
+                    </ShowcaseItem>
+                ))}
+            </Showcase>
+            <Showcase title="Button">
+                {BUTTON_SIZES.map((size) => (
+                    <ShowcaseItem key={size} title={size}>
+                        <Flex direction="column" gap={1} alignItems="flex-start">
+                            <Skeleton>
+                                <Button size={size}>Real</Button>
```
amje: Button's border radius is different for different sizes. Also children will be rendered to the dom and we cannot constrain what user can render. I think it's not a good idea. Instead it will be more practical to have predefined set of sizes for Skeleton which control height and border roundness instead of passing children.

amje (changes requested):

src/components/Skeleton/__stories__/Skeleton.stories.tsx
qradle-yndx: Ok. I'll give it a try and see what happens

src/components/Skeleton/Skeleton.tsx
qradle-yndx: As @amje wrote. I'll try to make it without children rendering

src/components/Skeleton/Skeleton.tsx
qradle-yndx: You right. That was experiments with the children. I'll return it back

src/components/Skeleton/Skeleton.scss
qradle-yndx: visibility: hidden doesn't work for components like Label with the `:after` blocks.
Anyway i'll try a version without children

src/components/Skeleton/__stories__/Skeleton.stories.tsx
```diff
@@ -29,4 +43,144 @@ AnimationsNone.storyName = 'Animations/None';
...
+                            <Button size={size}>Real</Button>
+                        </Flex>
+                    </ShowcaseItem>
+                ))}
+            </Showcase>
+            <Showcase title="Label">
+                {LABEL_SIZES.map((size) => (
+                    <ShowcaseItem key={size} title={size}>
+                        <Flex direction="column" gap={1} alignItems="flex-start">
+                            <Skeleton>
```
Ruminat: @amje what do you think about this idea:
- what if we add a `loaded` prop to `Skeleton`
- when it's set to `true`, we stop showing the skeleton and just show its children

why? To avoid copy pasting and being able to do this:

```tsx
function ComponentWithSkeleton() {
  return (
    <Skeleton loaded={hasLoaded(content)}>
      <Label size="l" theme="danger">{hasLoaded(content) ? content : "loading"}</Label>
    </Skeleton>
  );
}
```

instead of this:

```tsx
function ComponentWithSkeleton() {
  return hasLoaded(content) ? (
    <Label size="l" theme="danger">{content}</Label>
  ) : (
    <Skeleton>
      <Label size="l" theme="danger">loading</Label>
    </Skeleton>
  );
}
```

Ruminat: Ah, saw your comment up there

Image

Well, if refusing from using `children`, it it's not a problem anymore I guess

Ruminat: I guess it might make sense to create a common heights' map as first suggested in the ticket
Like this one:

```typescript
const text = {
  'body-1': 'var(--g-text-body-1-line-height)',
  'body-2': 'var(--g-text-body-2-line-height)',
  'body-3': 'var(--g-text-body-3-line-height)',
  'body-short': 'var(--g-text-body-short-line-height)',
  'caption-1': 'var(--g-text-caption-1-line-height)',
  'caption-2': 'var(--g-text-caption-2-line-height)',
  'header-1': 'var(--g-text-header-1-line-height)',
  'header-2': 'var(--g-text-header-2-line-height)',
  'subheader-1': 'var(--g-text-subheader-1-line-height)',
  'subheader-2': 'var(--g-text-subheader-2-line-height)',
  'subheader-3': 'var(--g-text-subheader-3-line-height)',
  'display-1': 'var(--g-text-display-1-line-height)',
  'display-2': 'var(--g-text-display-2-line-height)',
  'display-3': 'var(--g-text-display-3-line-height)',
  'display-4': 'var(--g-text-display-4-line-height)',
  'code-1': 'var(--g-text-code-1-line-height)',
  'code-2': 'var(--g-text-code-2-line-height)',
  'code-3': 'var(--g-text-code-3-line-height)',
  'code-inline-1': 'var(--g-text-code-inline-1-line-height)',
  'code-inline-2': 'var(--g-text-code-inline-2-line-height)',
  'code-inline-3': 'var(--g-text-code-inline-3-line-height)',
} satisfies Record<(typeof TEXT_VARIANTS)[number], string>;

const label = {
  xxs: 16,
  xs: 20,
  s: 24,
  m: 28,
} satisfies Record<Required<LabelProps>['size'], number>;

const button = {
  xs: 20,
  s: 24,
  m: 28,
  l: 36,
  xl: 44,
} satisfies Record<ButtonSize, number>;

const input = {
  s: button.s,
  m: button.m,
  l: button.l,
  xl: button.xl,
} satisfies Record<InputControlSize, number>;

export const HeightPreset = {
  text,
  label,
  button,
  input,
} as const;
```

src/components/Skeleton/__stories__/Skeleton.stories.tsx
qradle-yndx: I made the update adding the size prop

Ruminat (approved):

src/components/Skeleton/__stories__/Skeleton.stories.tsx
qradle-yndx: @amje please take a look at the last changes

amje (approved):

src/components/Skeleton/__stories__/Skeleton.stories.tsx
amje: @Ruminat Let's discuss height maps it in a separate issue/PR

src/components/Skeleton/Skeleton.scss
```diff
@@ -6,15 +6,68 @@ $block: '.#{variables.$ns}skeleton';
...
+
+        &_xl {
+            --_--border-radius: var(--g-border-radius-xl);
+            --_--size: #{variables.$xl-height};
+        }
+    }
+
+    // Apply height from the size token. Only has effect when a specific size
+    // modifier is present (i.e. --_--size is defined by one of the &_size_* rules above).
+    height: var(--_--size);
```
ogonkov: Move this next to the other block styles above for consistency.

Also, without `size` set we probably want `auto` as default value here

ogonkov (approved):

Issue #2270, opened 2025-05-20
[TextArea]: synchronicity of design and component (errorPlacement)


nikitalameko:
### Description

The design on the errorPlacement props is there, but the component does not allow for this 
Old issue: https://github.com/gravity-ui/uikit/pull/994
Design: https://gravity-ui.com/design/guides/text-area

<img width="893" alt="Image" src="https://github.com/user-attachments/assets/c5d42c71-e582-48bb-ba1f-7f75e2ea26ec" />

### Suggested solution

Ability to customize where the error is displayed in TextArea
errorPlacement: "inside" | "outside"

### Validations

- [x] Check that there isn't [already an issue](https://github.com/gravity-ui/uikit/issues) that request the same feature to avoid creating a duplicate.

KrotovPetr:
@nikitalameko I took it to work, but forgot create PR, sorry

PR #2689, opened 2026-05-22
feat(TextArea): enable errorplacement prop ISSUE-2270

KrotovPetr:
#2270

## Summary by Sourcery

Add support for placing TextArea validation errors inside the control via an icon tooltip and update styles, tests, and docs accordingly.

New Features:
- Introduce the errorPlacement prop for TextArea to render validation errors either outside or inside the control via an error icon tooltip.

Documentation:
- Document the new TextArea errorPlacement prop, its default behavior, and add examples for inside error placement.

Tests:
- Extend TextArea unit and visual tests to cover the new errorPlacement options, including inside placement with and without a clear button.


src/components/controls/TextArea/TextArea.scss
```diff
@@ -184,6 +227,20 @@ $block: '.#{variables.$ns}text-area';
...
+
+    &_has-clear#{$block}_has-error-icon {
+        #{$block}__clear {
+            inset-inline-end: calc(var(--_--clear-offset) + 20px);
+        }
+
+        &#{$block}_has-scrollbar #{$block}__clear {
+            inset-inline-end: calc(var(--g-scrollbar-width) + 20px);
+        }
     }
```
korvin89, line 292: **[minor] On `l`/`xl` the error icon overlaps the clear button (~4px hit-box overlap).**

Here the clear button is pushed to a hardcoded `clear-offset + 20px` (≈22px). But the icon's far edge is `error-icon-offset-inline + 16px`: for `s`/`m` that's `6+16=22px` (flush, ok), while for `l`/`xl` it's `10+16=26px`, i.e. it slides ~4px under the clear button. The `+20px` only matches the `s`/`m` offset. This is invisible in review because the showcase and both visual tests use size `m`.

I'd tie the offset to `error-icon-offset-inline` so the clear button lands exactly at the icon's far edge for every size (`s`/`m` unchanged: `6+16=22`; `l`/`xl`: `10+16=26`):

```suggestion
    &_has-clear#{$block}_has-error-icon {
        #{$block}__clear {
            inset-inline-end: calc(var(--_--clear-offset) + var(--_--error-icon-offset-inline) + 14px);
        }

        &#{$block}_has-scrollbar #{$block}__clear {
            inset-inline-end: calc(var(--g-scrollbar-width) + var(--_--error-icon-offset-inline) + 14px);
        }
    }
```
Would also be good to add a visual scenario for `l`/`xl` + `hasClear` + `errorPlacement="inside"` so this case is screenshot-covered.

src/components/controls/TextArea/TextArea.scss
```diff
@@ -92,16 +92,35 @@ $block: '.#{variables.$ns}text-area';
         }
     }
 
+    &__error-icon-wrap {
+        position: absolute;
+        inset-inline-end: var(--_--error-icon-offset-inline);
+        inset-block-start: var(--_--error-icon-offset-block);
+    }
```
korvin89, line 100: **[nit / robustness] Equal-specificity `position` conflict relies on stylesheet source order.**

`className={b('error-icon-wrap')}` (TextArea.tsx:188) lands on the legacy Popover's root `<div>`, so the same element carries both `.g-popover-legacy { position: relative }` and `.g-text-area__error-icon-wrap { position: absolute }` — both `(0,1,0)` selectors. `absolute` only wins because TextArea's styles are emitted after the Popover's; a change in import/bundler order could silently break the positioning. `TextInput` avoids this by not passing `className` to its Popover.

Simplest fix — raise the rule's specificity so it wins deterministically:

```suggestion
    &__error-icon-wrap#{$block}__error-icon-wrap {
        position: absolute;
        inset-inline-end: var(--_--error-icon-offset-inline);
        inset-block-start: var(--_--error-icon-offset-block);
    }
```
A cleaner alternative — don't put the class on the Popover at all; wrap it in a dedicated `<span className={b('error-icon-wrap')}>` and leave the Popover with its `position: relative`.

korvin89 (commented):
5 follow-up notes on `errorPlacement="inside"` from the review (CI/tests are green and rendering is correct — this is polish). I'd fix #1–#2 before merge; #3–#5 are optional. All a11y findings are inherited from `TextInput` and split out into a separate follow-up.

src/components/controls/TextArea/TextArea.scss
```diff
@@ -110,10 +129,18 @@ $block: '.#{variables.$ns}text-area';
                 @include control-mixins.input-control(m);
             }
 
-            &#{$block}_has-clear #{$block}__control {
+            &#{$block}_has-clear #{$block}__control,
+            &#{$block}_has-error-icon #{$block}__control {
                 padding-inline-end: 26px;
```
korvin89: **[minor] `view="clear"` + `errorPlacement="inside"` keeps the control's `26/36px` end padding.**

These `_has-error-icon __control { padding-inline-end }` rules have specificity `(0,3,0)`, while the clear-view reset `.g-text-area_view_clear .g-text-area__control { padding-inline: 0 }` is `(0,2,0)`, so it loses: in clear view the text doesn't sit flush as expected.

I'd scope the per-size padding under `view_normal` so it doesn't fight the clear-view reset, e.g.:
```scss
&#{$block}_view_normal#{$block}_has-clear #{$block}__control,
&#{$block}_view_normal#{$block}_has-error-icon #{$block}__control {
    padding-inline-end: 26px;
}
```
(same for `46px`, and for `l`/`xl`). Alternatively, zero out `--_--error-icon-offset-inline` and the padding inside the `&_view_clear` block. And add a `clear`+`inside` screenshot. Affects all sizes: `s` (≈113), `m` (132), `l` (152), `xl` (172).

src/components/controls/TextArea/TextArea.scss
```diff
@@ -184,6 +227,20 @@ $block: '.#{variables.$ns}text-area';
...
         }
+
+        #{$block}__error-icon-wrap {
+            inset-inline-end: var(--g-scrollbar-width);
+        }
+    }
+
+    &_has-clear#{$block}_has-error-icon {
+        #{$block}__clear {
+            inset-inline-end: calc(var(--_--clear-offset) + 20px);
```
korvin89: **[nit] On `l`/`xl` the clear button loses its `+1px` horizontal nudge when the icon appears.**

The base clear button on `l`/`xl` sits at `calc(var(--_--clear-offset) + 1px)`, but this override (and its scrollbar variant on line 242) uses `+ 20px` without the `+1`, causing a 1px horizontal jitter when the icon shows up. Vertical is fine (`inset-block-start` is untouched).

If the offset-inline-based formula from the overlap comment is adopted, positioning becomes consistent across sizes and this point essentially goes away. Otherwise, carry the per-size `+1px` via a dedicated variable. Low priority.

src/components/controls/TextArea/TextArea.scss
```diff
@@ -92,16 +92,35 @@ $block: '.#{variables.$ns}text-area';
         }
     }
 
+    &__error-icon-wrap {
+        position: absolute;
+        inset-inline-end: var(--_--error-icon-offset-inline);
+        inset-block-start: var(--_--error-icon-offset-block);
```
korvin89: **[nit] The icon's vertical center is ~1px off from the clear button's center.**

The clear button is centered exactly on the control's vertical center at every size, but because of the hardcoded `--_--error-icon-offset-block` (5/7/11/15px) the icon's center is 1px lower on `m/l/xl` and 1px higher on `s`. Only noticeable when both the clear button and the inside icon are shown.

Cleaner to center generically instead of the per-size hardcode (then the `--_--error-icon-offset-block` variables can be dropped):
```scss
&__error-icon-wrap {
    /* ... */
    inset-block-start: 50%;
    transform: translateY(-50%);
}
```
This mirrors the auto-centering `TextInput` already uses (flex). NB: a blanket `-1px` on the offsets won't work — `s` would get worse.

KrotovPetr:
telegram-cloud-photo-size-2-5274043269547498574-y
telegram-cloud-photo-size-2-5274043269547498575-y

korvin89 (commented):
🤖 AI-generated review — align the inside error icon (errorPlacement="inside") with TextInput.

After the fix the measurements match on both axes for every size (centerY 12/14/18/22, right-gap 5/5/9/13).

src/components/controls/TextArea/TextArea.scss
```diff
@@ -92,16 +92,43 @@ $block: '.#{variables.$ns}text-area';
         }
     }
 
+    &__error-icon-wrap {
+        position: absolute;
+        display: inline-flex;
+        inset-inline-end: 0;
```
korvin89: 🤖 _AI-generated suggestion_

The wrap is pinned to the outer edge (outside the 1px border), so the icon sits 1px closer to the right edge than in TextInput. Offsetting by the border width aligns them pixel-perfectly.

```suggestion
        inset-inline-end: var(--g-text-area-border-width, var(--_--border-width));
```

src/components/controls/TextArea/TextArea.scss
```diff
@@ -92,16 +92,43 @@ $block: '.#{variables.$ns}text-area';
...
 
+            &#{$block}_view_normal#{$block}_has-error-icon #{$block}__control {
+                padding-inline-end: 22px;
+            }
+
+            &#{$block}_view_normal#{$block}_has-clear#{$block}_has-error-icon #{$block}__control {
+                padding-inline-end: 42px;
+            }
+
+            --_--error-icon-padding-block: 5px;
```
korvin89: 🤖 _AI-generated suggestion_

`padding-block` is 1px short of true centering, so the glyph sits 1px above center (size s). +1px matches the glyph center to the field center, like TextInput.

```suggestion
            --_--error-icon-padding-block: 6px;
```

src/components/controls/TextArea/TextArea.scss
```diff
@@ -110,10 +137,22 @@ $block: '.#{variables.$ns}text-area';
...
+
+            &#{$block}_view_normal#{$block}_has-error-icon #{$block}__control {
                 padding-inline-end: 26px;
             }
 
+            &#{$block}_view_normal#{$block}_has-clear#{$block}_has-error-icon #{$block}__control {
+                padding-inline-end: 48px;
+            }
+
+            --_--error-icon-padding-block: 5px;
```
korvin89: 🤖 _AI-generated suggestion_

Same as size s: the error icon's vertical centering is 1px short (size m).

```suggestion
            --_--error-icon-padding-block: 6px;
```

src/components/controls/TextArea/TextArea.scss
```diff
@@ -122,10 +161,22 @@ $block: '.#{variables.$ns}text-area';
...
 
+            &#{$block}_view_normal#{$block}_has-error-icon #{$block}__control {
+                padding-inline-end: 30px;
+            }
+
+            &#{$block}_view_normal#{$block}_has-clear#{$block}_has-error-icon #{$block}__control {
+                padding-inline-end: 62px;
+            }
+
+            --_--error-icon-padding-block: 9px;
```
korvin89: 🤖 _AI-generated suggestion_

The error icon's vertical centering is 1px short (size l).

```suggestion
            --_--error-icon-padding-block: 10px;
```

src/components/controls/TextArea/TextArea.scss
```diff
@@ -134,10 +185,22 @@ $block: '.#{variables.$ns}text-area';
...
 
+            &#{$block}_view_normal#{$block}_has-error-icon #{$block}__control {
+                padding-inline-end: 38px;
+            }
+
+            &#{$block}_view_normal#{$block}_has-clear#{$block}_has-error-icon #{$block}__control {
+                padding-inline-end: 66px;
+            }
+
+            --_--error-icon-padding-block: 13px;
```
korvin89: 🤖 _AI-generated suggestion_

The error icon's vertical centering is 1px short (size xl).

```suggestion
            --_--error-icon-padding-block: 14px;
```

src/components/controls/TextArea/TextArea.scss
```diff
@@ -184,6 +247,46 @@ $block: '.#{variables.$ns}text-area';
         #{$block}__clear {
             inset-inline-end: var(--g-scrollbar-width);
         }
+
+        #{$block}__error-icon-wrap {
+            inset-inline-end: var(--g-scrollbar-width);
```
korvin89: 🤖 _AI-generated suggestion_

With a scrollbar the icon must also be offset by the border width so the right-side gap matches TextInput.

```suggestion
            inset-inline-end: calc(
                var(--g-scrollbar-width) + var(--g-text-area-border-width, var(--_--border-width))
            );
```

PR #2690, opened 2026-05-22
feat(Ellipsis): add component

PahaN47:
## Summary by Sourcery

Add a reusable Ellipsis text component with configurable truncation behavior and RTL-safe styling, including stories and theme support for flow direction.

New Features:
- Introduce an Ellipsis React component that supports start, center, and end truncation of long strings with optional offsets and separators.
- Add a center-ellipsis hook that dynamically measures available space and computes a middle truncation responsive to container resize events.
- Provide Storybook stories demonstrating Ellipsis usage with adjustable width, different positions, and separator-based truncation.

Enhancements:
- Extend common theme flow-direction variables to include an opposite-flow variable used for RTL-safe text ellipsis styling.


korvin89:
@PahaN47  Hi! We looked (me and @amje ) at the stories and the general approach in the code. Before review, we'd like to discuss the following questions:

1. Why is the `separator` prop needed? It's unclear from the stories what it does.
2. In the default story, let's replace the range slider component with a resizable div (like [here](https://preview.gravity-ui.com/components/?path=/story/components-adaptivetabs--default)).
3. We suggest replacing `offsetStart` and `offsetEnd` with a single `offset` that can accept negative values and will only be responsible for the position of the ellipsis.
4. For setting the minimum number of elements from the start or end, we suggest introducing `minCharsStart` and `minCharsEnd` props.
5. Use a shared hook for working with `ResizeObserver`. Your optimization can optionally be added to it, but it should be done via `WeakMap` rather than `Map`.
6. Consider dropping `--g-flow-opposite`. You named the variable using the public convention, which implies it can be mutated by the user — but the user shouldn't be able to control this value from their side.
7. When using native `overflow: ellipsis`, a gap appears around the ellipsis (on the left or right). We suggest keeping this approach only when `position=end` is used and the `minCharsEnd` prop is absent — this should cover most cases. For the rest, use JS-based calculations.

PahaN47:
@korvin89 
> 1. Why is the `separator` prop needed? It's unclear from the stories what it does.

By default the ellipsis is offset by a certain number of characters, but if we wanted to use it for, say, file paths and wanted to always keep the actual file name with the extension visible, we would not know the exact number of characters we would need to use for this. So when there is a separator, we split the text into parts delimited by it, and then apply offset to these parts, not the characters. This way we could specify to separate the text with `'/'` and set `offsetEnd={1}` so that the actual file name is always displayed.

> 3. We suggest replacing `offsetStart` and `offsetEnd` with a single `offset` that can accept negative values and will only be responsible for the position of the ellipsis.

Then we would not be able to specify both at the same time. For instance, if we always wanted to keep the first and last 3 characters of the text.

> 4. For setting the minimum number of elements from the start or end, we suggest introducing `minCharsStart` and `minCharsEnd` props.

Is this not the same as offset? 

> 6. Consider dropping `--g-flow-opposite`. You named the variable using the public convention, which implies it can be mutated by the user — but the user shouldn't be able to control this value from their side.

The variable is needed for the `start` ellipsis. In order to make it work without involving JS, we need to know the opposite direction to the one that is used by default. Any suggestions on how I could rename it maybe?

> 7. When using native `overflow: ellipsis`, a gap appears around the ellipsis (on the left or right). We suggest keeping this approach only when `position=end` is used and the `minCharsEnd` prop is absent — this should cover most cases. For the rest, use JS-based calculations.

I was trying to avoid JS as much as possible, as the calculations are quite heavy. The only case that could not be implemented without it is the `center` position, so this is the only one that uses JS. I am very hesitant to expand JS usage as compared to accepting the small gap. One day, JS could even be dropped entirely, if [`text-overflow: <string>`](https://caniuse.com/mdn-css_properties_text-overflow_string) gets widely supported

src/components/Ellipsis/hooks.ts
```diff
@@ -0,0 +1,146 @@
...
+    if (!container || !measure) {
+        return;
+    }
+
+    const availableSpace = getElementWidth(container);
+    if (availableSpace <= 0) {
+        return;
+    }
+
+    measure.textContent ||= text;
```
korvin89, line 78: 🤖 **AI generated**

🔴 **blocker**

`measure.textContent ||= text;` only fills the probe element while it is empty, so once it holds a value it is never refreshed. When the `text` prop changes on a mounted instance, the "does the whole text fit?" check on line 79 measures the *previous* text.

If the old text fitted, the check passes and the new, longer text is rendered in full and hard-clipped by `overflow: hidden` with no ellipsis — and it does not recover, because every later resize short-circuits on the same stale measurement. Reproduced in Chromium with `position="center"` in a 200px box: changing `short.txt` to a 386px-wide string leaves the probe at `short.txt`.

This looks like a regression from 9363713f — before that commit the assignment was unconditional:

```
8f83443c  measure.textContent = text;
86dd0e8b  measure.textContent = text;
af6e672d  measure.textContent = text;
9363713f  measure.textContent ||= text;   ← here
```

Reverting to `measure.textContent = text;` should fix it; the write is cheap, and `calculateCollapsedText` already does exactly that on line 59.

src/components/Ellipsis/Ellipsis.tsx
```diff
@@ -0,0 +1,193 @@
...
+        [text, endOffset, startOffset, isCenterPosition],
+    );
+
+    return (
+        <span
+            className={b(null, className)}
+            style={style}
+            ref={ref}
+            onCopy={handleCopy}
+            aria-label={text}
```
korvin89, line 175: 🤖 **AI generated**

🔴 **blocker**

`aria-label={text}` sits on a `<span>` with no role (implicit `generic`), while every descendant is marked `aria-hidden` (lines 32, 52, 57, 62). ARIA prohibits `aria-label` on `generic`, so the component may end up with no accessible name at all — and there is nothing else left to announce, since all the text is hidden.

I ran `axe-core@4.10.3` (the version in this repo's `node_modules`) against the rendered markup and it reports `aria-prohibited-attr` with impact **serious**. `.storybook/test-runner.ts:26` calls `checkA11y` on every story, so I would expect the Storybook test run to go red on this.

Two fixes both come back clean under axe:

- add `role="img"` alongside the `aria-label`, or
- drop the `aria-label` and render a visually-hidden `<span>` carrying the full text next to the `aria-hidden` visual part.

Worth noting that for `position="start"` / `"end"` the complete text is already in the DOM, so the `aria-hidden` + `aria-label` pair is a pure loss there — hiding it and re-labelling only removes information.

src/components/Ellipsis/Ellipsis.tsx
```diff
@@ -0,0 +1,193 @@
...
+const getPartsWithoutSeparator = (
+    text: string,
+    offsetStart: number,
+    offsetEnd: number,
+): TextParts => {
+    const textLength = text.length;
+    const offsetEndLocal = -offsetEnd || textLength;
+
+    return [
+        text.slice(0, offsetStart),
```
korvin89, line 86: 🤖 **AI generated**

🔴 **blocker**

Neither `getPartsWithoutSeparator` nor `getPartsWithSeparator` checks that the two offsets do not cross, so overlapping offsets silently duplicate characters. Reproduced against the component itself:

```tsx
<Ellipsis offsetStart={4} offsetEnd={7}>a.tar.gz</Ellipsis>
// rendered:   "a.ta.tar.gz"
// aria-label: "a.tar.gz"
```

Same class of bug on the separator path — `path/to/file` with `separator="/"` and offsets 3/3 renders `path/toto/file`.

This is reachable straight from the example in the README (`offsetEnd={7}` for a `.tar.gz` extension) as soon as the filename is short, and the rendered text then disagrees with `aria-label`.

Would it make sense to clamp before slicing — `endOffsetStart = Math.max(endOffsetStart, startOffsetEnd)` on the separator path and the equivalent on the character path? Negative offsets are unguarded too, if that is worth covering while you are there.

src/components/Ellipsis/Ellipsis.tsx
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    return [
+        text.slice(0, startOffsetEnd),
+        text.slice(startOffsetEnd, endOffsetStart),
+        text.slice(endOffsetStart),
+    ];
+};
+
+export const Ellipsis = React.forwardRef<HTMLSpanElement, EllipsisProps>(function Ellipsis(
```
korvin89, line 126: 🤖 **AI generated**

🔴 **blocker**

The component bypasses the library's standard prop plumbing:

- no `useDefaultProps('Ellipsis', rawProps)`, and no entry in `ComponentDefaultPropsMap` (`src/components/theme/PrivateDefaultPropsProvider.tsx`) — 63 components go through this;
- `EllipsisProps` does not extend `AriaLabelingProps` (line 67);
- no `...restProps` + `filterDOMProps(restProps, {labelable: true})` spread.

`src/components/Divider/Divider.tsx:20-46` is a compact reference for all three.

The practical effect on the public contract is that a consumer cannot pass `id`, `title`, `data-*`, `aria-describedby`, or any event handler, and cannot override the hard-coded `aria-label` — which is also the escape hatch that would let them work around the accessibility problem flagged separately.

styles/themes/common/_index.scss
```diff
@@ -18,18 +18,21 @@
     @at-root body#{&} {
         // default document direction is ltr
         --g-flow-direction: 1;
+        --g-flow-opposite: rtl;
```
korvin89, line 21: 🤖 **AI generated**

🔴 **blocker**

Following up on point 6 from the earlier thread — I do not think this variable needs to exist at all.

Beyond the naming concern (a `--g-*` name reads as something a consumer may set, and this is not), it is also subtly incorrect: it is defined only on `.g-root` / `body.g-root`, while `direction` is an inherited property that legitimately changes mid-tree. An element carrying its own `dir="ltr"` inside `.g-root[dir="rtl"]` still resolves `--g-flow-opposite: ltr` — the same direction it already has — and `position="start"` silently degrades into end-truncation.

The repo already has a pattern for direction-specific rules that follows the actual `dir` in the tree rather than the root (`Progress.scss:58`, `SliderTooltip.scss:7`). Applied here:

```scss
&_start {
    #{$ellipsis}-content {

        // Flip the writing direction so the browser puts the ellipsis
        // at the inline start.
        direction: rtl;

        [dir='rtl'] & {
            direction: ltr;
        }
    }
}
```

That keeps the flip local to the component, fixes the nested-`dir` case, and lets `styles/themes/common/_index.scss` stay untouched.

src/components/index.ts
```diff
@@ -22,6 +22,7 @@ export * from './Disclosure';
 export * from './Divider';
 export * from './Drawer';
 export * from './DropdownMenu';
+export * from './Ellipsis';
```
korvin89, line 25: 🤖 **AI generated**

🔴 **blocker**

`@gravity-ui/icons` already exports `Ellipsis` (and `EllipsisVertical`) — `node_modules/@gravity-ui/icons/index.d.ts:321`. Since uikit and icons are usually installed together, the two cannot be imported side by side without aliasing one of them, and the name gives no hint which is which.

Nothing has shipped yet, so the rename is free now and breaking later. `EllipsisText` would sit naturally next to `Text` (which already owns the `ellipsis` / `ellipsisLines` props) and makes clear this is about text rather than the "…" glyph; `TruncatedText` would work too. The BEM block and the SCSS file would follow.

src/components/Ellipsis/Ellipsis.tsx
```diff
@@ -0,0 +1,193 @@
...
+    );
+
+    const handleCopy = React.useCallback(
+        (e: React.ClipboardEvent<HTMLSpanElement>) => {
+            if (!isCenterPosition) {
+                return;
+            }
+
+            const clipboardText = (window.getSelection()?.toString() || '').trim();
+            const currentText =
```
korvin89, line 158: 🤖 **AI generated**

🟠 **should-fix**

`handleCopy` builds `currentText = startOffset + ellipsisContentRef.current.textContent + endOffset`, but for `position="center"` the ref points at the span holding `visibleText`, and `calculateCollapsedText` already returns `startOffset + … + endOffset` (`hooks.ts:35,47`). So the offsets are counted twice and the check on line 161 can never match whenever either offset is non-empty:

```
offsetStart=4, offsetEnd=7
selection   : "a-very-long…name.tar.gz"
currentText : "a-vea-very-long…name.tar.gz.tar.gz"
```

The result is that the "copy the full text" behaviour silently does nothing in exactly the case the README advertises it for. Comparing against the DOM text as-is (or against `visibleText` directly) would fix it.

Side note: the ref exists here only to read back out of the DOM a value `useCenterEllipsis` already holds in state — lifting `visibleText` up would remove the ref along with the bug.

src/components/Ellipsis/Ellipsis.tsx
```diff
@@ -0,0 +1,193 @@
...
+        <React.Fragment>
+            {startOffset && <span aria-hidden>{startOffset}</span>}
+            <span
+                className={b('ellipsis', {
+                    [position]: true,
+                })}
+                aria-hidden
+            >
+                {/* The extra characters are needed to prevent text shuffling due to direction: rtl with position: 'start' */}
+                <span className={b('ellipsis-content')}>{`${FSI}${collapsibleText}${PDI}`}</span>
```
korvin89, line 60: 🤖 **AI generated**

🟠 **should-fix**

The U+2068 / U+2069 isolates go into a real text node for both `start` and `end`, while `handleCopy` returns early for non-center positions (line 153). So with the default `position="end"` the user copies invisible control characters along with the text:

```
render(<Ellipsis position="end">hello-world.txt</Ellipsis>)
// textContent === "⁨hello-world.txt⁩"
```

The PR's own test acknowledges this — `__tests__/Ellipsis.test.tsx:7` adds a `stripIsolation()` helper before asserting "keeps the full text in the DOM (for copying)", so that assertion passes precisely because it does not check what a real copy would produce.

The comment on line 59 says the isolates are only needed for the `direction: rtl` trick on `position="start"`, so wrapping only in that case would be the smallest fix. Alternatively `unicode-bidi: isolate` in CSS gives the same protection without putting anything into the text node.

src/components/Ellipsis/Ellipsis.tsx
```diff
@@ -0,0 +1,193 @@
...
+            {endOffset && <span aria-hidden>{endOffset}</span>}
+        </React.Fragment>
+    );
+});
+
+export interface EllipsisProps extends DOMProps, QAProps {
+    position?: EllipsisPosition;
+    offsetStart?: number;
+    offsetEnd?: number;
+    separator?: string | string[];
```
korvin89, line 71: 🤖 **AI generated**

🟠 **should-fix**

`separator?: string | string[]` promises more than the implementation delivers. `Array.from(separator)` on line 99 followed by a single-code-unit comparison on lines 109/113 means:

- `separator="::"` is treated as the character *set* `{':'}`, not as the token `"::"`;
- `separator={['->']}` never matches, so the offsets are silently ignored and the whole text becomes collapsible;
- a non-BMP separator (an emoji, for instance) never matches either.

Since this is the same prop we were discussing in the thread above, it would be good to settle the contract at the same time: either narrow the type to a single character and say so in the README, or normalise to string tokens and scan with `startsWith`.

korvin89 (commented):
🤖 **AI generated**

**Severity legend**

src/components/Ellipsis/Ellipsis.tsx
```diff
@@ -0,0 +1,193 @@
...
+    offsetStart: number,
+    offsetEnd: number,
+): TextParts => {
+    const textLength = text.length;
+    const offsetEndLocal = -offsetEnd || textLength;
+
+    return [
+        text.slice(0, offsetStart),
+        text.slice(offsetStart, offsetEndLocal),
+        text.slice(offsetEndLocal),
```
korvin89, line 88: 🤖 **AI generated**

🟠 **should-fix**

All the slicing works on UTF-16 code units (`Ellipsis.tsx:86-88`, `hooks.ts:28-31`, `hooks.ts:44-45`), so offsets and the binary-search split can cut surrogate pairs in half:

```
offsetEnd={2} on "a👍b"  →  ["", "a\ud83d", "\udc4db"]
```

which renders two lone surrogates in separate spans. For a component whose entire job is cutting strings, `Array.from(text)` (or `Intl.Segmenter` if you want grapheme clusters) would be safer — `Array.from` is already used a few lines above on line 99.

src/components/Ellipsis/hooks.ts
```diff
@@ -0,0 +1,146 @@
...
+                startOffset,
+                endOffset,
+                setVisibleText,
+                measureRef,
+                containerRef,
+            }),
+        [text, startOffset, endOffset, setVisibleText],
+    );
+
+    // Does the same as useLayoutEffect, but does not trigger warnings with SSR
```
korvin89, line 126: 🤖 **AI generated**

🟠 **should-fix**

This ref-callback trick reimplements `src/hooks/useLayoutEffect/useLayoutEffect.ts`, which exists for exactly this reason:

```ts
export const useLayoutEffect = typeof window === 'undefined' ? () => {} : React.useLayoutEffect;
```

and is already used by `useCollapseChildren`, `useFocusWithin`, `useListNavigation` and `useTabListGapFromContainer`. So the comment on line 126 describes a problem the repo has already solved.

The hand-rolled version also swallows the detach call (`if (!node) return;`), which means `containerRef.current` keeps pointing at a removed DOM node after unmount.

`useLayoutEffect(onResize, [onResize])` plus a plain object ref would cover both.

src/hooks/useResizeObserver/useResizeObserver.ts
```diff
@@ -1,7 +1,10 @@
 import * as React from 'react';
 
 interface UseResizeObserverProps<T> {
-    ref: React.RefObject<T | null | undefined> | undefined;
+    ref:
```
korvin89, line 4: 🤖 **AI generated**

🟠 **should-fix**

This widens a public, documented hook, and I think the change costs more than it buys here.

Two concrete problems with the new shape:

- `ref` is in the effect deps (line 52), so any caller passing an array literal re-creates the `ResizeObserver` on every render. This PR has to work around it with a `useMemo` (`Ellipsis/hooks.ts:141`), and nothing in the types or the README tells the next caller they must.
- The guard on line 26 only catches a fully empty list. If some refs are not attached yet when the effect runs, those elements are never observed and nothing re-runs it.

On top of that, `ResizeInfo` does not carry which element resized, so with several elements the callback cannot tell them apart, and `src/hooks/useResizeObserver/README.md:14` still documents `ref: React.RefObject`.

Would it work to leave the shared hook alone and call it twice in `useCenterEllipsis` instead?

```ts
useResizeObserver({ref: containerRef, onResize});
useResizeObserver({ref: measureRef, onResize});
```

Both calls are unconditional, so there is no Rules-of-Hooks issue, and it removes the memoisation footgun, the partial-attachment hole and the public API change in one go. The cost is two observers per instance instead of one — worth measuring if `Ellipsis` is expected to appear in long lists. If that does turn out to matter, the array API is justified, but I would land it as its own change with a README update, a test for the array branch, and an explicit answer to the deps question.

src/components/Ellipsis/__tests__/Ellipsis.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+const FSI = '\u2068';
+const PDI = '\u2069';
+const stripIsolation = (value: string | null | undefined) =>
+    (value ?? '').replaceAll(FSI, '').replaceAll(PDI, '');
+
+const LONG_TEXT = 'a-very-long-long-text-that-should-be-truncated-in-somewhere.tar.gz';
+
+// The `center` position relies on ResizeObserver and document.fonts, which jsdom does
+// not implement, so for it we can only rely on the visual tests.
+describe('Ellipsis', () => {
```
korvin89, line 14: 🤖 **AI generated**

🟠 **should-fix**

There are 6 tests here and none of them covers the behaviour this PR adds. Missing: `position="center"` in any form, the copy handler, overlapping offsets, offsets larger than the text, empty `children`, a separator that is absent from the text, multi-character separators, RTL. The new array branch in `useResizeObserver` has no test either. And the one test that touches the copy promise strips the isolates before asserting, so it passes without checking what a real copy produces.

On the "center cannot be tested" note — I would suggest splitting it by what actually needs layout:

- **Pure logic does not.** `getPartsWithoutSeparator` / `getPartsWithSeparator` (and `calculateCollapsedText`, given a stub measurer) can be exported and unit-tested here directly. That is where the overlapping-offset and separator cases belong, and they are much sharper as assertions than as screenshots.
- **Anything depending on a measured width does.** Playwright is the right place for it and the visual suite is already set up. Concretely: `center` at a few container widths rather than the single 200px wrapper in `helpers.tsx`, and — this one would have caught the stale-probe bug — a scenario that changes `text` on an already-mounted instance.

Would that split work for you?

src/components/Ellipsis/__tests__/helpers.tsx
```diff
@@ -0,0 +1,16 @@
...
+import type {EllipsisProps} from '../Ellipsis';
+
+const wrapDivStyles = {width: '200px'} as const;
+
+const DEFAULT_TEXT = 'a-very-long-long-text-that-should-be-truncated-in-somewhere.tar.gz';
+
+export const TestEllipsis = ({children = DEFAULT_TEXT, ...props}: Partial<EllipsisProps>) => {
+    return (
+        <div style={wrapDivStyles}>
+            <Ellipsis {...props} style={{width: '100%'}}>
```
korvin89, line 11: 🤖 **AI generated**

🟠 **should-fix**

Every story and this visual-test helper pass `style={{width: '100%'}}`. That means nothing in the suite exercises the component in its default, out-of-the-box configuration — the harness is compensating for the default styles rather than testing them.

Could we drop the extra style from the helper and from the stories, and cover the component the way a consumer would first reach for it? If it does not lay out correctly without it, that is a defect in the default rules (`Ellipsis.scss:5-8`, `inline-flex` + `max-width: 100%`) worth fixing; if a definite container width is genuinely required, it would be good to state that in the README as a container contract.

src/components/Ellipsis/Ellipsis.scss
```diff
@@ -0,0 +1,53 @@
+@use '../variables';
+
+$block: '.#{variables.$ns}ellipsis';
+
+#{$block} {
+    display: inline-flex;
+    min-width: 0;
+    max-width: 100%;
+    white-space: nowrap;
+    overflow-wrap: break-word;
```
korvin89, line 10: 🤖 **AI generated**

🟠 **should-fix**

`overflow-wrap: break-word` and `word-break: break-all` cannot take effect here — `white-space: nowrap` on line 9 suppresses all wrapping, so both declarations are inert. I would suggest removing them, so the rule does not imply a wrapping behaviour that never happens.

src/components/Ellipsis/Ellipsis.scss
```diff
@@ -0,0 +1,53 @@
...
+        &-content {
+            max-width: fit-content;
+        }
+
+        $ellipsis: &;
+
+        &_start {
+            #{$ellipsis}-content {
+                direction: var(--g-flow-opposite);
+                overflow: hidden;
```
korvin89, line 27: 🤖 **AI generated**

🟠 **should-fix**

`overflow: hidden; text-overflow: ellipsis;` is spelled out twice (lines 27-28 and 34-35) where `styles/mixins.scss` already provides it:

```scss
```

That is what Text, Table, User, Select and Breadcrumbs use, so it would be good to stay consistent with them.

src/components/Ellipsis/Ellipsis.scss
```diff
@@ -0,0 +1,53 @@
...
+        min-width: 0;
+        overflow: hidden;
+
+        &-content {
+            max-width: fit-content;
+        }
+
+        $ellipsis: &;
+
+        &_start {
```
korvin89, line 24: 🤖 **AI generated**

🟠 **should-fix**

Coming back to point 7 from the earlier thread, since I do not think it is settled yet.

I take the point that JS measurement is expensive and that `text-overflow: <string>` would eventually make all of this unnecessary. But the gap is not uniformly harmless — it shows in two specific places:

- **`position="end"` with a non-zero trailing offset.** The ellipsis lands at the clip edge of the middle span with leftover space up to roughly a character wide, and the pinned tail span follows immediately after, so there is a visible gap between `…` and the tail. This is also the most common use of the component — keeping the file extension.
- **`position="start"`.** In the approved `Positions-1` snapshot in this PR the `start` row is already indented by about one character relative to the `center` and `end` rows below it, even though the labels are flush: the `direction: rtl` flip right-aligns the line and the sub-character remainder lands on the visible edge. In a list where rows are meant to line up, that is noticeable.

So the suggestion is not "use JS everywhere" — it is to narrow the pure-CSS path to the combinations where the gap genuinely is not visible (`position="end"` with no trailing offset covers most real usage) and measure for the rest. If the measurement cost is the blocker, it would help to see numbers for a realistic case — a list of a few hundred rows — so we can decide on evidence rather than on the concern.

src/components/Ellipsis/Ellipsis.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+export type EllipsisPosition = 'start' | 'center' | 'end';
+
+interface EllipsisContentProps {
+    startOffset: string;
+    endOffset: string;
+    children: string;
+}
+
+const CenterEllipsis = React.memo(
```
korvin89, line 20: 🤖 **AI generated**

🔵 **nit**

`React.memo` on `CenterEllipsis` (and on `EdgeEllipsis`, line 44) does not buy much: all of their props are derived from the parent's own props, so they change whenever the parent re-renders for a reason that matters. The only case it skips is a `className` / `style` / `qa`-only change. Not a problem, just probably not worth the wrapper.

src/components/Ellipsis/Ellipsis.tsx
```diff
@@ -0,0 +1,193 @@
...
+        children: text,
+        qa,
+    },
+    ref,
+) {
+    const ellipsisContentRef = React.useRef<HTMLSpanElement>(null);
+
+    const isCenterPosition = position === 'center';
+
+    const [startOffset, collapsibleText, endOffset] = React.useMemo<[string, string, string]>(
```
korvin89, line 143: 🤖 **AI generated**

🔵 **nit**

Three small readability points around this block:

- `EllipsisContentProps.children` (line 17) means "the full text" for `CenterEllipsis` and "the collapsible middle" for `EdgeEllipsis` — one type carrying two different meanings.
- The labelled tuple `TextParts` is declared on line 75 but not used at the one call site that would benefit; this `useMemo` re-spells `<[string, string, string]>`.
- `collapsibleText` is computed here and then discarded for `position="center"`, where `hooks.ts:28-31` recomputes the same slice from `startOffset.length` / `endOffset.length` — the same rule expressed in two places.

src/components/Ellipsis/__tests__/Ellipsis.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+// Isolation characters that wrap the truncated part to keep bidi rendering stable.
+const FSI = '\u2068';
+const PDI = '\u2069';
+const stripIsolation = (value: string | null | undefined) =>
+    (value ?? '').replaceAll(FSI, '').replaceAll(PDI, '');
+
+const LONG_TEXT = 'a-very-long-long-text-that-should-be-truncated-in-somewhere.tar.gz';
+
+// The `center` position relies on ResizeObserver and document.fonts, which jsdom does
```
korvin89, line 12: 🤖 **AI generated**

🔵 **nit**

This comment says `center` relies on `document.fonts`, but the current implementation does not use it at all — that went away with the rework in 9363713f. As written it gives a reason for the coverage gap that no longer holds.

Rather than explaining why `center` is not covered, it would be better to actually cover it in the Playwright suite (see the separate note on this file) and drop the caveat.

src/components/EllipsisText/EllipsisText.scss
```diff
@@ -0,0 +1,53 @@
...
+        min-width: 0;
+        overflow: hidden;
+
+        &-content {
+            max-width: fit-content;
+        }
+
+        $ellipsis: &;
+
+        &_start {
```
PahaN47: a few hundred lines with js noticeably lag. based on my personal testing, about ~100 lines or so is the bearable amount

PR #2692, opened 2026-05-29
feat(Suggest): add Suggest component

imgeorgius:


korvin89 (commented):
🤖 _AI generated_

Review of the new `Suggest` component. Solid test/story/docs coverage. Two themes inline: (1) several behaviors the docs advertise don't actually work yet, and (2) a new component like this should ship under `lab`/`unstable` first so the API can still evolve before it's frozen.

src/components/Suggest/Suggest.tsx
```diff
@@ -0,0 +1,259 @@
...
+            return (
+                <div className={b('loader')}>
+                    <Loader />
+                </div>
+            );
+        }
+
+        if (!items?.length) {
+            return null;
+        }
```
korvin89: 🤖 _AI generated_

`renderPopupContent` returns `null` when `items` is empty **before** `renderPopup` is ever called (lines 182–184), and `hasContent = loading || Boolean(items?.length)` (line 100) keeps the `Popup` closed via `open={hasContent && open}` (line 247).

As a result the empty-state branch of `renderPopup` (e.g. "Nothing matched '{value}'") never renders — even though that exact pattern is shown in the README (*Custom popup content*) and in the `CustomPopupContent` story. Repro: type a query with no matches → `items=[]` → the popup stays closed and `renderPopup` is not invoked.

When `renderPopup` is provided, it should be called (and the popup opened) even for an empty list.

src/components/Suggest/Suggest.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+const b = block('suggest');
+
+type SuggestComponent = <T>(
+    props: SuggestProps<T> & {ref?: React.Ref<HTMLSpanElement>},
+) => React.ReactElement;
+
+export const Suggest = React.forwardRef(function Suggest<T>(
+    {
+        value = '',
```
korvin89: 🤖 _AI generated_

The default `value = ''` forces `TextInput` into controlled mode permanently: a value is always passed via `value={value}` (line 225), and inside `TextInput` → `useControlledState(value, defaultValue ?? '', …)` where `isControlled = value !== undefined` is therefore always `true`.

Consequences: the documented `defaultValue` (uncontrolled mode, types.ts:15) is ignored, and in a pure uncontrolled scenario (only `defaultValue`, no `value`/`onUpdate`) the field is frozen empty and cannot be typed into.

Declare `value` without a default so it stays `undefined` when the prop is omitted, letting `TextInput` fall back to uncontrolled mode.

src/components/Suggest/Suggest.tsx
```diff
@@ -0,0 +1,259 @@
...
+    const hasContent = loading || Boolean(items?.length);
+
+    const handleValueChange = React.useCallback(
+        (newValue: string) => {
+            onUpdate?.(newValue);
+            if (newValue) {
+                toggleOpen(true);
+            } else {
+                toggleOpen(false);
+            }
```
korvin89: 🤖 _AI generated_

`handleValueChange` unconditionally calls `toggleOpen(true/false)` based on whether there's a value, ignoring externally controlled `open`/`onOpenChange`.

Repro from the docs (*Show options on empty input* / multi-select): the consumer sets `setOpen(true)` in `onUpdate`, but when the field is cleared `handleValueChange('')` immediately follows with `toggleOpen(false)` → `onOpenChange(false)`, closing the popup. This breaks the advertised "keeping a multi-select popup open after clearing" use case.

The auto-open logic should only apply when `open` is uncontrolled.

src/components/Suggest/Suggest.tsx
```diff
@@ -0,0 +1,259 @@
...
+                        activeIndex === undefined ? undefined : `${listId}-item-${activeIndex}`,
+                }}
+            />
+            <Popup
+                placement="bottom-start"
+                {...popupProps}
+                id={popupId}
+                open={hasContent && open}
+                onOpenChange={toggleOpen}
+                anchorElement={inputWrapperRef.current}
```
korvin89: 🤖 _AI generated_

`anchorElement={inputWrapperRef.current}` (line 249) and the `popupStyle` width for `popupWidth='fit'` via `inputWrapperRef.current.offsetWidth` (lines 88–89) are read during render. On the first render the ref is `null`, and assigning a ref doesn't trigger a re-render.

Repro: with `defaultOpen`/`open=true` on mount, the popup opens with `anchorElement={null}` and no width → mispositioned until some external re-render populates the ref. The usual "type → re-render → open" flow only works by accident, since the ref is already set by the time the popup opens.

Better to keep the anchor/width in state (callback ref) or compute it in `useLayoutEffect`.

src/components/index.ts
```diff
@@ -51,6 +51,7 @@ export * from './Skeleton';
 export * from './Slider';
 export * from './Stepper';
 export * from './Spin';
+export * from './Suggest';
```
korvin89: 🤖 _AI generated_

This exports `Suggest` straight into the stable public API. In this repo new components are introduced under `src/components/lab/*` and re-exported from `src/unstable.ts` with the `unstable_` prefix (e.g. `unstable_Menu`, `unstable_ColorPicker`, `unstable_FileDropZone`).

A freshly added component usually keeps evolving after its first release, so shipping `Suggest` as stable freezes the API: any breaking change (e.g. the `items → options` rename suggested below) would then require a major bump. Placing it in `lab` as `unstable_Suggest` leaves room to land breaking changes within minor releases until the API settles.

src/components/Suggest/Suggest.tsx
```diff
@@ -0,0 +1,259 @@
...
+                className={b('list')}
+                style={virtualizedHeight ? {height: virtualizedHeight} : undefined}
+            >
+                <List<T>
+                    ref={listRef}
+                    id={listId}
+                    role="listbox"
+                    filterable={false}
+                    items={items}
+                    renderItem={renderItem}
```
korvin89: 🤖 _AI generated_

Consider using `ListItemView` as the default option renderer (when `renderItem`/`renderOption` is not provided) instead of relying on List's plain `defaultRenderItem`. It gives a look consistent with the rest of the library and covers the common option layouts out of the box.

It's currently `unstable_ListItemView` (https://preview.gravity-ui.com/uikit/?path=/story/lab-uselist-listitemview--list-item-view) and is planned to be promoted out of `unstable` in the next major — which fits well with shipping `Suggest` in `lab` for now.

src/components/Suggest/types.ts
```diff
@@ -0,0 +1,87 @@
...
+    renderItem?: ListProps<T>['renderItem'];
+    /** Enable virtualization for long lists */
+    virtualized?: boolean;
+    /**
+     * Height of the scrollable list viewport in pixels when `virtualized` is enabled.
+     * Defaults to `300`. Has no effect when `virtualized` is `false`.
+     */
+    listHeight?: number;
+    /** Returns height for each option row (enables variable-height rows) */
+    getOptionHeight?: (option: ListItemData<T>, index: number) => number;
```
korvin89: 🤖 _AI generated_

The naming of this prop group is inconsistent: `items` / `onItemClick` / `renderItem` use "item", while the adjacent `getOptionHeight` (line 36) uses "option".

Suggest aligning everything on "option", as in `Select` (`options`, `renderOption`, `getOptionHeight`): `items → options`, `onItemClick → onOptionClick`, `renderItem → renderOption`. This removes the mismatch and keeps the API consistent with the rest of the library.

playwright/run-docker-command.sh
```diff
@@ -20,10 +20,10 @@ run_command() {
     /bin/bash -c "$*"
 }
 
-if command_exists docker; then
-  CONTAINER_TOOL="docker"
-elif command_exists podman; then
+if command_exists podman; then
```
korvin89: 🤖 _AI generated_

This change (preferring `podman` over `docker`) is unrelated to the `Suggest` component and alters the playwright run environment for all contributors: anyone with both tools installed will switch from docker to podman.

Consider splitting it into a separate PR or explaining the rationale in the description.

src/components/Suggest/Suggest.tsx
```diff
@@ -0,0 +1,259 @@
...
+                    'aria-activedescendant':
+                        activeIndex === undefined ? undefined : `${listId}-item-${activeIndex}`,
+                }}
+            />
+            <Popup
+                placement="bottom-start"
+                {...popupProps}
+                id={popupId}
+                open={hasContent && open}
+                onOpenChange={toggleOpen}
```
korvin89: 🤖 _AI generated_

`onOpenChange` is typed with `Popup`'s signature `(open, event?, reason?)` (types.ts:67), but at runtime it's routed through `useOpenState`'s `toggleOpen`, which only carries the boolean: `Popup` calls `onOpenChange(isOpen, event, reason)` → `toggleOpen(val)` → `handleOpenChange(newOpen)` → `onOpenChange(newOpen)`. So `event` and `reason` always arrive as `undefined`, unlike a plain `Popup`.

Forward `Popup`'s full `(open, event, reason)` to the consumer — e.g. pass `onOpenChange` to `Popup` directly, or thread the extra args through `toggleOpen` — so the advertised (and `Popup`-compatible) signature is actually honored.

src/components/Suggest/Suggest.tsx
imgeorgius: Thanks! Leaning against ListItemView here for now.

Suggest uses the class List, which already renders the full row (hover/active/selected + click) — see the row wrapper with onClick/onMouseEnter and the b('item', {active, selected, …}) classes in [ListItem.tsx#L63-L90](https://github.com/gravity-ui/uikit/blob/main/src/components/List/components/ListItem.tsx#L63-L90), styled in [List.scss#L34-L45](https://github.com/gravity-ui/uikit/blob/main/src/components/List/List.scss#L34-L45). A custom renderItem only fills the inner .item-content ([ListItem.tsx#L110-L113](https://github.com/gravity-ui/uikit/blob/main/src/components/List/components/ListItem.tsx#L110-L113)). ListItemView is a full row too, so nesting it inside a List row would double the highlight and interaction.

The consistent fix for a List-based component is a lightweight content renderer – same as Select's [DefaultOption](https://github.com/gravity-ui/uikit/blob/main/src/components/Select/components/SelectList/OptionWrap.tsx#L21-L28) – not ListItemView.

imgeorgius: Related question I'd want to sort out first: does the option even need to be fully generic? Today Suggest's option is [ListItemData<T> = T & {disabled}](https://github.com/gravity-ui/uikit/blob/main/src/components/List/types.ts#L12), so there's no guaranteed content field – that's really why the default renderer is awkward.

Select does it differently: a fixed option shape ([ControlGroupOption](https://github.com/gravity-ui/uikit/blob/main/src/components/types.ts#L45-L51): value, content, disabled, …) with the generic scoped to just a [data?: T](https://github.com/gravity-ui/uikit/blob/main/src/components/Select/types.ts#L162-L166) payload. 

Maybe we should follow the same approach – that would make a content-based default type-safe and match how Select models its options.

imgeorgius: Circled back on this with the team – here's where it landed:

Option shape – went Select-like. The option is now a fixed shape (`SuggestOption = QAProps & ControlGroupOption & {data?: T}`, same base as `SelectOption`), with the generic scoped to the `data` payload. `options` / `onOptionClick` / `renderOption` / `getOptionHeight` are all typed on it.

Default renderer – now that `content` is guaranteed, the default renders `option.content` (falling back to children), same as Select's `DefaultOption`.

`ListItemView` – holding off: the upcoming major ships a new `List` that uses `ListItemView` by default, so once Suggest moves onto it the styled row comes for free with nothing to wire on our side. Until then, nesting `ListItemView` inside the class `List` would double the styles and handlers, so the Select-style `content` default keeps it consistent.

Done in the [follow-up commit](https://github.com/gravity-ui/uikit/pull/2692/changes/3df3412f73668e2ac66041582b01ac3eadfbde2d).

PR #2707, opened 2026-06-18
refactor!: migrate from lodash to es-toolkit

vtaits:
Hello. It's a suggestion to migrate from `lodash` to `es-toolkit`.

The main advantages are: it's smaller, faster, more es-friendly and ts-friendly and popular enough.

https://github.com/toss/es-toolkit

## Summary by Sourcery

Replace lodash usage across the codebase with es-toolkit utilities and update dependencies accordingly.

Enhancements:
- Refactor components, hooks, and HOCs to use es-toolkit helpers instead of lodash for common operations.
- Update list, table, pagination, layout, and legacy UI utilities to align with es-toolkit APIs while preserving behavior.

Build:
- Swap lodash for es-toolkit in runtime dependencies and remove @types/lodash from development dependencies in package.json.


ogonkov:
Why not lodash-es then?

amje:
> Why not lodash-es then?

I suppose lodash-es has the same flaws as normal lodash: legacy code, bigger size, slower and poor TypeScript support

ogonkov:
What is a size difference in uikit case between lodash-es and new library?

In our case we already have lodash-es in project, and new library woul be just additional code, that won't be reused

amje:
> What is a size difference in uikit case between lodash-es and new library?

It depends whether or not your other dependencies bring es-toolkit into bundle.

> In our case we already have lodash-es in project, and new library woul be just additional code, that won't be reused

So normal lodash is also adds additional code that is not resused. This PR should reduce your bundle size, cuz es-toolkit is much smaller than lodash.

ogonkov:
We already making our bundle smaller by aliasing lodash to lodash-es in our bundler (this is triggering tree shacking for lodash imports)

PR #2708, opened 2026-06-18
feat: default props provider

amje:
## Summary by Sourcery

Introduce a themable default props system and integrate it across core UI components.

New Features:
- Add a default props context and hook that allow configuring component default props via ThemeProvider.

Enhancements:
- Wire multiple components (inputs, layout, feedback, navigation, overlay, etc.) to consume default props from the new context.
- Extend props merging utilities to support concatenating className-style fields when combining defaults with user props.
- Expose the default props map type from the theme module for external configuration.

Tests:
- Add unit tests for the useDefaultProps hook and its interaction with ThemeProvider.
- Add unit tests for the mergeProps utility to validate the new merging behavior for className and handlers.

### DIFF
--- a/src/components/theme/PrivateDefaultPropsProvider.tsx
+++ b/src/components/theme/PrivateDefaultPropsProvider.tsx
@@ -1,0 +1,144 @@
+'use client';
+
+import * as React from 'react';
+
+import type {AccordionProps} from '../Accordion';
+import type {ActionTooltipProps} from '../ActionTooltip';
+import type {ActionsPanelProps} from '../ActionsPanel';
+import type {AlertProps} from '../Alert';
+import type {ArrowToggleProps} from '../ArrowToggle';
+import type {AvatarProps} from '../Avatar';
+import type {AvatarStackProps} from '../AvatarStack';
+import type {BreadcrumbsProps} from '../Breadcrumbs';
+import type {ButtonCommonProps} from '../Button/Button';
+import type {CardProps} from '../Card';
+import type {CheckboxProps} from '../Checkbox';
+import type {ClipboardButtonProps} from '../ClipboardButton';
+import type {ClipboardIconProps} from '../ClipboardIcon';
+import type {CopyToClipboardProps} from '../CopyToClipboard';
+import type {DefinitionListProps} from '../DefinitionList';
+import type {DialogProps} from '../Dialog';
+import type {DisclosureProps} from '../Disclosure';
+import type {DividerProps} from '../Divider';
+import type {DrawerProps} from '../Drawer';
+import type {DropdownMenuProps} from '../DropdownMenu';
+import type {FilePreviewProps} from '../FilePreview';
+import type {HelpMarkProps} from '../HelpMark';
+import type {HotkeyProps} from '../Hotkey';
+import type {IconProps} from '../Icon';
+import type {LabelProps} from '../Label';
+import type {LinkProps} from '../Link';
+import type {LoaderProps} from '../Loader';
+import type {MenuProps} from '../Menu';
+import type {ModalProps} from '../Modal';
+import type {NumberInputProps} from '../NumberInput';
+import type {OverlayProps} from '../Overlay';
+import type {PaginationProps} from '../Pagination';
+import type {PaletteProps} from '../Palette';
+import type {PinInputProps} from '../PinInput';
+import type {PlaceholderContainerProps} from '../PlaceholderContainer';
+import type {PopoverProps} from '../Popover';
+import type {PopupProps} from '../Popup';
+import type {PortalProps} from '../Portal';
+import type {ProgressProps} from '../Progress';
+import type {RadioProps} from '../Radio';
+import type {RadioGroupProps} from '../RadioGroup';
+import type {SegmentedRadioGroupProps} from '../SegmentedRadioGroup';
+import type {SelectProps} from '../Select';
+import type {SheetProps} from '../Sheet';
+import type {SkeletonProps} from '../Skeleton';
+import type {SliderProps} from '../Slider';
+import type {SpinProps} from '../Spin';
+import type {StepperProps} from '../Stepper';
+import type {SwitchProps} from '../Switch';
+import type {TableColumnSetupProps} from '../TableColumnSetup';
+import type {TextProps} from '../Text';
+import type {TocProps} from '../Toc';
+import type {TooltipProps} from '../Tooltip';
+import type {UserProps} from '../User';
+import type {UserLabelProps} from '../UserLabel';
+import type {TextAreaProps} from '../controls/TextArea';
+import type {TextInputProps} from '../controls/TextInput';
+import type {TabListProps, TabPanelProps, TabProps, TabProviderProps} from '../tabs';
+
+export interface ComponentDefaultPropsMap {
+    Accordion?: Partial<AccordionProps<any>>;
+    ActionsPanel?: Partial<ActionsPanelProps>;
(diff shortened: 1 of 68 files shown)


DakEnviy (approved):

korvin89 (commented):
🤖 _AI generated_

Docs (not covered by the inline comments): for a public feature of this scope, please add
- docs for `ThemeProvider.defaultProps` (README / docs site);
- a Storybook story covering the edge cases raised inline (explicit `undefined`, handler chaining, nested providers).

src/components/utils/mergeProps.ts
```diff
@@ -24,6 +24,12 @@ export function mergeProps(...propsList: (React.HTMLProps<HTMLElement> & {})[])
...
                     };
                 }
+            } else if (
+                (key === 'className' || key.endsWith('ClassName')) &&
+                typeof acc[key] === 'string' &&
+                typeof value === 'string'
+            ) {
+                acc[key] = `${acc[key]} ${value}`;
             } else {
                 acc[key] = value;
```
korvin89, line 34: 🤖 _AI generated_

A theme default is applied only when the prop is **fully omitted**. As soon as a user passes an explicit `undefined` — the idiomatic passthrough pattern (`<Button size={props.size} />`, conditional props, spreads from wrappers) — it overwrites the default, and the component falls back to its own internal default.

Reproduced on this branch with the real `Button`:

```tsx
// theme default: {Button: {size: 'l'}}
<Button size={undefined}>…</Button>   // → g-button_size_m  (default lost)
<Button>…</Button>                    // → g-button_size_l  (default applied)
```

Root cause: this `else` branch assigns `acc[key] = value` even when `value === undefined`.

MUI handles this in `resolveProps` (a key is taken from defaults when the external value is `=== undefined`). Suggest fixing in `useDefaultProps` rather than the shared `mergeProps` (see the comment there) — strip `undefined` user props before merging:

```ts
const cleaned: Record<string, unknown> = {};
for (const k in props) if (props[k as keyof T] !== undefined) cleaned[k] = props[k as keyof T];
return mergeProps(defaults as any, cleaned) as unknown as T;
```

Worth adding a regression test for the `size={undefined}` case.

src/components/utils/mergeProps.ts
```diff
@@ -24,6 +24,12 @@ export function mergeProps(...propsList: (React.HTMLProps<HTMLElement> & {})[])
                             .find((v) => v !== undefined);
                     };
                 }
+            } else if (
+                (key === 'className' || key.endsWith('ClassName')) &&
+                typeof acc[key] === 'string' &&
+                typeof value === 'string'
+            ) {
+                acc[key] = `${acc[key]} ${value}`;
```
korvin89, line 32: 🤖 _AI generated_

`mergeProps` isn't used only by `useDefaultProps` — `lab/Menu/MenuItem` calls `mergeProps(triggerProps, componentProps)` ([MenuItem.tsx#L217](https://github.com/gravity-ui/uikit/blob/default-props-provider/src/components/lab/Menu/MenuItem.tsx#L217)). Previously `className`/`*ClassName` were **overwritten** (last wins); with this added branch they're **concatenated**. That's an unannounced behavior change for existing code.

For MenuItem it's probably more correct, but please verify the submenu trigger case and note it in the PR description. If concatenation is only needed for the default-props feature, consider keeping that logic inside `useDefaultProps` and leaving the shared `mergeProps` untouched.

src/components/theme/useDefaultProps.ts
```diff
@@ -0,0 +1,21 @@
...
+): T {
+    const ctx = React.useContext(PrivateDefaultPropsContext);
+    const defaults = ctx[componentName];
+
+    if (!defaults) {
+        return props;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    return mergeProps(defaults as any, props as any) as unknown as T;
```
korvin89: 🤖 _AI generated_

Via `mergeProps`, a default `on*` handler from the theme is **chained** with the user handler rather than overridden, and the return value comes from the **first** handler (`.find(v => v !== undefined)` → the default's). Two consequences worth a product decision:

- The user **cannot override** a theme default handler — it always runs first.
- For handlers where the return value matters, the default wins, which is unintuitive.

This differs from the usual "defaultProps" mental model (defaults are overridable). Please decide whether default handlers should be **overridden** (like regular defaults) or **chained**, and document the chosen behavior.

src/components/theme/useDefaultProps.ts
```diff
@@ -0,0 +1,21 @@
...
+export function useDefaultProps<K extends keyof ComponentDefaultPropsMap, T extends object>(
+    componentName: K,
+    props: T,
+): T {
+    const ctx = React.useContext(PrivateDefaultPropsContext);
+    const defaults = ctx[componentName];
+
+    if (!defaults) {
+        return props;
+    }
```
korvin89: 🤖 _AI generated_

A performance note about the merge below: because `mergeProps` wraps every `on*` handler in a fresh closure (even when defaults don't define that handler), setting a single style default destabilizes all handler identities. E.g. `defaultProps={{Button: {size: 'l'}}}` makes every `Button`'s `onClick` a new function on every render — breaking referential stability for memoized children and effect deps. `useDefaultProps` also recreates the props object each render with no memoization.

Mitigating factor: this fast path (`if (!defaults) return props`) means only components with configured defaults pay the cost, which is reasonable. Still worth (a) documenting it and (b) considering `React.useMemo` keyed on `[defaults, props]`, and/or not re-wrapping handlers that aren't present in defaults.

src/components/theme/PrivateDefaultPropsProvider.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+export function PrivateDefaultPropsProvider({
+    value = EMPTY,
+    children,
+}: {
+    value?: ComponentDefaultPropsMap;
+    children: React.ReactNode;
+}) {
+    return (
+        <PrivateDefaultPropsContext.Provider value={value}>
```
korvin89, line 140: 🤖 _AI generated_

A nested `ThemeProvider` fully replaces the outer `defaultProps` instead of merging (the test `inner ThemeProvider defaultProps override outer` documents `{view:'action'}` wiping `{view:'outlined', size:'l'}`). This differs from how themes compose in MUI and from the intuition that a nested provider *refines* the outer config.

At minimum, document this prominently. Ideally, merge with the parent context here:

```tsx
const parent = React.useContext(PrivateDefaultPropsContext);
const merged = React.useMemo(() => mergeDefaults(parent, value), [parent, value]);
// <PrivateDefaultPropsContext.Provider value={merged}>
```

src/components/utils/mergeProps.ts
amje: The concatination of classNames is correct for all cases

src/components/theme/useDefaultProps.ts
amje: I've chosen normal override behavior

src/components/utils/mergeProps.ts
amje: Fixed

src/components/theme/useDefaultProps.ts
amje: I've chosen normal override behavior. No merging on* handlers anymore

src/components/theme/PrivateDefaultPropsProvider.tsx
amje: Fixed

korvin89 (approved):

PR #2710, opened 2026-06-22
fix(TextInput): replace legacy `Popover` with new one

ogonkov:
## Summary by Sourcery

Bug Fixes:
- Ensure TextInput error messages are displayed via the new Popover implementation using an accessible Alert container.

### DIFF
--- a/src/components/controls/TextInput/TextInput.scss
+++ b/src/components/controls/TextInput/TextInput.scss
@@ -148,6 +148,14 @@ $block: '.#{variables.$ns}text-input';
             var(--_--error-icon-padding-inline-end);
     }
 
+    &__error-popover {
+        max-width: 300px;
+    }
+
+    &__error-popover-content {
+        padding: var(--g-spacing-5) var(--g-spacing-4);
+    }
+
     &__additional-content {
         display: flex;
         align-items: center;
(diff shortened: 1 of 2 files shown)


korvin89:
We need to keep the old pop-up opening position.

ogonkov:
@amje why there is a positioning difference between popover and legacy popover?

ogonkov:
@korvin89 looks like in legacy Popover direction is defaulted to right/bottom, i can backport this behaviour

https://github.com/gravity-ui/uikit/blob/65e2728a609470040d9e768ecd7526e1064feacc/src/components/legacy/Popover/Popover.tsx#L102

korvin89:
> @korvin89 looks like in legacy Popover direction is defaulted to right/bottom, i can backport this behaviour
> 
> https://github.com/gravity-ui/uikit/blob/65e2728a609470040d9e768ecd7526e1064feacc/src/components/legacy/Popover/Popover.tsx#L102

As part of this pull request, we need to return the old behavior, as we discussed it separately with the design.

You can of course make a report in Popup

ogonkov:
@korvin89 tests passed

korvin89 (approved):

PR #2721, opened 2026-07-01
feat(Dialog): introduce responsive size props

amje:
## Summary by Sourcery

Introduce responsive dialog sizing and mobile-specific modal behavior, including experimental mobile modal support and updated Dialog APIs and styles for width control.

New Features:
- Add `maxWidth` and `fullWidth` props to Dialog to control responsive width and deprecate the old `size` prop.
- Enable experimental mobile modal behavior wired through MobileProvider and MobileContext to render Dialog/Modal in full-screen mobile layout.

Enhancements:
- Adjust Dialog, Modal, and their subcomponents (header, footer, close button) to adapt layout, typography, and button sizes for mobile contexts.
- Update Modal and Dialog CSS variables and docs to support max width/height configuration and clarify documented prop and token types.


src/components/Modal/Modal.scss
```diff
@@ -27,13 +27,12 @@ $modal-transform-transition: transform $modal-transition-duration ease-out;
...
-    &__content-wrapper {
-        margin: var(--g-modal-margin, #{variables.$modal-default-margin});
-    }
-
     &__content {
+        margin: var(--g-modal-margin, #{variables.$modal-default-margin});
         height: var(--g-modal-height, auto);
+        max-height: var(--g-modal-max-height, none);
         width: var(--g-modal-width, auto);
+        max-width: var(--g-modal-max-width, none);
```
korvin89, line 35: 🤖 AI generated

**[bug] `maxWidth` looks like it gets ignored when `contentOverflow="auto"`.** This `max-width: var(--g-modal-max-width, none)` and the `&_has-scroll { max-width: calc(100vw - margin*2) }` rule below have the same specificity (0,1,0), and `_has-scroll` comes later in the source, so it wins — the `maxWidth` cap ends up overridden by ~100vw.

Repro: `<Dialog maxWidth="s" contentOverflow="auto">` with content wider than 480px grows past 480px (with `contentOverflow="visible"` the cap holds).

One option is to let the var win in the scroll case too, e.g. `&_has-scroll { max-width: var(--g-modal-max-width, calc(100vw - …)); }`.

src/components/mobile/MobileContext.ts
```diff
@@ -17,13 +17,15 @@ export interface Location {
 
 export interface MobileContextProps {
     mobile: boolean;
+    __experimentalMobileModals: boolean;
```
korvin89: 🤖 AI generated

**[api] Could this be optional (`__experimentalMobileModals?: boolean`)?** It's a required field on the publicly exported `MobileContextProps` (and transitively via `WithMobileProps extends MobileContextProps`), so it can break types for anyone building the context value by hand (`const v: MobileContextProps = {…}` → `TS2741`). `initialValue` already provides a default, so `?` would be enough. Narrow case, but cheap.

src/components/Dialog/README.md
```diff
@@ -37,23 +37,24 @@ const dialogTitleId = 'app-confirmation-dialog-title';
...
+| onOutsideClick        | Event handler on a mouse click outside the dialog                                             | `(event: MouseEvent) => void`                                                                    |             |
+| onClose               | Event handler on closing the dialog                                                           | `(event: MouseEvent or KeyboardEvent,` `reason: ModalCloseReason or "closeButtonClick") => void` |             |
+| className             | `className` of the dialog content wrapper                                                     | `string`                                                                                         |             |
+| modalClassName        | `className` of modal box in which the dialog is embedded                                      | `string`                                                                                         |             |
+| maxWidth              | Dialog content max width                                                                      | `'s'` `'m'` `'l'`                                                                                |             |
```
korvin89, line 50: 🤖 AI generated

**[docs] I'd suggest keeping the `size` row here**, marked deprecated and pointing to `maxWidth`/`fullWidth`. The prop is only `@deprecated` and still works, so removing it from the table leaves current `size` users without a migration hint. Worth noting the equivalent of `size` is `maxWidth` + `fullWidth` — `maxWidth` alone changes the semantics from a fixed width to a max width.

src/components/Modal/Modal.scss
amje: Fixed

src/components/mobile/MobileContext.ts
amje: Fixed

src/components/Dialog/README.md
amje: Updated docs

korvin89 (approved):

PR #2732, opened 2026-07-07
chore(Breadcrumbs): add story reproducing false collapse on content-sized container

ykamendrovskiy:
## What & why

`Breadcrumbs` with a few short items collapses part of them into the `…` popover immediately — even when there is plenty of free space and nothing overflows. It happens whenever the breadcrumbs' `<ol>` is sized to its content (shrink-wrap), e.g. when `<Breadcrumbs>` sits in a flex row without `flex: 1` — a very common page-header layout.

This PR adds a **Storybook story** reproducing the issue: `Components/Navigation/Breadcrumbs` → **`FalseCollapseOnContentWidth`**. No fix is included — the proposed solution is described below for your consideration.

## Root cause

`useCollapseChildren` compares two measurements taken in **different box models**:

- the container — `container.getBoundingClientRect().width`, which on the flex `<ol>` equals the sum of the items' **margin-boxes**;
- each item — measured with `getBoundingClientRect().width`, i.e. the **border-box, without margins**.

The current item (`.g-breadcrumbs__item_current`) carries a negative inset margin (`margin: -2px`, paired with `padding: 2px` to inset the focus ring). So on a content-sized container the summed item widths exceed the container width by that margin, `availableWidth` becomes `-4px < 0`, and the list collapses. It is deterministic (measured Δ = exactly −4px), not a sub-pixel or real-overflow effect.

## Proposed solution (not applied)

Measure items in the same box model as the container by including their horizontal margins in `Breadcrumbs`' `getChildWidth`:

```diff
 getChildWidth: (child) => {
-    const width = child.getBoundingClientRect().width;
+    const {marginLeft, marginRight} = window.getComputedStyle(child);
+    const width =
+        child.getBoundingClientRect().width + parseFloat(marginLeft) + parseFloat(marginRight);
     const maxWidth = child.dataset.current ? 200 : Infinity;
     return Math.min(maxWidth, width);
 },
```

With this change the summed item widths equal the container width (`availableWidth === 0`), so the list no longer collapses when it fits, while real overflow still collapses as before. I verified it locally against the story above.

**Why scoped to `Breadcrumbs` and not the hook.** The mismatch itself is more general — `useCollapseChildren` measures the container in its margin-box but each child in its border-box, so any consumer whose children carry margins can hit it. The other consumer, `Tabs`, already accounts for its item margins by passing them as the hook's `gap`, so making the hook's *default* measurement margin-aware would double-count them for `Tabs`. Hence the local fix. If you'd rather solve it once at the hook, that path would need to reconcile the two (e.g. document the box-model contract the hook expects, or make margin-vs-`gap` handling explicit).

Happy to open a follow-up with the fix + a regression test if you'd prefer.


ykamendrovskiy:
hey guys, will you check the preview please?
@korvin89 @ValeraS

ValeraS:
@ykamendrovskiy fixed here #2818

PR #2753, opened 2026-07-15
fix: add `@types/react` to `peerDependencies`

ogonkov:
Compiled output imports React types, but package didn't declare @types/react as a dependency. Under pnpm's strict node_modules isolation, transitive/hoisted @types/react isn't guaranteed resolvable, breaking consumer typechecks.

Add @types/react as an optional peerDependency to make the requirement explicit without forcing it on JS-only consumers.

## Summary by Sourcery

New Features:
- Expose @types/react as an optional peer dependency for consumers that rely on TypeScript React types.


package.json
```diff
@@ -211,9 +211,15 @@
     "typescript": "^5.9.3"
   },
   "peerDependencies": {
+    "@types/react": "*",
```
amje: Let’s keep it in sync with react: ^16 || ^17 || ^18 || ^19

amje (changes requested):

package.json
ogonkov: Changed

amje (approved):

PR #2754, opened 2026-07-16
chore(DefaultDisclosureSummary): new render props

GermanVor:
WIP

## Summary by Sourcery

Add layout customization options to Disclosure.Summary and expose the default summary component for reuse.

New Features:
- Introduce width and justifyContent render props to DisclosureSummary to control trigger width and content alignment.
- Expose DefaultDisclosureSummary via DisclosureSummary.Default for easier access to the default summary implementation.

Enhancements:
- Extend Disclosure summary styles to support full-width triggers and space-between content alignment.
- Update Russian and English Disclosure.Summary documentation to describe the new layout props and defaults.


src/components/Disclosure/DisclosureSummary/DisclosureSummary.tsx
```diff
@@ -29,6 +32,8 @@ export interface DisclosureSummaryRenderFunctionProps extends QAProps {
     onKeyDown?: (e: React.KeyboardEvent<HTMLButtonElement>) => void;
     disabled?: boolean;
     className?: string;
+    width?: DisclosureSummaryWidth;
+    justifyContent?: DisclosureSummaryJustifyContent;
```
Raubzeug, line 36: `width` and `justifyContent` are documented as `Disclosure.Summary` props, but they are declared on `DisclosureSummaryRenderFunctionProps`. `DisclosureSummary` does not include either value in the object passed to the render function, while `DisclosureSummaryProps` does not accept them at all. As a result, the API shown in the README cannot type-check. Could we move these options to `DisclosureSummaryProps` and pass them to the default summary when it is created?

Raubzeug (commented):
One API issue to address:

src/components/Disclosure/DisclosureSummary/DisclosureSummary.tsx
```diff
@@ -21,6 +21,9 @@ function warnAboutPhysicalValues() {
     );
 }
 
+export type DisclosureSummaryWidth = 'auto' | 'max';
+export type DisclosureSummaryJustifyContent = 'initial' | 'space-between';
```
Raubzeug: The implementation accepts `"initial" | "space-between"`, while both READMEs document `"start" | "space-between"`. This makes the documented `justifyContent="start"` value invalid in TypeScript. Could we use the same values throughout? I would prefer `"start" | "space-between"` because `initial` is a CSS-wide keyword rather than a semantic layout option.

src/components/Disclosure/DisclosureSummary/DisclosureSummary.tsx
```diff
@@ -93,4 +113,5 @@ export const DefaultDisclosureSummary = React.forwardRef<
     );
 });
 
+DisclosureSummary.Default = DefaultDisclosureSummary;
```
Raubzeug, line 159: Non-blocking suggestion: if `width` and `justifyContent` become `DisclosureSummaryProps`, would it make sense to make `children` optional and render `defaultSummary` when it is omitted? That would support `<Disclosure.Summary width="max" justifyContent="space-between" />` without forcing consumers to recreate `Disclosure.Summary.Default` and spread the interaction props manually. The existing render-function behavior could remain unchanged whenever `children` is provided.

src/components/Disclosure/Disclosure.scss
```diff
@@ -37,6 +37,12 @@ $gap: 8px;
     &__trigger_arrow_end {
         flex-direction: row-reverse;
     }
+    &__trigger_width_max {
+        width: 100%;
+    }
+    &__trigger_justify-content_space-between {
+        justify-content: space-between;
```
Raubzeug, line 50: Could we add regression coverage for these new layout modifiers? At minimum, a component test should verify that `width_max` and `justify-content_space-between` are applied and that omitting the props preserves the current classes. A Storybook or visual case with an end-positioned arrow and a long summary would also verify that the full-width layout behaves as intended.

GermanVor: 1a8460094942894fa53e80c7a2eb460914ebe530

src/components/Disclosure/DisclosureSummary/DisclosureSummary.tsx
GermanVor: had to add more props for `<Disclosure.Summary width="max" justifyContent="space-between" />`

2c77faf18da9782e59de8cec3018971605b0c140

src/components/Disclosure/DisclosureSummary/DisclosureSummary.tsx
GermanVor: 416cf61795faac23196085ff482fc12ca0876c8f

src/components/Disclosure/DisclosureSummary/DisclosureSummary.tsx
GermanVor: 2c77faf18da9782e59de8cec3018971605b0c140

Raubzeug: Thanks — the new props are now accepted by `DisclosureSummaryProps`, but the render-function contract is still expanded and all layout/content values are included in `props` (lines 74–86). This breaks the documented `{(props) => <Button {...props}>…}` path: a `size="xl"` Disclosure now changes the custom Button to `xl` and adds `width_auto`, while `summary` is forwarded to the DOM and React warns about `justifyContent`/`arrowPosition`. I reproduced this against the current head. Could we keep render props interaction-only and introduce separate `DefaultDisclosureSummaryProps` (or pass these values only when building `defaultSummary`)?

src/components/Disclosure/DisclosureSummary/DisclosureSummary.tsx
```diff
@@ -85,12 +148,13 @@ export const DefaultDisclosureSummary = React.forwardRef<
             ref={ref}
         >
             <ArrowToggle
-                size={ComponentSizeToIconSizeMap[size]}
+                size={ComponentSizeToIconSizeMap[mergedSize]}
```
Raubzeug, line 151: `size` currently overrides only the `ArrowToggle` icon through `mergedSize`. The trigger typography is still selected by the parent `.g-disclosure_size_* .g-disclosure__trigger` rules, so `<Disclosure size="m"><Disclosure.Summary size="xl" /></Disclosure>` renders an `xl` icon with `m` text. This contradicts the documented “overrides `Disclosure.size`” contract. Could we either remove this extra override from the scope, or add a trigger-level size modifier so the entire summary changes size?

src/components/Disclosure/DisclosureSummary/DisclosureSummary.tsx
```diff
@@ -29,38 +32,94 @@ export interface DisclosureSummaryRenderFunctionProps extends QAProps {
...
         defaultSummary: React.ReactElement,
     ) => React.ReactElement;
 }
 
-export function DisclosureSummary({children: renderFunction, qa}: DisclosureSummaryProps) {
+export function DisclosureSummary({
+    children: renderFunction,
+    qa,
+    justifyContent = 'start',
+    width = 'auto',
```
Raubzeug, line 58: Defaulting these values here adds `_width_auto` and `_justify-content_start` classes (and explicit CSS declarations) to every existing `Disclosure`, including the implicit summary path. That changes the DOM/CSS contract and may override consumer styles, while the browser’s natural flex defaults already provide the same `auto`/`start` behavior. Could we keep both values `undefined` when omitted and emit modifier classes only for explicit props? The omitted-props test should then assert that these new classes are absent.

PR #2755, opened 2026-07-16
feat(Disclosure): add CollapsedDetails property

GermanVor:
## Summary by Sourcery

Introduce a Disclosure.CollapsedDetails subcomponent that shows overlay preview content only while the disclosure is collapsed, with supporting styling, types, stories, tests, and documentation.

New Features:
- Add Disclosure.CollapsedDetails to render content that is visible only when the disclosure is collapsed and hidden when expanded.
- Export DisclosureCollapsedDetailsProps and a new DisclosureQa.COLLAPSED_DETAILS identifier for testing and QA integration.

Enhancements:
- Update Disclosure layout and animations to support an overlaid collapsed-content area without affecting the collapsed height.
- Add a dedicated Storybook example demonstrating usage of Disclosure.CollapsedDetails.

Documentation:
- Document the Disclosure.CollapsedDetails component and its props in both English and Russian READMEs, including usage examples.

Tests:
- Add tests covering collapsed details visibility in collapsed/expanded states, toggle behavior on user interaction, and enforcement of a single CollapsedDetails instance per Disclosure.


src/components/Disclosure/Disclosure.tsx
```diff
@@ -111,6 +124,14 @@ function prepareChildren(children: React.ReactNode, {disclosureQa}: PrepareParam
             summary = item;
             continue;
         }
+        const isDisclosureCollapsedDetails = isDisclosureCollapsedDetailsComponent(item);
```
Raubzeug, line 127: Could we handle an explicit `<Disclosure.Details>` as a first-class child here as well? In the intended composition:

```tsx
<Disclosure>
  <Disclosure.CollapsedDetails>Preview</Disclosure.CollapsedDetails>
  <Disclosure.Details>Full view</Disclosure.Details>
</Disclosure>
```

the explicit `Disclosure.Details` currently falls through to `content` and is then wrapped in another `DisclosureDetails` below. This produces two transition containers and two `role="region"` elements with the same `ariaControls` ID; with `CollapsedDetails`, both content layers also receive the transition classes.

Could we preserve an explicit `Disclosure.Details` instance (and define the behavior when it is mixed with raw children) and add a regression test using the exact composition above?

src/components/Disclosure/Disclosure.scss
```diff
@@ -50,12 +52,43 @@ $gap: 8px;
...
+    &__collapsed-content#{$block}_exit_active {
+        display: block;
+        animation-name: #{variables.$ns}disclosure-collapsed;
+        animation-duration: $collapse-animation-duration;
+        opacity: 0;
+    }
+    // Inside the shared body wrapper the disappearing content is taken out of the
+    // flow and overlaid on top of the appearing one, so their heights don't briefly
+    // sum up and make the layout jump. At rest, both stay in the normal flow.
+    &__body &__content#{$block}_exit_active,
```
Raubzeug, line 87: On collapse, `__content` is the exiting element, so this rule removes the full details from normal flow immediately. If the expanded details are taller than the collapsed preview, `__body` shrinks to the preview height and the following sibling moves up while the outgoing details are still painted for 100 ms, causing them to overlap.

This is the layout jump/overlap that the new API is intended to avoid. Could we instead keep the two direct children in the same CSS Grid area so the track uses the larger height while both are present, or otherwise preserve/animate the `__body` height during the transition?

It would also be useful to cover this with an interaction component test that collapses details taller than the preview and checks the position of the following sibling during the transition; the current jsdom tests only assert class changes.

src/components/Disclosure/DisclosureContent/DisclosureContent.tsx
```diff
@@ -0,0 +1,45 @@
...
+    ...htmlProps
+}: DisclosureContentProps) {
+    const containerRef = React.useRef<HTMLDivElement>(null);
+
+    return (
+        <CSSTransition
+            nodeRef={containerRef}
+            in={visible}
+            addEndListener={(done) => containerRef.current?.addEventListener('animationend', done)}
+            classNames={getCSSTransitionClassNames(b)}
```
Raubzeug, line 33: The earlier Sourcery thread was marked as addressed after this logic moved out of `DisclosureCollapsedDetails`, but the listener lifecycle is unchanged here. Every transition adds another `animationend` listener, and with the default `keepMounted={true}` the same DOM node remains mounted, so completed and cancelled callbacks stay attached and are invoked by later animation events.

Could we register a removable one-shot handler for each transition (ideally filtering `event.target === containerRef.current` so animated descendants cannot complete it), or use another completion mechanism that provides cleanup?

src/components/Disclosure/Disclosure.tsx
```diff
@@ -19,6 +20,7 @@ export type DisclosureArrowPosition = 'left' | 'right' | 'start' | 'end';
 export interface DisclosureComposition {
     Summary: typeof DisclosureSummary;
     Details: typeof DisclosureDetails;
+    CollapsedDetails: typeof DisclosureCollapsedDetails;
```
Raubzeug, line 23: Could we reconsider the public name before merging? `CollapsedDetails` is somewhat difficult to interpret because “details” normally refers to the panel shown in the expanded state.

Would `Disclosure.Preview` describe the intended role more clearly?

```tsx
<Disclosure>
  <Disclosure.Preview>Short view</Disclosure.Preview>
  <Disclosure.Details>Full view</Disclosure.Details>
</Disclosure>
```

If the intended use case is instead a truncated preview of the same content, that may be better represented by a separate `Spoiler` component or a `collapsedHeight` API.

PR #2757, opened 2026-07-20
fix(Tabs): correctly render custom components + migration guide

amje:
## Summary by Sourcery

Update Tabs and related components to better support custom tab components, improve collapsed/overflow behavior, and centralize link security handling, while adding a migration guide from legacy tabs APIs.

New Features:
- Support rendering Tabs with custom components via a new TabInner wrapper and a dedicated story example.
- Introduce a shared utility to compute safe link rel values for components that render anchors.
- Add a migration guide documenting how to move from legacy Tabs and AdaptiveTabs to the new TabList/Tab/TabPanel API.

Enhancements:
- Refine Tab styling for icons, counters, and menu-item variants, including focused/hovered and disabled states.
- Improve TabList collapse behavior by simplifying measurement logic and refining collapsed trigger styling, including handling expanded state and ellipsis.
- Adjust tab list gap calculation to better handle cases where margins cannot be determined without altering existing layout.
- Align MenuItem and BreadcrumbsItem internals with the new link rel helper and menu item prop handling.
- Tighten Tab hook typing and className generation to support a dedicated menu-item BEM block and internal isMenuItem usage.

Tests:
- Extend TabList stories to demonstrate content overflow in a resizable container and usage with a custom tab component.

Chores:
- Remove obsolete isMenuItem from public Tab types and introduce a dedicated tab-menu-item BEM block for internal styling.

### DIFF
--- a/src/components/lab/Menu/MenuItem.tsx
+++ b/src/components/lab/Menu/MenuItem.tsx
@@ -11,6 +11,8 @@ import {useDirection} from '../../theme';
 import {block} from '../../utils/cn';
 import {getLinkRelWithFallback} from '../../utils/getLinkRelWithFallback';
 import {mergeProps} from '../../utils/mergeProps';
+import type {PolymorphicOverloadProps} from '../../utils/polymorphic';
+import {isPolymorphicComponentProps, isPolymorphicLinkProps} from '../../utils/polymorphic';
 import {ListItemView} from '../ListItemView/ListItemView';
 import type {ListItemViewProps} from '../ListItemView/ListItemView';
 
@@ -31,13 +33,16 @@ import './MenuItem.scss';
 function isMenuItemComponentProps<T extends MenuItemComponentElementType>(
     p: MenuItemProps<T>,
 ): p is MenuItemComponentProps<Exclude<T, undefined>> {
-    return p.component !== undefined;
+    return isPolymorphicComponentProps<
+        MenuItemProps<T>,
+        MenuItemComponentProps<Exclude<T, undefined>>
+    >(p);
 }
 
 function isMenuItemLinkProps<T extends MenuItemComponentElementType>(
     p: MenuItemProps<T>,
 ): p is MenuItemLinkProps {
-    return p.href !== undefined;
+    return isPolymorphicLinkProps<MenuItemProps<T>, MenuItemLinkProps>(p);
 }
 
 const b = block('lab-menu-item');
@@ -138,8 +143,9 @@ export const MenuItem = React.forwardRef(
         };
 
         let component: React.ElementType;
-        let componentProps: React.ComponentProps<typeof component>;
+        let componentProps: React.ComponentPropsWithoutRef<typeof component>;
         const commonComponentProps = menuContext.getItemProps({
+            'data-qa': qa,
             ...restComponentProps,
             role: 'menuitem',
             tabIndex,
@@ -150,7 +156,6 @@ export const MenuItem = React.forwardRef(
                 },
                 className,
             ),
-            'data-qa': qa,
             onClick: handleClick,
             onFocus: handleFocus,
             onPointerEnter: handlePointerEnter,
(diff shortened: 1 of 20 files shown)


amje:
@sourcery-ai dismiss

amje:
@sourcery-ai review

src/components/lab/Menu/MenuItem.tsx
```diff
@@ -148,14 +151,11 @@ export const MenuItem = React.forwardRef(
                 className,
             ),
             'data-qa': qa,
```
korvin89: 🤖 **AI generated**

**should-fix** — Heads up: this reorder looks like it quietly drops `data-qa` for tabs collapsed into the overflow menu. A `Tab` passes its `qa` through `useTab` as a `data-qa` key, which lands in `restComponentProps` — but here `'data-qa': qa` (and `MenuItem`'s own `qa` is `undefined` in that case) sits *after* the `...restComponentProps` spread, so it overwrites the real value with `undefined`. Before, `'data-qa': qa` was the base and `getItemProps({...restComponentProps})` was spread on top, so the tab's `data-qa` survived.

There's no test around this, so it'd be easy to miss. Something like this keeps both cases working:

```ts
'data-qa': qa ?? (restComponentProps as {['data-qa']?: string})['data-qa'],
```

src/components/tabs/Tab.tsx
```diff
@@ -36,15 +45,29 @@ export const Tab = React.forwardRef<HTMLAnchorElement | HTMLButtonElement, TabPr
...
+                    {...tabProps}
+                    ref={ref as React.Ref<T extends string ? React.ComponentRef<T> : T>}
+                    component={props.component}
+                >
+                    {content}
+                </MenuItemComponent>
+            );
+        }
+
+        return React.createElement(props.component, {...tabProps, ref}, content);
```
korvin89, line 67: 🤖 **AI generated**

**should-fix** — This is the core fix of the PR — a custom `component` tab now gets the full `TabContent` (icon/counter/label) instead of just raw children — but I couldn't find a test that actually exercises it. `TestTabListWithCustomTabs` uses `withCustomChildren`, which is a different path (custom *children* already worked), and the new `CustomComponent` story isn't part of the visual snapshots. Would you consider a small unit test — e.g. `<Tab component={Div} icon counter label>text</Tab>` asserting the icon/counter/label/text render inside the custom element? That'd lock in exactly the behavior this PR restores.

src/components/tabs/types.ts
```diff
@@ -35,8 +35,6 @@ interface TabCommonProps extends QAProps, DOMProps {
     };
     disabled?: boolean;
     children?: React.ReactNode;
```
korvin89, line 38: 🤖 **AI generated**

**question** — Removing `isMenuItem` from `TabCommonProps` also drops it from the exported `TabProps`/`TabButtonProps`/`TabLinkProps`/`TabComponentProps`, so `<Tab isMenuItem>` no longer type-checks. It was clearly a service prop (the JSDoc even said "Tab was collapsed and rendered in Menu"), so the practical impact is probably tiny — just want to confirm it's fine to ship under a `fix:` and doesn't need a changelog/semver note for anyone who happened to reference it.

src/components/tabs/Tab.tsx
```diff
@@ -90,4 +113,20 @@ export const Tab = React.forwardRef<HTMLAnchorElement | HTMLButtonElement, TabPr
     );
 });
 
+TabInner.displayName = 'TabInner';
+
+export const Tab = React.forwardRef<HTMLAnchorElement | HTMLButtonElement, TabProps>(
```
korvin89, line 119: 🤖 **AI generated**

**nit** — The type assertions here are doing a lot of heavy lifting — the public `Tab` cast (`... as (<T …>) => React.ReactElement) & {displayName: string}`) and the `MenuItem as unknown as React.ForwardRefExoticComponent<…>` double-cast on line 49. It works and it's internal, so no objection to the behavior — it's just tricky to read and maintain. If it's low-effort, pulling this into a named type alias / small helper would make it friendlier for the next person. (sourcery flagged the same spot.)

src/components/tabs/hooks/useTab.ts
```diff
@@ -77,14 +79,16 @@ export function useTab<T extends TabComponentElementType>(tabProps: TabProps<T>)
 
     return {
         ...htmlRest,
-        ...(tabProps.isMenuItem ? {} : tabRoleProps),
+        ...(tabProps.isMenuItem ? {} : {...tabRoleProps}),
```
korvin89: 🤖 **AI generated**

**nit** — Tiny one: `...(tabProps.isMenuItem ? {} : {...tabRoleProps})` has an extra spread that isn't needed — `: tabRoleProps` does the same thing.

korvin89 (commented):
🤖 **AI generated**

Left a few inline comments. Severity legend for the tags used:

- **blocker** — must be fixed before merge (correctness / security / data loss).
- **should-fix** — a real problem worth fixing before or right after merge.
- **nit** — minor, author's discretion.
- **question** — I need more info to judge; not a claimed defect.

src/components/tabs/types.ts
amje: `isMenuItem` was never intended to be a public API, removing with fix is OK

src/components/lab/Menu/MenuItem.tsx
amje: Fixed

src/components/tabs/Tab.tsx
amje: Added test

src/components/tabs/hooks/useTab.ts
amje: Fixed

src/components/tabs/Tab.tsx
amje: Added type alias and common utils for polymorphic pattern

korvin89 (approved):

PR #2760, opened 2026-07-22
fix(Select): fix availability of error popover with placement inside

### LINKED ISSUE #2759: [Select] Error tooltip does not open with errorPlacement="inside"
### Describe the bug

When Select is invalid and errorPlacement="inside" is used, hovering the error icon does not open the error popover.

The main control button has a ::before pseudo-element with position: absolute and inset: 0. It is painted above .g-select-control__error-icon and intercepts pointer events.

Expected: hovering the error icon displays the error message.
Actual: the hover is intercepted by button[role="combobox"] and the popover does not open.

Setting z-index: 1 on .g-select-control__error-icon fixes the issue. The neighboring .g-select-clear element already uses the same z-index.

**Impact: After deploying our application release 2.388.2, validation error tooltips in Select controls stopped opening. Users can see the error icon but cannot access the validation message. This issue is currently blocking our release.**

### Reproduction

https://preview.gravity-ui.com/uikit/iframe.html?id=components-inputs-select--with-inside-error&viewMode=story

### Steps to reproduce

1. Open the WithInsideError Select story.
2. Hover the error icon inside the Select control.
3. Observe that the error popover is not displayed.

### Environment

@gravity-ui/uikit: 7.46.0
OS: macOS Sequoia 15.6.1
Browser: Yandex Browser 26.3.5.866

Also reproducible in the official Gravity UI Storybook.

### Validations

benax-se:
Closes  #2759

## Summary by Sourcery

Ensure Select shows its inside error popover when hovering the error icon by adjusting stacking and adding a regression test.

Bug Fixes:
- Fix inside-placed Select error icon so its popover opens on hover instead of being blocked by the control’s decorative layer.

Enhancements:
- Document the design rationale and scope for the Select inside error popover stacking fix in a new spec file.

Tests:
- Add a Playwright visual regression test that verifies the inside error popover becomes visible when hovering the Select error icon.

### DIFF
--- a/src/components/Select/__tests__/Select.error.test.tsx
+++ b/src/components/Select/__tests__/Select.error.test.tsx
@@ -1,3 +1,5 @@
+import userEvent from '@testing-library/user-event';
+
 import {render, screen} from '../../../../test-utils/utils';
 import {CONTROL_ERROR_MESSAGE_QA} from '../../controls/utils';
 import {Select} from '../Select';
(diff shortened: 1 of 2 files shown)


src/components/Select/__tests__/Select.visual.test.tsx
```diff
@@ -232,6 +233,20 @@ test.describe('Select', {tag: '@Select'}, () => {
         });
     });
 
+    test('opens inside error popover on hover', async ({mount, page}) => {
```
amje: This test should be placed inside the Select.error.test.tsx. Current file is for visual checks with screenshots

amje (changes requested):

src/components/Select/__tests__/Select.visual.test.tsx
benax-se: Moved to Select.error.test.tsx and rewrited it to unit test

amje (approved):

korvin89 (approved):

PR #2761, opened 2026-07-23
fix(Tabs): optimize styles and internal markup

amje:
## Summary by Sourcery

Rename and reimplement the Tabs overflow "More" control while aligning tab list layout and focus styles with a gap-based, menu-friendly design.

New Features:
- Introduce the TabMore component as the overflow "More" control for collapsed tabs with dedicated styling.

Enhancements:
- Simplify Tab markup by using native button structure and centralizing text color and focus styles at the root element.
- Switch tab list spacing from per-item margins to container gap, and update collapse logic to work with the new layout.
- Adjust tab typography and max-width behavior to improve truncation and alignment of titles and counters.
- Refine color states for tabs and the "More" control, including hover, focus, active, disabled, and expanded states.

Documentation:
- Update the Tabs migration guide to reference TabMore and reflect the new overflow control behavior.


src/components/tabs/TabList.scss
```diff
@@ -19,8 +18,7 @@ $tabListCollapseItemBlock: '.#{variables.$ns}tab-list-collapse-item';
             --_--item-border-width: 2px;
 
             #{$tabBlock}__title,
-            #{$tabBlock}__counter,
-            #{$tabListCollapseItemBlock} {
+            #{$tabBlock}__counter {
```
korvin89: 🤖 **AI generated**

**should-fix** — The collapsed "More" trigger's count looks like it loses its size typography here. These per-size selectors now match only `.g-tab__title` and `.g-tab__counter`, but the count is rendered as `.g-tab-more__count`. Since `.g-tab` pulls in `button-reset()` (`font-size: inherit`) and `.g-tab-list` sets no root font-size, the count falls back to the ambient document size instead of `text-body-1/2/subheader-3` — so it renders at a different size than the adjacent "More" label. Previously `.g-tab-list-collapse-item` was part of these groups, which kept it in sync.

Would you consider adding `.g-tab-more__count` (or `.g-tab-more`) to the three per-size groups? For example:

```scss
#{$tabBlock}__title,
#{$tabBlock}__counter,
#{$tabMoreBlock}__count {
    @include mixins.text-body-1();
}
```

src/components/tabs/Tab.scss
```diff
@@ -54,18 +44,15 @@ $blockMenuItem: '.#{variables.$ns}tab-menu-item';
     }
 
     &:focus-visible {
-        #{$block}__content {
-            @include mixins.focus-outline();
-            outline-offset: -2px;
-        }
+        @include mixins.focus-outline();
```
korvin89, line 46: 🤖 **AI generated**

**question** — The focus outline moved from `.g-tab__content` to the root `.g-tab`. Just to confirm the intent here: with `--g-focus-border-radius: 2px` and `outline-offset: -2px`, the effective corner radius was already ~0 on `main` (the 2px radius is fully consumed by the -2px inset), so the ring was effectively square there too — this change doesn't appear to regress the rounding. What it does change is that the ring now wraps the whole `.g-tab` box rather than just the content.

Two small follow-ups: (1) the `border-radius: var(--g-focus-border-radius)` still on `.g-tab__content` (line 26) is now dead — nothing draws it anymore, so it could be dropped; (2) if a rounded focus ring is actually intended, it would need a radius on the root that isn't fully eaten by the `-2px` offset. Is the square, full-box ring the intended look?

src/components/tabs/TabList.scss
amje: Fixed

src/components/tabs/Tab.scss
amje: Dropped redundant radius, square focus ring is ok

korvin89 (approved):

PR #2769, opened 2026-08-05
feat: add new List

korvin89:
## Summary by Sourcery

Introduce a new unstable List component with virtualization, selection, and drag-and-drop support, integrate it with existing Virtualizer and ListItemView, and add comprehensive tests, stories, and Storybook configuration for raw source imports.

New Features:
- Add an unstable List component with keyboard navigation, typeahead, selection modes, drag-and-drop integration hooks, and support for both listbox and grid ARIA roles.
- Introduce a ListVirtualizer wrapper around the Virtualizer to provide configurable row height estimation, measurement, and overscan for virtualized lists.
- Expose the new unstable List API (component, helpers, types, and moveItem utility) from the unstable entry point for external consumption.

Bug Fixes:
- Adjust ListItemView focus and active-state styling to remove stale hover-on-focus behavior and correctly indicate active and selected states, including for disabled rows.
- Fix Virtualizer containment and spacer flex behavior so scroll containers with max-height and flex layouts maintain correct scroll height and overflow behavior.
- Resolve Storybook raw import handling so ?raw TSX/TS files are served as asset/source instead of being transformed by Babel.

Enhancements:
- Improve Virtualizer measurement by correcting height estimation based on measured rows and handling zero-height measurements to avoid visual jumps during drag-and-drop.
- Refine ListItemView DOM props handling by simplifying prop forwarding and aligning onClick typing with standard React mouse event handlers.
- Add dev-time diagnostics and contract checks for List drag-and-drop adapters, focus owners, and grid keyboard behavior to ensure stable refs and single-tab-stop compliance.

Documentation:
- Add a comprehensive README for the unstable List component documenting usage, sizing, item getters, renderItem patterns, selection, virtualization, drag-and-drop integration, focus ownership, accessibility, and public types.
- Create Storybook stories and MDX docs for the List component, including interactive examples for sizes, selection modes, virtualization, drag-and-drop with pragmatic-dnd, dnd-kit, and hello-pangea/dnd, and focus-owner scenarios.

Tests:
- Add extensive unit tests for the List component covering ARIA structure, keyboard navigation, hover activation, item actions, selection behavior (including ranges and controlled modes), typeahead, drag-and-drop state wiring, text selection suppression, virtualization behavior, and focus-owner integration.
- Add unit tests for ListItemView to validate DOM prop pass-through, click handling, disabled behavior, and state modifiers in the new activation modality model.
- Add tests ensuring List grid-role behavior, including cell navigation with arrow keys, tabbable-children diagnostics, and ARIA numbering under virtualization.

Chores:
- Add drag-and-drop library dev dependencies (@atlaskit/pragmatic-drag-and-drop, @dnd-kit/*) to support List drag-and-drop examples and tests.
- Extend Storybook webpack config with helpers to exclude ?raw imports from script rules and register a dedicated asset/source rule for raw source loading.


src/components/lab/List/SectionHeader.scss
```diff
@@ -0,0 +1,24 @@
...
+    color: var(--g-color-text-secondary);
+
+    @include mixins.text-caption-2;
+
+    &_size_l,
+    &_size_xl {
+        @include mixins.text-caption-1;
+    }
+
+    &:not(:first-child) {
```
amje: The `&:not(:first-child)` selector works in a non-virtualized `List`, but the `Virtualizer` places every header inside its own wrapper, making it always the wrapper’s first child.

src/components/lab/Virtualizer/Virtualizer.tsx
```diff
@@ -77,15 +90,89 @@ export function Virtualizer({
...
+                if (size === 0 && lastMeasured !== undefined) {
+                    return lastMeasured;
+                }
+                // A stable dataset can hold at most `count` distinct keys; more
+                // than that means the data shrank or got replaced and the cache
+                // now carries orphaned keys — their sizes skew the ratio and
+                // leak memory, so drop the whole cache and rebuild it from the
+                // current window. Append-only growth (infinite scroll) keeps
+                // size <= count and is preserved.
+                if (correction.sizes.size > count) {
```
amje, line 166: The cache retains measurements from previous items and is cleared only when `sizes.size > count`. When a large list is replaced with another list having the same `count`, stale measurements may continue to distort the correction ratio, total scroll height, and row positions. Similarly, changing `size` or `estimateItemSize` continues to use the old `estimatedTotal`. The cache should either be invalidated or its stored estimates updated.

src/styles.d.ts
```diff
@@ -0,0 +1,7 @@
+declare module '*.scss';
```
amje: Why is this needed?

src/components/lab/List/__stories__/List.stories.tsx
```diff
@@ -0,0 +1,668 @@
...
+            control: 'text',
+            description: 'The accessible name of the list (or use `aria-labelledby`)',
+        },
+    },
+};
+
+// The four row sizes. The size is passed to the row view (its height, paddings
+// and typography) and is the source of the default height estimate of the
+// virtualization layer
+const sizes = ['s', 'm', 'l', 'xl'] as const;
```
amje, line 146: The size grid is inconsistent with other components, Button has 24, 28, 36, 44. LabMenu has 24, 28, 32, 36 which uses the same ListItemView, while this List has 24, 28, 34, 42

src/components/lab/List/List.scss
```diff
@@ -0,0 +1,46 @@
+@use '../../variables';
+
+$block: '.#{variables.$ns}list-v2';
```
amje: Maybe just `g-lab-list`?

src/components/lab/List/SectionHeader.scss
korvin89: 🤖 AI generated

Good catch — under virtualization every header is the first child of its own positioned wrapper, so the selector never matched and the flat and virtualized renders disagreed. Fixed in 93af068d: the spacing is driven by data (the core marks the header that opens the list with `data-first-row`) and uses padding instead of margin, so the wrapper measurement counts it.

src/components/lab/Virtualizer/Virtualizer.tsx
korvin89: 🤖 AI generated

Agreed, with one nuance on the first scenario: replacing the list with another one of the same count but new keys mostly self-heals — the first measured new key overflows `sizes.size > count` and the whole cache is rebuilt. What did not self-heal was a change of `size`/`estimateItemSize`: the accumulated `estimatedTotal` belonged to the old estimates and kept skewing the unmeasured tail. Fixed in 459b29ee — the list now passes a `getItemSize` whose identity is stable while the estimate configuration is, and the virtualizer restarts the correction when the identity changes. The remaining known limitation is rows that keep their keys but change height outside the window: they re-measure once scrolled into view.

src/styles.d.ts
korvin89: 🤖 AI generated

The `*?raw` declaration types the raw-source imports used by the docs page and the story Code panels (`List.stories.tsx`, `ListDragAndDropLibraries.stories.tsx` and `Docs.mdx` import example sources as text) — without it `tsc` fails with TS2307 on exactly those imports. The `*.scss`/`*.css` lines turned out to be unnecessary — the typecheck passes without them — so the file is trimmed to just the `*?raw` declaration in f4491d23.

src/components/lab/List/__stories__/List.stories.tsx
korvin89: 🤖 AI generated

Confirmed — the mechanics: the content line-height is a fixed 18px while the block paddings grow with the size (2/4/8/12px), so at l/xl the paddings overflowed the declared min-heights (32/36) and a single-line row rendered 34/42px tall; s/m were saved by min-height centering.

Fixed in fe0e8322 by shrinking the paddings (l 8→6px, xl 12→8px): single-line rows now land exactly on the declared 24/28/32/36 grid, and rows with a description hit min-height + 16 precisely (48/52). This also makes the virtualization height estimates of lab/List exact.

On which grid to target: aligned to 24/28/32/36 rather than Button's 24/28/36/44 deliberately — it is the grid the Select options live on today (`SIZE_TO_ITEM_HEIGHT`: l 32, xl 36), so migrating Select onto this list will not change the popup geometry. If the design direction is a single control grid instead, switching is a two-line change (min-height l→36 / xl→44) at the cost of taller Select popups — happy to discuss.

src/components/lab/List/List.scss
korvin89: 🤖 AI generated

Done in 0c4f80fb: the root block is `g-lab-list` and the section header is `g-lab-list-section-header`. Checked for collisions first — `g-lab-list` was unused in the repo, and the existing `g-lab-list-item-view` is a separate block, so the selectors do not intersect.

amje (approved):

PR #2772, opened 2026-08-06
feat(ColorPicker): replaced open state with shared useOpenState

### LINKED ISSUE #2771: ColorPicker: use shared onOpenChange signature
### Description

replace onOpenChange functionality with shared hook

### Additional context

_No response_

### Validations

- [x] Check that there isn't [already an issue](https://github.com/gravity-ui/uikit/issues) that pursue the same goal to avoid creating a duplicate.

skoptenco:
Closes [2771](https://github.com/gravity-ui/uikit/issues/2771)

Added useOpenState to ColorPicker component instead of useControlledState

## Summary by Sourcery

Update ColorPicker to use the shared open-state hook and add test coverage for its open/close behavior.

New Features:
- Expose a qa prop on ColorPicker to support data-qa test attributes.

Enhancements:
- Reuse shared useOpenState behavior in ColorPicker by aligning its props with UseOpenProps.

Tests:
- Add unit tests validating ColorPicker popup open/close behavior, including defaultOpen, onOpenChange, and onClose callbacks.

### DIFF
--- a/src/components/lab/ColorPicker/ColorPicker.tsx
+++ b/src/components/lab/ColorPicker/ColorPicker.tsx
@@ -6,7 +6,7 @@ import debounce from 'lodash/debounce';
 
 import {useControlledState} from '../../../hooks/useControlledState';
 import {Popup} from '../../Popup';
-import type {PopupPlacement} from '../../Popup';
+import type {PopupPlacement, PopupProps} from '../../Popup';
 import {Select} from '../../Select';
 
 import {ColorDisplay, ColorPointer, HexInput, RgbInputs} from './components';
@@ -51,7 +51,7 @@ export interface ColorPickerProps {
     /*
      * Open popup handler
      */
-    onOpenChange?: (open: boolean) => void;
+    onOpenChange?: PopupProps['onOpenChange'];
     /*
      * Enables alpha channel support for HEXA/RGBA mode and transparency slider.
      */
--- a/src/components/lab/ColorPicker/__tests__/ColorPicker.test.tsx
+++ b/src/components/lab/ColorPicker/__tests__/ColorPicker.test.tsx
@@ -1,0 +1,47 @@
+import userEvent from '@testing-library/user-event';
+
+import {render, screen} from '../../../../../test-utils/utils';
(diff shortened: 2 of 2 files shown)


korvin89 (changes requested):
- The main idea was to give the `onOpenChange` callback [this](https://github.com/gravity-ui/uikit/blob/a6bb0f429db390d507c7744840a3901b4b8903b6/src/components/Popup/Popup.tsx#L67) signature; in fact, the changes leave it as it is.
- We don’t know the fate of `useOpenState`, so it’s better to leave this place as it is and do not use this pattern.
-  It’s better to add the `qa` attribute in a separate task, as it’s not yet clear exactly where it needs to be applied. The prop will probably need to be named in a more unique way. I suggest waiting for the actual use case.
- There is exactly one test needed — to verify that the callback is called with the required number of arguments.

src/components/lab/ColorPicker/ColorPicker.tsx
```diff
@@ -51,7 +52,7 @@ export interface ColorPickerProps {
     /*
      * Open popup handler
      */
-    onOpenChange?: (open: boolean) => void;
+    onOpenChange?: (open: boolean, event?: Event, reason?: OpenChangeReason) => void;
```
korvin89: Let’s use [this](https://github.com/gravity-ui/uikit/blob/a6bb0f429db390d507c7744840a3901b4b8903b6/src/components/Popup/Popup.tsx#L67) type here; only Popup uses floating

skoptenco: > Let’s use [this](https://github.com/gravity-ui/uikit/blob/a6bb0f429db390d507c7744840a3901b4b8903b6/src/components/Popup/Popup.tsx#L67) type here; only Popup uses floating

Do you mean PopupProps['onOpenChange']?

korvin89 (approved):

src/components/lab/ColorPicker/ColorPicker.tsx
korvin89: Yep

Estasie (approved):

PR #2782, opened 2026-08-11
chore: self-host star history chart generation

vvtimofeev:
## Why

GitHub [restricted the stargazers API](https://star-history.com/blog/github-stargazer-api-restriction) (June 2026) to repository admins and collaborators. The live star-history.com embed in our README now renders a "GitHub restricted access to star data" placeholder instead of the chart — for every repository, globally.

## What

Self-host the chart generation using the repository's own credentials:

- **`.github/scripts/star-history.mjs`** — dependency-free Node script: fetches star timestamps via GraphQL (`stargazers` with `starredAt`, paginated), aggregates per day, and renders static light/dark SVG charts. Fails loudly (non-zero exit) if the token has no access to star data, so a broken chart never gets published.
- **`.github/workflows/star-history.yml`** — weekly cron + manual dispatch: generates the charts and force-pushes them as a single-commit orphan branch `star-history-data`, keeping `main` history clean.
- **README.md / README-ru.md** — the `<picture>` block now points to `raw.githubusercontent.com/gravity-ui/uikit/star-history-data/...` (light/dark structure preserved).

Chart colors are taken from the Gravity UI amber ramp (`#ffbe5c` brand yellow in the area fill; line `#bd8e4b` light / `#bc8a2e` dark) and pass contrast/CVD validation on both GitHub backgrounds.

Verified locally against live data (1,029 stars fetched via a collaborator token — collaborator access works under the new API restriction).

| Light | Dark |
|---|---|
| renders on `#ffffff` | renders on `#0d1117` |

## After merge

Until the `star-history-data` branch exists, the README image will 404. **Run the "Star History" workflow once manually** (workflow_dispatch) — it creates the branch and the chart goes live. The first run also confirms that the Actions `GITHUB_TOKEN` passes the stargazers restriction.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Self-host generation and publishing of the repository's GitHub star history chart and update READMEs to reference the new static SVG assets.

Enhancements:
- Add a Node-based script to fetch stargazer data via GitHub GraphQL and render light/dark SVG star history charts.
- Introduce a scheduled GitHub Actions workflow that generates charts and publishes them to a dedicated orphan branch for clean main history.

Documentation:
- Update English and Russian READMEs to load the star history chart from the repository-hosted SVG assets instead of the external star-history.com service.

### DIFF
--- a/.github/scripts/star-history.mjs
+++ b/.github/scripts/star-history.mjs
@@ -96,6 +96,7 @@ async function fetchStarDates(repo, token) {
             headers: {
                 Authorization: `bearer ${token}`,
                 'Content-Type': 'application/json',
+                'User-Agent': 'gravity-ui-star-history',
             },
             body: JSON.stringify({query, variables: {owner, name, cursor}}),
         });
@@ -198,8 +199,8 @@ function xTicks(minTime, maxTime) {
 
 const formatCount = (n) => new Intl.NumberFormat('en-US').format(n);
 
-function renderSvg({repo, series, total, updatedAt, theme}) {
-    const t = THEMES[theme];
+/** Turns the series into plot geometry: scales, path coordinates and the end-point anchor. */
+function preparePlot({series, total, updatedAt}) {
     const plotW = WIDTH - MARGIN.left - MARGIN.right;
     const plotH = HEIGHT - MARGIN.top - MARGIN.bottom;
     const baselineY = MARGIN.top + plotH;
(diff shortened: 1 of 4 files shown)


vvtimofeev:
Thanks for the review — addressed in a442d497 and ef225f8f.

**Applied:**

- **User-Agent header** — GitHub does expect an explicit one and may throttle requests without it. Now sends `gravity-ui-star-history`.
- **Full stack on failure** — `console.error(error?.stack ?? error)`. Diagnosing a failed nightly run from a single message line would have been painful.
- **`github.token` instead of `secrets.GITHUB_TOKEN`** — same value, no confusion with repository secrets.
- **`renderSvg` complexity** — split into `preparePlot` (scales, path coordinates, end-point anchor) plus `renderGrid` / `renderXLabels`, leaving `renderSvg` as composition only. Verified the rendered output is unchanged by diffing against a pre-refactor baseline; the only delta is one label coordinate now printing as `704.0` instead of `704`, which makes coordinate formatting consistent across the file.

**Not applied — pinning `node-version` instead of `.nvmrc`:** the concern was that global `fetch` may be missing on older runtimes, but `.nvmrc` in this repo pins `20.19`, which is a stricter floor than `node-version: 20` would be, and `fetch` has been stable since Node 18. Keeping `.nvmrc` also keeps a single source of truth for the Node version.

**Separately**, CI surfaced something the review didn't: this repo's ESLint config covers `.github/scripts/` too, where `no-console` only permits `info`/`warn`/`error`, and any `/** */` block triggers `jsdoc/require-param`. Three `console.log` calls were failing the Verify Files job (in the original commit as well). Fixed in ef225f8f — all checks are green now.

Issue #2773, opened 2026-08-07
Sheet: component refactoring


skoptenco:
### Description

- [x] Improve test coverage for the main functionality of the Sheet component. The goal is to capture the current behavior and minimize the risk of regressions during further changes.
- [x] The refactoring should also address the existing long-standing TODOs related to splitting the component into smaller subcomponents
- [x] Migrate the SheetContent from a class-based component to a functional component
- [ ] Migrate to useFloating
- [ ] Rework swipe functionality (think about third-party lib or maybe some DIY)

### Validations

- [x] Check that there isn't [already an issue](https://github.com/gravity-ui/uikit/issues) that pursue the same goal to avoid creating a duplicate.

PR #2791, opened 2026-08-25
feat(Sheet): migrate to useFloating

skoptenco:
## Summary

Related to [2773](https://github.com/gravity-ui/uikit/issues/2773)

- migrate `Sheet` to Floating UI using `useFloating`, `FloatingTree`, `FloatingNode`, and the shared floating transition lifecycle
- add `onOpenChange` as the unified API for closing via:
  - outside press
  - Escape
  - swipe
  - hash navigation
- preserve controlled behavior: dismissal remains a request until the parent updates `visible`
- keep deprecated `onClose` backward-compatible with its existing post-transition timing
- close only the topmost `Sheet` on Escape, including sibling sheets with independent `FloatingTree` instances
- keep accepted exits terminal and correctly restore interrupted or cancelled swipe gestures
- support reopening while an exit transition is still in progress
- keep mobile hash state synchronized across swipe, external, and navigation-driven dismissal
- prevent duplicate navigation callbacks during Android hash cleanup
- remove redundant test scenarios and strengthen coverage for controlled dismissal, lifecycle, gestures, and hash navigation


src/components/Sheet/hooks/useSheetDismiss.ts
```diff
@@ -0,0 +1,51 @@
...
+    const requestDismiss = React.useCallback(
+        (request: SheetDismissRequest) => {
+            if (!requestedOpen) {
+                return;
+            }
+
+            if (onOpenChange) {
+                onOpenChange(false, request.event, request.reason);
+            } else {
+                setLegacyDismissed(true);
```
mournfulCoroner, line 44: **AI-generated:** The `immediate` flag is included in `SheetDismissRequest` and is set by `useSwipe` when the drag reaches the full sheet height, but it is ignored here. Before this refactoring, that path called `hideSheet()` synchronously. Now a fully off-screen sheet remains mounted, keeps the body scroll locked, and delays the legacy `onClose` callback for another 300 ms.

Could we preserve the immediate path in the new presence lifecycle? This regression test demonstrates the previous behavior and currently fails:

```tsx
test('finishes a legacy full-height swipe immediately', () => {
    const onClose = jest.fn();

    render(
        <Sheet visible onClose={onClose}>
            Content
        </Sheet>,
    );

    finishTransition();

    swipe(screen.getByTestId(SheetQa.SWIPE_AREA), {
        from: TOUCH_START_POINT,
        to: TOUCH_START_POINT + SHEET_HEIGHT,
    });

    expect(onClose).toHaveBeenCalledTimes(1);
    expect(screen.queryByRole('dialog')).not.toBeInTheDocument();
    expect(document.body.style.overflow).toBe('');
});
```

src/components/Sheet/SheetContent.tsx
```diff
@@ -71,35 +79,22 @@ export function SheetContent(props: SheetContentProps) {
     const history = useHistory();
     const location = useLocation();
 
-    const veilRef = React.useRef<HTMLDivElement>(null);
     const sheetRef = React.useRef<HTMLDivElement>(null);
     const sheetTopRef = React.useRef<HTMLDivElement>(null);
     const sheetMarginBoxRef = React.useRef<HTMLDivElement>(null);
     const sheetScrollContainerRef = React.useRef<HTMLDivElement>(null);
+    const handleSheetRef = useForkRef(sheetRef, floatingRef);
```
mournfulCoroner, line 86: **AI-generated:** This new `useForkRef(sheetRef, floatingRef)` usage produces a React warning on every `Sheet` mount under React 18:

```text
Warning: Unexpected return value from a callback ref in div.
A callback ref should not return a function.
```

I reproduced it with the existing `Sheet` tests on the repository's React 18.3.1 setup. `mergeRefs` returns a cleanup function from the callback ref, which is supported by React 19 but is treated as an error in earlier supported React versions.

Could we use a ref merger that returns `void` on React <19, or otherwise resolve the cross-version behavior before introducing it to `Sheet`? This can also break consumer test suites that fail on unexpected `console.error` calls.

src/components/Sheet/README.md
```diff
@@ -35,7 +35,7 @@ const SheetExample = () => {
 ## Properties
 
 | Name                        | Description                                                                                                                                                      |     Type      |     Default     |
-| :-------------------------- | :--------------------------------------------------------------------------------------------------------------------------------------------------------------- | :-----------: | :-------------: |
+| :-------------------------- | :--------------------------------------------------------------------------------------------------------------------------------------------------------------- | :-----------: | :-------------: | --- | ---------- | ----------- |
```
mournfulCoroner, line 38: **AI-generated:** The properties table is malformed after this change: the header has four columns while the separator has seven, several rows are duplicated, and the second `maxContentHeightCoefficient` row contains fragments of another property. The same column-count problem is present in `README-ru.md`.

The deprecated `onClose` entry was also removed even though this PR explicitly preserves its backward-compatible behavior. Could we restore a consistent four-column table and keep `onClose` documented as deprecated?

src/components/Sheet/Sheet.tsx
```diff
@@ -2,21 +2,39 @@
...
+import {useLayer} from '../utils/layer-manager';
 
 import {SheetContentContainer} from './SheetContent';
-import {sheetBlock} from './constants';
+import {SHEET_TRANSITION_DURATION_MS, sheetBlock} from './constants';
+import {useSheetDismiss} from './hooks/useSheetDismiss';
 
 import './Sheet.scss';
 
+export type SheetOpenChangeReason = OpenChangeReason | 'swipe' | 'navigation';
```
mournfulCoroner, line 30: **AI-generated:** Could we keep the public dismissal-reason contract narrower and use it as the single source of truth? `OpenChangeReason` also includes reasons such as `hover`, `focus`, `click`, and `ancestor-scroll`, but `Sheet` only emits `escape-key`, `outside-press`, `swipe`, and `navigation`. The exact four-value union is then duplicated separately as `SheetDismissReason`.

Exposing the broader Floating UI type makes consumers handle states that `Sheet` cannot produce and couples the public API to future changes in a dependency type. I suggest moving the four-value union to a shared type module and reusing it in both `SheetProps['onOpenChange']` and `SheetDismissRequest`.

skoptenco: left only emitted dismissal-reasons in SheetOpenChangeReason

src/components/Sheet/README.md
skoptenco: fixed

src/components/Sheet/SheetContent.tsx
skoptenco: replaced useForkRef with useMergeRefs

src/components/Sheet/hooks/useSheetDismiss.ts
mournfulCoroner: 🤖 **AI generated**

🟠 **should-fix**

This issue is still present on the current head (`ff3b7202`): `immediate` is passed to `requestDismiss` but remains unused. A full-height swipe therefore keeps the Sheet mounted and the body scroll locked for another 300 ms.

src/components/Sheet/SheetContent.tsx
```diff
@@ -242,22 +240,35 @@ export function SheetContent(props: SheetContentProps) {
...
 
+    const cancelDrag = React.useCallback(
+        (options: CancelSwipeOptions) => {
+            resetContentTouch();
+            cancelSwipe(options);
+        },
+        [cancelSwipe, resetContentTouch],
+    );
+
+    const onTouchCancel = React.useCallback(() => {
```
mournfulCoroner, line 265: 🤖 **AI generated**

🟠 **should-fix**

A `touchcancel` without actual movement also triggers restoration through `show()`. After `touchstart → touchcancel` without `touchmove`, `show()` sets `isAnimatingRef.current = true`, but the styles already represent the open state, so no `transitionend` event is emitted. Subsequent veil clicks are then ignored and resize handling remains deferred.

Could we restore only when the gesture actually changed the position, or explicitly clear the animation guard for a no-op restoration? A regression test should attempt to close the Sheet through the veil after this sequence.

src/components/Sheet/README.md
```diff
@@ -45,7 +45,14 @@ const SheetExample = () => {
 | hideTopBar                  | Hides the top bar with the resize handle.                                                                                                                        |   `boolean`   |                 |
 | id                          | Sheet ID used as hash in a URL. Make sure to specify multiple `id` values if there can be more than one sheet on a page.                                         |   `string`    |     `modal`     |
 | maxContentHeightCoefficient | Coefficient that determines the maximum height of the `Sheet` relative to the height of the viewport (range 0-1)                                                 |   `number`    |      `0.9`      |
-| onClose                     | Handler for close event.                                                                                                                                         |  `function`   |   `undefined`   |
+| container                   | DOM element to which component is mounted via `Portal`                                                                                                           | `HTMLElement` | `document.body` |
```
mournfulCoroner: AI
These properties (container through maxContentHeightCoefficient) duplicate the rows immediately above. Could we remove this repeated block so the generated property table contains each prop only once?

skoptenco: fixed

mournfulCoroner (approved):

src/components/Sheet/Sheet.tsx
```diff
@@ -57,50 +75,89 @@ export const Sheet = (rawProps: SheetProps) => {
...
+            requestDismiss({reason: 'escape-key', event});
+        },
+        [requestDismiss],
+    );
 
-    if (visible !== prevVisible) {
-        setPrevVisible(visible);
-    }
+    const floatingNodeId = useFloatingNodeId();
+    const {refs, context} = useFloating({
```
korvin89, line 90: 🤖 **AI generated**

🟠 **should-fix**

`useFloating` is called without `onOpenChange`, and Escape is routed through `useLayer({onEscapeKeyDown})` (line 108) instead of `useDismiss`. Modal (`Modal.tsx:187`), Popup (`Popup.tsx:287`) and Drawer (`Drawer.tsx:159`) all pass `onOpenChange` into `useFloating` and compose `useDismiss(context, {escapeKey, outsidePress})`, keeping `useLayer` purely for stacking.

Two consequences:

- The `FloatingTree` / `FloatingNode` scaffolding this PR adds is inert for dismissal. `useDismiss` is what consults the tree (it only closes a node with no open children, and can stop propagation); with Escape on the `LayerManager` flat stack, "close only the topmost Sheet" is delivered by the layer stack instead, and the tree ends up unused. `useRole` + `useInteractions` likewise only contribute a duplicate `role`, plus `tabindex="-1"`, `data-floating-ui-focusable` and a generated `id=":r1:"` on the dialog.
- Outside press bypasses the layer stack entirely (a plain `onClick` on the veil), so `disableOutsideClick: true` on line 107 has no effect.

Could we either move dismissal onto `useDismiss` like the three siblings, or add a short comment explaining why `Sheet` can't?

src/components/Sheet/Sheet.tsx
```diff
@@ -57,50 +75,89 @@ export const Sheet = (rawProps: SheetProps) => {
...
+        duration: SHEET_TRANSITION_DURATION_MS,
+        skipTransitionOut: immediate,
+        onTransitionOutComplete: handleExitComplete,
+    });
+
+    useLayer({
+        open: isMounted,
+        type: 'sheet',
+        disableOutsideClick: true,
+        onEscapeKeyDown: handleEscapeKeyDown,
```
korvin89, line 108: 🤖 **AI generated**

🟠 **should-fix**

Escape-to-dismiss is a new, always-on behaviour for every existing consumer (the previous `Sheet` had no key handling at all), and there is currently no way to opt out.

Escape closing a `role="dialog" aria-modal="true"` surface is the right default - it is what the APG dialog pattern asks for, and what mobile-sheet implementations do across the ecosystem (MUI `Drawer`/`Modal`, react-aria `ModalOverlay`, Radix `Dialog` and vaul on top of it, Ionic `ion-modal`). What they all add is an opt-out, in one of two shapes: a boolean prop (`disableEscapeKeyDown`, `isKeyboardDismissDisabled`, `keyboardClose`) or a cancellable event (`onEscapeKeyDown` + `preventDefault()`).

This repo already picked the first shape: `disableEscapeKeyDown` and `disableOutsideClick` on Modal (`Modal/README.md:39-40`), `disableEscapeKeyDown` on Popup (`Popup/README.md:146`) and Drawer (`Drawer/README.md:105`). Could we add the same pair to `Sheet`?

src/components/Sheet/README.md
```diff
@@ -45,7 +45,8 @@ const SheetExample = () => {
...
 | id                          | Sheet ID used as hash in a URL. Make sure to specify multiple `id` values if there can be more than one sheet on a page.                                         |   `string`    |     `modal`     |
 | maxContentHeightCoefficient | Coefficient that determines the maximum height of the `Sheet` relative to the height of the viewport (range 0-1)                                                 |   `number`    |      `0.9`      |
-| onClose                     | Handler for close event.                                                                                                                                         |  `function`   |   `undefined`   |
+| onClose                     | **Deprecated.** Callback invoked after the sheet closing animation completes. Use `onOpenChange` instead.                                                        |  `function`   |   `undefined`   |
+| onOpenChange                | Handler for open state change requests.                                                                                                                          |  `function`   |   `undefined`   |
```
korvin89, line 49: 🤖 **AI generated**

🟠 **should-fix**

The `onOpenChange` row doesn't describe the signature, the `reason` values (`escape-key` / `outside-press` / `swipe` / `navigation`), or the central contract of this PR - that a dismissal is a *request*, and nothing happens until the parent updates `visible`. It would be great to spell all three out here and in `README-ru.md:49`, and to cross-reference `Popup`'s `onOpenChange` (`Popup.tsx:67`, `Popup/README.md:163`) so the two callbacks stay described in the same terms.

Also, the intro on line 13 still says "To close it, swipe down or tap the area outside the `Sheet`" - Escape belongs in that sentence now.

src/components/Sheet/Sheet.tsx
```diff
@@ -57,50 +75,89 @@ export const Sheet = (rawProps: SheetProps) => {
...
-    }
+    const floatingNodeId = useFloatingNodeId();
+    const {refs, context} = useFloating({
+        nodeId: floatingNodeId,
+        open: requestedOpen,
+    });
+    const handleExitComplete = React.useCallback(() => {
+        onClose?.();
+    }, [onClose]);
+    const {isMounted, status} = useFloatingTransition({
```
korvin89, line 97: 🤖 **AI generated**

🟠 **should-fix**

`useFloatingTransition` is wired up here, but only `onTransitionOutComplete` is used, internally, to fire the deprecated `onClose`. Modal (`Modal.tsx:97-104`), Popup (`Popup.tsx:145-147`) and Drawer (`Drawer.tsx:124-127`) all expose `onTransitionIn` / `onTransitionInComplete` / `onTransitionOut` / `onTransitionOutComplete` as props.

Exposing them here looks nearly free, and there is a concrete caller waiting: `Select.tsx:131` still carries `// FIXME: add handlers to Sheet like in https://github.com/gravity-ui/uikit/issues/1354` with a 300 ms `setTimeout` to clear the filter after close. Would you consider adding them in this PR?

src/components/Sheet/hooks/useSheetDismiss.ts
```diff
@@ -0,0 +1,59 @@
...
+export function useSheetDismiss({
+    visible,
+    onOpenChange,
+}: {
+    visible: boolean;
+    onOpenChange?: SheetProps['onOpenChange'];
+}): UseSheetDismissResult {
+    const [legacyDismissed, setLegacyDismissed] = React.useState(false);
+    const [immediate, setImmediate] = React.useState(false);
+    const isControlled = Boolean(onOpenChange);
```
korvin89, line 26: 🤖 **AI generated**

🟠 **should-fix**

`isControlled = Boolean(onOpenChange)` makes the mere presence of the callback switch off the internal legacy dismissal path. An existing `onClose`-only consumer that adds `onOpenChange` for telemetry would silently get a sheet that can no longer be dismissed - the test `keeps the sheet open when controlled ... dismissal is not accepted` pins exactly that, including `onClose` never firing. The flag is also recomputed every render, so a conditionally supplied callback flips the mode mid-flight with no warning.

The repo's own helper for this is `useControlledState` (`src/hooks/useControlledState`), and it's worth noting how it differs: it keys "controlled" off the *value* (`value !== undefined`) rather than off the callback, and it `console.error`s when a component flips between the two modes. It isn't a drop-in here, because `visible` is required and the legacy path needs the sheet to close itself while `visible` stays `true` - but making `visible` optional with a `defaultVisible` companion would let `useControlledState` do this job and would line the API up with Modal/Drawer (see the `visible` vs `open` comment).

If the bespoke hook stays, could we at least document the switch in the `onOpenChange` / `onClose` README rows and warn in dev when it flips?

src/components/Sheet/constants.ts
```diff
@@ -2,6 +2,8 @@ import {block} from '../utils/cn';
 
 export const sheetBlock = block('sheet');
 
+export const SHEET_TRANSITION_DURATION_MS = 300;
```
korvin89, line 5: 🤖 **AI generated**

🟠 **should-fix**

300 ms now lives in five places: this constant, `SheetContent.tsx:23` (`TRANSITION_DURATION = '0.3s'`), `Sheet.scss:14`, `SheetVeil.scss:4` and `SheetContentArea.scss:4`. After this PR the unmount and the deprecated `onClose` are driven by the TS copy while the animation is driven by the SCSS copies, so any drift silently truncates the animation or leaves the sheet mounted.

`Drawer` already solves this in the repo: `DRAWER_ANIMATION_DURATION_MS` (`Drawer/constants.ts:1`) is fed into a CSS custom property (`Drawer.tsx:195-197` -> `Drawer.scss:92`). Could we do the same here?

src/components/Sheet/hooks/useVeil.ts
```diff
@@ -33,38 +31,37 @@ export function useVeil({
...
+    requestDismiss,
     onResizeWindow,
 }: UseVeilProps): UseVeilResult {
-    const latestRef = React.useRef({setVeilTouched, hide, hideSheet, onResizeWindow});
-    latestRef.current = {setVeilTouched, hide, hideSheet, onResizeWindow};
-
-    const getVeilOpacity = React.useCallback(() => veilRef.current?.style.opacity || 0, [veilRef]);
+    const onClick = React.useCallback(
+        (event: React.MouseEvent<HTMLDivElement>) => {
+            if (isAnimatingRef.current) {
```
korvin89, line 39: 🤖 **AI generated**

🟠 **should-fix**

This `isAnimatingRef` guard lost its test. The deleted `Sheet.veil.test.tsx` had `ignores the veil click while the sheet is still animating`, and it has no replacement - every new veil-click test calls `finishTransition()` first, so the guard is now uncovered. The other four deleted files are genuinely re-covered elsewhere, so this looks like the one accidental gap.

A smaller related gap: there is no test for the deprecated `onClose` after a **veil** dismissal (only after Escape and after an external `visible` change). Could we add both?

src/components/Sheet/Sheet.tsx
```diff
@@ -57,50 +75,89 @@ export const Sheet = (rawProps: SheetProps) => {
...
+        context,
+        duration: SHEET_TRANSITION_DURATION_MS,
+        skipTransitionOut: immediate,
+        onTransitionOutComplete: handleExitComplete,
+    });
+
+    useLayer({
+        open: isMounted,
```
korvin89, line 105: 🤖 **AI generated**

🟠 **should-fix**

`useLayer({open: isMounted})` diverges from Modal (`Modal.tsx:144`), Popup (`Popup.tsx:200`) and Drawer (`Drawer.tsx:138`), which all pass the controlled `open`. Because `isMounted` stays true for the whole 300 ms exit, the sheet keeps the top layer while closing: Escape during that window is swallowed and never reaches the layer below, and `getLayersCount()` / the `layerschange` event report the sheet longer than its siblings would.

`Sheet.test.tsx:165` pins this as intended, so it may well be deliberate - but could we align it with the other three (`open: requestedOpen`) so layer behaviour stays consistent across the library? A dismissal request while `requestedOpen` is already false is a no-op anyway.

src/components/Sheet/SheetContent.tsx
```diff
@@ -328,63 +337,70 @@ export function SheetContent(props: SheetContentProps) {
...
-                {...veilHandlers}
-            />
+            <SheetVeil veilRef={veilRef} withTransition={withTransition} {...veilHandlers} />
             <div
-                ref={sheetRef}
-                className={sheetBlock('sheet', sheetTransitionMod)}
+                ref={handleSheetRef}
+                className={sheetBlock('sheet', {'with-transition': withTransition})}
                 role="dialog"
                 aria-modal="true"
```
korvin89, line 401: 🤖 **AI generated**

🟠 **should-fix**

`role="dialog" aria-modal="true"` with no focus management: after opening, `document.activeElement` is still `<body>` - nothing traps focus, nothing sets initial focus, nothing returns it on close. Modal (`Modal.tsx:291-301`), Popup and Drawer (`Drawer.tsx:221-249`) all wrap their content in `FloatingFocusManager` with `initialFocus` / `returnFocus` / `visuallyHiddenDismiss`.

The gap predates this PR, but the migration is the natural place to close it, and most of the groundwork is already here - `getFloatingProps()` now puts `tabindex="-1"` and `data-floating-ui-focusable` on the dialog. It also matters more now that Escape closes the sheet: keyboard users can dismiss it without ever being able to reach it.

src/components/Sheet/Sheet.tsx
```diff
@@ -2,21 +2,38 @@
...
+export type SheetOpenChangeReason = 'escape-key' | 'outside-press' | 'swipe' | 'navigation';
+
 export interface SheetProps extends Pick<PortalProps, 'container' | 'disablePortal'>, QAProps {
     children?: React.ReactNode;
+    /** @deprecated Use onOpenChange instead */
     onClose?: () => void;
+    /** Callback for open state changes, when dismiss happens for example */
+    onOpenChange?: (open: boolean, event?: Event, reason?: SheetOpenChangeReason) => void;
     /** Show/hide sheet */
     visible: boolean;
```
korvin89, line 38: 🤖 **AI generated**

🟠 **should-fix**

`visible` stays while Drawer, Modal and Popup all use `open` (`Drawer/README.md:27`: `<Drawer onOpenChange={setVisible} open={isVisible}>`). Since this PR already introduces `onOpenChange` and deprecates `onClose`, this looks like the right moment to add `open` (plus an optional `defaultOpen`) and deprecate `visible`, so the whole family reads the same way. It would also give `useControlledState` something to work with - see the `useSheetDismiss` comment.

src/components/Sheet/hooks/useSheetHash.ts
```diff
@@ -29,83 +29,65 @@ export interface UseSheetHashResult {
...
-                currentHistory.replace(newLocation);
+                replace(newLocation);
                 break;
             case Platform.ANDROID:
-                currentHistory.push(newLocation);
+                push(newLocation);
                 break;
         }
-    }, []);
+    }, [id, location, platform, push, replace]);
```
korvin89, line 65: 🤖 **AI generated**

🟠 **should-fix**

Dropping `latestRef` ties `setHash` / `removeHash` - and through them `show` / `hide` - to the whole `location` object. The default `useLocationMock` (`mobile/MobileProvider.tsx:13-15`) returns a fresh object on every render, so `hide` changes identity on every render and the `status === 'close'` effect (`SheetContent.tsx:367-371`) re-invokes it on every render while the sheet is closing.

It's harmless today (the writes are idempotent and `removeHash` is guarded by `hashSetRef`), but it's avoidable churn on the closing path. Could we narrow the deps to `location.hash` / `location.pathname`?

korvin89 (commented):
🤖 **AI generated**

This review was done with AI assistance. I validated every comment myself, but the wording - including the points I found myself - was written by the AI.

**Severity legend**
- 🔴 **blocker** - should be fixed before merge.
- 🟠 **should-fix** - a real problem; worth fixing before or right after merge. If you can't get to it now, please open an issue and leave a comment linking it.
- 🔵 **nit** - minor; author's discretion.
- ❓ **question** - I need more context; not necessarily a defect.

src/components/Sheet/Sheet.tsx
```diff
@@ -57,50 +75,89 @@ export const Sheet = (rawProps: SheetProps) => {
         disablePortal,
         qa,
     } = useDefaultProps('Sheet', rawProps);
-    const [open, setOpen] = React.useState(visible);
-    const [prevVisible, setPrevVisible] = React.useState(visible);
+    const {requestedOpen, immediate, requestDismiss} = useSheetDismiss({visible, onOpenChange});
+    const veilRef = React.useRef<HTMLDivElement>(null);
```
korvin89, line 79: 🤖 **AI generated**

🟠 **should-fix**

`veilRef` and `isAnimatingRef` are created in `Sheet` but never read there - they're only forwarded to `SheetContentContainer` (lines 138-139). Besides widening `SheetContentBaseProps` for nothing, it makes their lifetime longer than `SheetContent`'s mount cycle: after an `immediate` exit no `transitionend` ever arrives on the veil, so `isAnimatingRef.current` is left `true` across the unmount. It recovers on the next open only because a freshly mounted veil starts at `opacity: 0` and animates again - a coincidence rather than a contract.

Could we move both refs back into `SheetContent`?

src/components/Sheet/Sheet.tsx
```diff
@@ -57,50 +75,89 @@ export const Sheet = (rawProps: SheetProps) => {
...
-        }
-        setOpen(false);
-    };
+    const role = useRole(context, {role: 'dialog'});
+    const {getFloatingProps} = useInteractions([role]);
 
-    if (!open) {
-        return null;
+    if (!isMounted) {
+        return <FloatingNode id={floatingNodeId} />;
```
korvin89, line 115: 🤖 **AI generated**

🟠 **should-fix**

`Modal.tsx:280-322` and `Drawer.tsx:254` keep a single `<FloatingNode id>{...}</FloatingNode>` and hide the contents inside it; here the not-mounted case early-returns a childless `<FloatingNode />` instead. The two are functionally equivalent, so could we use the sibling shape for consistency?

src/components/Sheet/SheetContent.tsx
```diff
@@ -328,63 +337,70 @@ export function SheetContent(props: SheetContentProps) {
...
-                withTransition={veilTransitionMod['with-transition']}
-                {...veilHandlers}
-            />
+            <SheetVeil veilRef={veilRef} withTransition={withTransition} {...veilHandlers} />
             <div
-                ref={sheetRef}
-                className={sheetBlock('sheet', sheetTransitionMod)}
+                ref={handleSheetRef}
+                className={sheetBlock('sheet', {'with-transition': withTransition})}
                 role="dialog"
```
korvin89, line 400: 🤖 **AI generated**

🔵 **nit**

`role="dialog"` is written literally here and then written again by `{...getFloatingProps()}` on line 403 - for `role: 'dialog'`, `useRole` returns `{id, role}`. `aria-label={title}` also sits before the spread, so it would lose to any `aria-labelledby` the hook starts returning later. And the spread now puts a generated `id=":r1:"` on a user-facing element, which can break consumer DOM snapshots.

Modal, Popup and Drawer don't write the role literally. Could we drop either the literal or `useRole` + `useInteractions` (which currently contribute little else)?

src/components/Sheet/SheetContent.tsx
```diff
@@ -19,16 +24,22 @@ const TRANSITION_DURATION = '0.3s';
 const DEFAULT_MAX_CONTENT_HEIGHT_FROM_VIEWPORT_COEFFICIENT = 0.9;
 const WINDOW_RESIZE_TIMEOUT = 50;
 
+export type SheetPresenceStatus = 'unmounted' | 'initial' | 'open' | 'close';
```
korvin89, line 27: 🤖 **AI generated**

🔵 **nit**

`SheetPresenceStatus` repeats `UseFloatingTransitionResult['status']` verbatim - importing it would keep the two in sync for free. Having it next to the existing `Status` (`'showing' | 'hiding'`) in the same file is also what forced the `status: nextStatus` rename on line 138; one of the two names could be more specific.

src/components/Sheet/hooks/useSheetDismiss.ts
```diff
@@ -0,0 +1,59 @@
...
+        if (visible) {
+            setLegacyDismissed(false);
+        }
+    }, [visible]);
+
+    React.useEffect(() => {
+        if (requestedOpen) {
+            setImmediate(false);
+        }
+    }, [immediate, requestedOpen]);
```
korvin89, line 39: 🤖 **AI generated**

🔵 **nit**

`immediate` is in the dependency list although the effect body never reads it. As far as I can tell it is load-bearing - it re-fires the reset when a refused `immediate` request flips the flag while the sheet is still open - but that's impossible to infer from the code. Could we add a comment, or reset `immediate` explicitly on the refusal path instead?

src/components/Sheet/hooks/useSwipe.ts
```diff
@@ -145,6 +195,7 @@ export function useSwipe({
         setDeltaY,
         setSwipeAreaTouched,
```
korvin89, line 196: 🤖 **AI generated**

🔵 **nit**

`setSwipeAreaTouched` is declared in `UseSwipeResult` (line 47) and returned here, but `SheetContent.tsx:218-229` no longer destructures it and there are no other consumers. Looks like it can go.

src/components/Sheet/hooks/useSwipe.ts
```diff
@@ -16,38 +18,44 @@ export interface UseSwipeProps {
...
+    startYRef: React.MutableRefObject<number | null>;
     deltaYRef: React.MutableRefObject<number>;
     swipeAreaTouchedRef: React.MutableRefObject<boolean>;
     setDeltaY: (value: number) => void;
     setSwipeAreaTouched: (value: boolean) => void;
-    onTouchEndAction: (deltaY: number) => void;
-    swipeAreaHandlers: SwipeAreaHandlers;
+    onTouchEndAction: (deltaY: number, event: React.TouchEvent<HTMLDivElement>) => void;
+    cancelSwipe: (options: CancelSwipeOptions) => void;
+    swipeAreaHandlers: Omit<SwipeAreaHandlers, 'onTouchCancel'>;
```
korvin89, line 50: 🤖 **AI generated**

🔵 **nit**

`onTouchCancel` is required in `SwipeAreaHandlers` (and in `ContentAreaHandlers`, `useContentScroll.ts:44`), yet both hooks return `Omit<..., 'onTouchCancel'>` and `SheetContent` attaches the handler by hand in JSX (lines 417 and 427). Since `useSwipe` already owns `cancelSwipe`, would it be cleaner for the hook to return the handler too, so the types and the returned objects agree?

src/hooks/private/useFloatingTransition/useFloatingTransition.ts
```diff
@@ -8,6 +8,8 @@ import {usePrevious} from '../usePrevious';
 export interface UseFloatingTransitionProps {
     context: FloatingContext;
     duration: NonNullable<UseTransitionStatusProps['duration']>;
+    /** Completes the close phase without waiting for the transition duration. */
+    skipTransitionOut?: boolean;
```
korvin89, line 12: 🤖 **AI generated**

🔵 **nit**

`src/hooks/private/useFloatingTransition/README.md` isn't updated for this new `skipTransitionOut` prop. That README is already out of date (it documents `cssTransitionEnabled`, `cssTransitionProperty` and `handleTransitionEnd`, none of which exist in the hook), so a refresh here would cover both.

src/components/Sheet/Sheet.tsx
```diff
@@ -57,50 +75,89 @@ export const Sheet = (rawProps: SheetProps) => {
...
-                />
-            </FloatingOverlay>
-        </Portal>
+        <FloatingNode id={floatingNodeId}>
+            <Portal container={container} disablePortal={disablePortal}>
+                <FloatingOverlay
+                    data-qa={qa}
+                    data-floating-ui-status={status}
+                    className={sheetBlock({'without-top-bar': hideTopBar}, className)}
+                    lockScroll={isMounted}
```
korvin89, line 125: 🤖 **AI generated**

🔵 **nit**

`lockScroll={isMounted}` sits after the `if (!isMounted) return ...` above, so it is always `true`. `lockScroll` alone would read more clearly.

src/components/Sheet/Sheet.tsx
skoptenco: A direct switch to useDismiss would lose our global stacking guarantees: independent Sheets can belong to separate FloatingTrees. LayerManager ensures Escape affects only the topmost Sheet and keeps lower Sheets blocked until its exit animation finishes. Preserving this behavior would require additional coordination.

src/components/Sheet/hooks/useSheetDismiss.ts
skoptenco: I’d handle this in a separate PR. In my view, the right solution requires migrating to useControlledState, making visible optional, and adding defaultVisible, which also means changing the component’s API

src/components/Sheet/SheetContent.tsx
skoptenco: This is addressed in #2822, which builds on this PR and adds FloatingFocusManager, initial focus, focus containment, focus restoration, and visually hidden dismiss controls, with regression tests.

src/components/Sheet/Sheet.tsx
skoptenco: I’d handle this in a separate PR together with the useControlledState migration. Adding open / defaultOpen and deprecating visible requires defining compatibility rules and preserving the legacy dismissal behavior.

src/components/Sheet/Sheet.tsx
skoptenco: fixed

src/hooks/private/useFloatingTransition/useFloatingTransition.ts
skoptenco: fixed

src/components/Sheet/hooks/useSwipe.ts
skoptenco: fixed

src/components/Sheet/hooks/useSwipe.ts
skoptenco: fixed

src/components/Sheet/hooks/useSheetDismiss.ts
skoptenco: immediate is required in dependencies. Added comment with explanations

src/components/Sheet/Sheet.tsx
skoptenco: added onTransitionIn, onTransitionInComplete, onTransitionOut, onTransitionOutComplete

src/components/Sheet/README.md
skoptenco: fixed

src/components/Sheet/Sheet.tsx
skoptenco: added disableEscapeDown and disableOutsideClick

src/components/Sheet/Sheet.tsx
korvin89: 🤖 **AI generated**

Thank you - and the `FloatingTree` half of this is more right than it reads. I checked it on `main`: two sibling `<Modal open>` elements each create their own `FloatingTree` (`Modal.tsx:327-333`, the same root detection `Sheet` uses), and a single Escape fires `onOpenChange` on **both**, because `useDismiss` only coordinates inside its own tree. So routing Escape through `LayerManager` isn't a shortcut here - it's what gives `Sheet` a guarantee `Modal` doesn't have. No objection to keeping it.

Two follow-ups, though:

- The other half of the rationale - "keeps lower Sheets blocked until its exit animation finishes" - no longer holds after the `useLayer({open: requestedOpen})` change; the new test is called `routes the next Escape to the lower sheet while the top sheet is exiting`. Could we restate it so the thread doesn't mislead the next reader?
- `Sheet` is now the only component in the repo that dismisses through `useLayer` rather than `useDismiss`, and nothing in the code says why. Could we record the surviving reason as a short comment next to the `useLayer` call? Otherwise the next person migrating a component to `useFloating` will very reasonably "align" it back and quietly lose the guarantee.

Optionally: if sibling `Modal`s / `Popup`s really do double-dismiss on Escape, that looks worth its own issue - happy to be corrected if there's coordination elsewhere I've missed.

src/components/Sheet/README.md
korvin89: 🤖 **AI generated**

The new rows read really well, thank you - the `onOpenChange` cross-reference to `Popup` and the reason list are exactly what was missing, and `README-ru.md` is in full parity.

One row stayed behind, in both files: `onClose` still says "Callback invoked after the sheet closing animation completes. Use `onOpenChange` instead." (`README.md:50`, and the same in `README-ru.md:50`), while the JSDoc on `Sheet.tsx:33` is now sharper - "Use `onOpenChange` for dismissal requests or `onTransitionOutComplete` for exit cleanup". Since `onTransitionOutComplete` is the actual replacement for the post-exit timing and `onOpenChange` isn't, could we bring both README rows in line with the JSDoc?

src/components/Sheet/hooks/useSheetDismiss.ts
korvin89: 🤖 **AI generated**

That split sounds right, thank you. Two requests so it doesn't get lost, and one question about the v7 side.

1. Could we open an issue for the follow-up and link it here, as the 🟠 should-fix legend suggests? Scope as I understand it: migrate to `useControlledState` and replace `visible` outright with `open` + `defaultOpen` (the other thread) - not an interim `defaultVisible`, since the rename makes that shape legacy on arrival.
2. Could that PR target the `next` branch once this one is merged, rather than `main`? The next major is about a month out, and a breaking prop rename belongs there rather than as another deprecation in a minor.
3. The two halves land differently, though. The rename waits for the major, but `isControlled = Boolean(onOpenChange)` ships in v7 the moment this PR merges - so for the rest of v7, adding `onOpenChange` next to an existing `onClose` turns off the internal dismissal. That isn't detectable at runtime (passing the callback *is* the opt-in), so documentation is the only real mitigation, and the `onOpenChange` README row already carries it. Would a dev-only warning when the mode flips between renders - the way `useControlledState` warns on controlled / uncontrolled - be worth adding here as well?

src/components/Sheet/Sheet.tsx
korvin89: 🤖 **AI generated**

Agreed, same follow-up. I've folded this into the request on the `useSheetDismiss` thread: one issue covering the `useControlledState` migration plus replacing `visible` outright with `open` / `defaultOpen`, with that PR targeting the `next` branch so the rename lands in the upcoming major.

src/components/Sheet/Sheet.tsx
skoptenco: Yeah, I've left this comment before useLayer({open: requestedOpen}) fixes, my bad. Now LayerManager ensures Escape reaches only the topmost layer across independent FloatingTrees, whereas useDismiss coordinates within each tree. The layer is released when closing starts, so the next Escape can reach the lower Sheet while the upper one is still exiting. Added a comment next to useLayer documenting this rationale.

src/components/Sheet/hooks/useSheetDismiss.ts
skoptenco: created [sub-issue](https://github.com/gravity-ui/uikit/issues/2773?issue=gravity-ui%7Cuikit%7C2843)
pr for that issue will upload in next branch
also fixed onOpenChange description in docs

src/components/Sheet/SheetContent.tsx
korvin89: 🤖 **AI generated**

Thanks for pointing at #2822 - the scope there looks right for this.

One practical note, though: #2822 currently reports `mergeable: false` / `mergeable_state: dirty`. Its merge base with this branch is `ff3b7202`, so it sits 27 commits behind its own base, and four of the files it touches - `Sheet.tsx`, `SheetContent.tsx`, `README.md` and `README-ru.md` - have changed here since, in exactly the places the migration reworked: the veil and animation refs moved into `SheetContent`, `status` became `presenceStatus`, the literal `role="dialog"` gave way to `getFloatingProps({'aria-label': title})`, and several props plus their JSDoc are new.

This PR is now `mergeable: true`, so could #2822 be rebased onto the current head before this one merges? Otherwise the only remaining 🟠 should-fix from this review is parked in a branch that won't merge cleanly, and once this lands it will need a retarget to `main` on top of resolving those conflicts.

src/components/Sheet/README.md
korvin89: 🤖 **AI generated**

The updated `onClose` rows match the JSDoc now - thank you.

The parity slipped the other way in one spot, though: the `onOpenChange` README row now carries the **Warning** that providing the callback disables internal dismissal (even alongside `onClose`, or when it is used only for analytics) plus the note about not adding or removing it while mounted - but `Sheet.tsx:35` still reads just "Callback for open state changes, when dismiss happens for example". That warning is the thing a consumer most needs at the call site, and the JSDoc is what they see in autocomplete. Could we carry a short version of it into the JSDoc as well?

src/components/Sheet/SheetContent.tsx
skoptenco: We discussed with Polina about merging her PR and came to agreement, that she will rebase on main after my merge

korvin89 (approved):

PR #2793, opened 2026-08-25
fix(Drawer): capture pointer while resizing

### LINKED ISSUE #2792: Drawer resize stops when pointer crosses an iframe
### Describe the bug

When a resizable `Drawer` is rendered with `hideVeil` and the resize pointer moves over an iframe behind the Drawer, resizing stops.

The resize implementation subscribes to `pointermove` and `pointerup` on the parent `window`. Pointer events dispatched inside an iframe belong to a different browsing context and do not reach that window. As a result:

- the Drawer stops following the pointer while it is over the iframe;
- the next event received after leaving the iframe may cause a visible size jump;
- if `pointerup` happens over the iframe, the resize interaction may remain unfinished.

This is reproducible with the current published version, `@gravity-ui/uikit@7.48.1`.

I intend to submit a pull request. The proposed change is to capture the active pointer on the resize handle during `pointerdown`:

```ts
event.currentTarget.setPointerCapture(event.pointerId);
```

Pointer capture keeps subsequent events targeted at the resize handle even when normal hit testing moves over the iframe. No Drawer API or sizing logic needs to change.

The pull request will include a regression test verifying that the active `pointerId` is captured when resizing starts.

@amje @korvin89, could you please confirm that this approach is acceptable?

### Reproduction

https://stackblitz.com/edit/vite-react-ts-bzt8osfs

### Steps to reproduce

1. Open the StackBlitz preview at a width of at least 600 px.
2. Keep the Drawer open.
3. Grab the resize handle at the vertical center of the red iframe.
4. Drag the handle left into the iframe.
5. Release the pointer while it is still over the iframe.

The reported Drawer width remains unchanged when the pointer enters the iframe. In an automated Chromium check, the initial width was `300px`; after moving the pointer `120px` into the iframe, it was still `300px`.

### Environment

- `@gravity-ui/uikit`: `7.48.1`
- React: `18.3.1`
- Browsers: Yandex Browser `26.4`, Chromium `141`
- OS: macOS, Linux

### Validations

xfast1205:
## Summary

- capture the active pointer on the Drawer resize handle
- keep resize and pointerup handling working when the pointer crosses an iframe
- add a Playwright regression test that reproduces the iframe drag and verifies that the resize interaction finishes on pointerup

Fixes #2792.

## Testing

- `npm run lint`
- `npm run typecheck`
- `npm test -- --runInBand`
- `npm run build`
- `npm run playwright -- src/components/Drawer/__tests__/Drawer.visual.test.tsx --reporter=line`

I hereby agree to the terms of the CLA available at: https://yandex.ru/legal/cla/?lang=en.


korvin89:
🤖 **AI generated**

Thanks for the fix — `setPointerCapture` looks like the right mechanism here, and the reasoning in #2792 matches how pointer capture actually behaves: events are retargeted to the handle and still bubble up to `window`, so the existing listeners keep working without any change.

I have a suggestion about the **test**, not about the fix itself.

## Why I think a different test would serve this better

`useResizeHandlers.test.tsx` stubs `setPointerCapture` onto the node and then asserts that the stub was called. To be fair, that isn't really avoidable in jsdom: the version this repo uses (jsdom 20) implements neither `Element.prototype.setPointerCapture` nor `PointerEvent`, and `fireEvent.pointerDown(el, {pointerId})` silently drops the prop because there is no `PointerEvent` constructor to fall back on. So the test can assert that the new line exists, but it cannot observe the behavior the issue describes — the bug is about a pointer crossing into a different browsing context, which jsdom has no notion of at all.

## What a Playwright test can check instead

The repo already has component-test infrastructure, so I put a test together and ran it both ways — once as-is, and once with the `setPointerCapture` line removed:

| | without the fix | with the fix |
| :--- | ---: | ---: |
| initial width | 308 | 308 |
| after dragging 400px into the iframe | **308** (unchanged) | **408** |
| after `mouse.up()` over the iframe | 308 | 408 |
| after moving the mouse with no button held | **208** | 408 |

The second row is the symptom from the issue, reproduced with the same numbers you reported. The last row is worth a look too: without the fix, the `pointerup` over the iframe never reaches the parent window, so the `pointermove` listener stays attached and the Drawer keeps resizing with the button already released, collapsing all the way to `minSize`. That is the "resize interaction may remain unfinished" part of the report, and this test pins it down as well.

### `src/components/Drawer/__tests__/components.tsx`

```tsx
export function ResizeOverIframeShowcase() {
    const [size, setSize] = React.useState(300);
    // Stable identity on purpose: an inline arrow here changes `onResize` on every
    // render, which tears down the window listeners in useResizeHandlers mid-drag.
    const onResize = React.useCallback((next: number) => setSize(next), []);

    return (
        <div style={{position: 'relative', width: '1200px', height: '600px'}}>
            <iframe
                title="background"
                srcDoc="<body style='margin:0;height:100%;background:#c00'></body>"
                style={{position: 'absolute', inset: 0, width: '100%', height: '100%', border: 0}}
            />
            <Drawer
                open
                hideVeil
                disableTransition
                placement="right"
                resizable
                size={size}
                minSize={200}
                maxSize={400}
                onResize={onResize}
                qa="test-drawer"
            >
                <div style={{height: '100%', padding: '20px'}} data-qa="test-drawer-item" />
            </Drawer>
        </div>
    );
}
```

### `src/components/Drawer/__tests__/Drawer.visual.test.tsx`

```tsx
test('should keep resizing while the pointer is over an iframe', async ({mount, page}) => {
    await page.setViewportSize({width: 1200, height: 600});
    await mount(<ResizeOverIframeShowcase />);

    const item = page.locator('.g-drawer__item');
    const resizer = page.locator('.g-drawer__resizer');
    await expect(item).toBeVisible();

    const widthOf = async () => {
        const box = await item.boundingBox();
        return box ? Math.round(box.width) : -1;
    };

    // The rendered box is `size` plus the 8px resizer gutter, so assert on deltas.
    const initialWidth = await widthOf();

    const handle = await resizer.boundingBox();
    if (!handle) {
        throw new Error('resizer is not rendered');
    }
    const startX = handle.x + handle.width / 2;
    const y = handle.y + handle.height / 2;

    await page.mouse.move(startX, y);
    await page.mouse.down();

    // Drag 400px to the left. maxSize caps the drawer at 400px, so its edge stops
    // at x=800 while the pointer travels on to x=startX-400, ~300px deep into the
    // iframe -- the pointer is unambiguously over the iframe, not over the Drawer.
    await page.mouse.move(startX - 400, y, {steps: 20});

    await expect.poll(widthOf).toBe(initialWidth + 100);

    // Release while still over the iframe: the interaction must actually finish.
    await page.mouse.up();
    await expect.poll(widthOf).toBe(initialWidth + 100);

    // Moving back over the Drawer with no button held must not resize anything.
    // Without a processed pointerup the window pointermove listener is still
    // attached and this would drag the Drawer down to minSize.
    await page.mouse.move(1100, y, {steps: 10});
    expect(await widthOf()).toBe(initialWidth + 100);
    await expect(item).toBeVisible();
});
```

### A few notes on why it is written this way

- **`maxSize={400}` is load-bearing, not cosmetic.** The handle follows the pointer, so the pointer normally sits right at the Drawer's edge and "over the iframe" stays ambiguous. Capping the size stops the edge at `x=800` while the pointer travels on to `x≈496`, roughly 300px inside the iframe.
- **`srcDoc` rather than `src`.** `about:srcdoc` is same-origin with the parent, and the bug still reproduces — what matters is the separate browsing context, not the origin.
- **Assertions on deltas.** `.g-drawer__item_resizable` gets `padding-inline-start: 8px` with `content-box` sizing, so the rendered box is `size + 8`. Absolute numbers would encode a layout detail.
- **The showcase uses a stable `onResize` deliberately.** With an inline arrow the test fails for an unrelated reason: `onResize` changes identity on every render, `handleMove`/`handleUp` change with it, and the cleanup effect in `useResizeHandlers` removes the `window` listeners without re-adding them, so the drag dies after the first `pointermove`. Might be worth a separate look, but it is out of scope for this PR.

### Two caveats

- The file has to be named `*.visual.test.tsx` and live in `__tests__/` for Playwright to pick it up (`testMatch: '**/__tests__/*.visual.test.tsx'`), even though this test takes no screenshots. Happy to defer to you on whether that naming is acceptable or whether `testMatch` should be widened instead.
- The config's default `timeout: 10 * 1000` was not enough for page setup on a cold local run; it was comfortable once warm (~3.5s per test). Worth keeping an eye on in CI.

Either way, this is a suggestion rather than a merge condition — the fix itself looks correct to me.

xfast1205:
Thanks for the detailed suggestion and for validating the scenario both with and without the fix.

I've replaced the jsdom test with a Playwright regression based on your scenario in [`25442528`](https://github.com/gravity-ui/uikit/commit/25442528ddab2ed8d31ba58b2f19741806921a13). It now verifies that:

- resizing continues while the pointer is over the iframe;
- `pointerup` over the iframe finishes the resize interaction;
- subsequent pointer movement with no button held does not resize the Drawer.

The showcase keeps `onResize` stable, and the reason for the `maxSize` cap is documented in the test. I left the callback identity issue out of scope for this PR.

korvin89 (approved):

PR #2799, opened 2026-08-28
feat: expose default props provider

amje:


src/components/theme/DefaultPropsProvider.tsx
```diff
@@ -127,18 +129,27 @@ export interface ComponentDefaultPropsMap {
 
 const EMPTY: ComponentDefaultPropsMap = {};
 
-export const PrivateDefaultPropsContext = React.createContext<ComponentDefaultPropsMap>(EMPTY);
+export const DefaultPropsContext = React.createContext<ComponentDefaultPropsMap>(EMPTY);
 
-export function PrivateDefaultPropsProvider({
-    value = EMPTY,
+export function DefaultPropsProvider({
```
korvin89: 🤖 **AI generated**

🟠 **should-fix** — Now that this provider is public, would you consider giving it a named, exported props interface? The rest of the repo consistently does that — `ThemeProviderProps` (`ThemeProvider.tsx:33`), `PrivateLayoutProviderProps` (`layout/LayoutProvider/LayoutProvider.tsx:10`), `TabProviderProps` — and `ComponentDefaultPropsMap` in this very file follows the same rule.

The new test already has to reach for `React.ComponentProps<typeof DefaultPropsProvider>['value']` (`__tests__/useDefaultProps.test.tsx:8`), which is a good sign the type is missing. Suggestion:

```tsx
export interface DefaultPropsProviderProps {
    value?: ComponentDefaultPropsMap;
    children: React.ReactNode;
}
```

…and re-export it from `src/components/theme/index.ts` alongside the component.

src/components/theme/index.ts
```diff
@@ -12,4 +12,5 @@ export * from './withDirection';
 export * from './getThemeType';
 export {useLang} from './useLang';
 export type {Theme, RealTheme, ThemeType, Direction, ThemeContextProps} from './types';
-export type {ComponentDefaultPropsMap} from './PrivateDefaultPropsProvider';
+export {DefaultPropsProvider} from './DefaultPropsProvider';
```
korvin89, line 15: 🤖 **AI generated**

🟠 **should-fix** — It would be great to land some documentation together with this export. `docs/theming.md` and `docs/theming-ru.md` cover `ThemeProvider` in detail, but a `grep` for `defaultProps` across all `*.md`/`*.mdx` returns nothing — neither `DefaultPropsProvider` nor the existing `ThemeProvider.defaultProps` is described anywhere, and there is no story in `src/components/theme/__stories__/Theme.stories.tsx` either.

A short section in `docs/theming.md` (+ the `-ru` counterpart) with a usage example and the merge rules would make this API discoverable. This is the moment where the whole mechanism becomes public, so it seems like the right place to write it down.

src/components/theme/DefaultPropsProvider.tsx
```diff
@@ -127,18 +129,27 @@ export interface ComponentDefaultPropsMap {
...
     children: React.ReactNode;
 }) {
+    const parentDefaultProps = React.useContext(DefaultPropsContext);
+    const mergedDefaultProps = React.useMemo(() => {
+        if (!value) {
+            return parentDefaultProps;
+        }
+
+        return {...parentDefaultProps, ...value};
+    }, [parentDefaultProps, value]);
```
korvin89: 🤖 **AI generated**

🟠 **should-fix** — The `useMemo` is keyed on the identity of `value`, so an inline literal (`value={{Button: {view: 'action'}}}`) produces a fresh context value on every render of the parent, which re-renders every subscriber below. There are currently 60 components calling `useDefaultProps`, so the blast radius is the whole subtree.

This was much less of a concern while the provider lived only inside `ThemeProvider` at the app root, but now people will place it inside the tree, where the parent re-renders often. I'd suggest at minimum documenting "memoize or hoist `value`", and possibly comparing `value` shallowly by component key so a structurally identical object doesn't invalidate the context.

src/components/controls/PasswordInput/PasswordInput.tsx
```diff
@@ -36,7 +37,8 @@ export type PasswordInputProps = Omit<TextInputProps, 'type'> & {
     onRevealValueUpdate?: (value: boolean) => void;
 };
 
-export const PasswordInput = (props: PasswordInputProps) => {
+export const PasswordInput = (rawProps: PasswordInputProps) => {
+    const props = useDefaultProps('PasswordInput', rawProps);
```
korvin89: 🤖 **AI generated**

🟠 **should-fix** — `size` is read from `props` with a local `'m'` fallback (line 50), `actionButtonSize` is derived from it (line 65), and then the whole `{...props}` is spread into `TextInput` (line 107) — where `TextInput` applies *its own* defaults on top.

So with `defaultProps={{TextInput: {size: 'l'}}}` the input renders at `size="l"` while the reveal/copy buttons render at `size="s"` (`getActionButtonSizeAndIconSize('m')` → `'s'`) instead of the expected `'m'`. With `defaultProps={{PasswordInput: {size: 'l'}}}` everything lines up correctly.

The bug predates this PR, but this PR is what wires `PasswordInput` into the defaults system and makes the mismatch reachable through two different keys — so it feels like a natural place to close it. Suggested fix: derive `actionButtonSize` from the same size that actually reaches `TextInput`, and pass `size` to `TextInput` explicitly rather than via the spread.

src/components/theme/DefaultPropsProvider.tsx
```diff
@@ -127,18 +129,27 @@ export interface ComponentDefaultPropsMap {
...
     value?: ComponentDefaultPropsMap;
     children: React.ReactNode;
 }) {
+    const parentDefaultProps = React.useContext(DefaultPropsContext);
+    const mergedDefaultProps = React.useMemo(() => {
+        if (!value) {
+            return parentDefaultProps;
+        }
+
+        return {...parentDefaultProps, ...value};
```
korvin89: 🤖 **AI generated**

🟠 **should-fix** — The merge is asymmetric: components are merged at the map level, but the props of a single component are replaced wholesale. A nested `{Button: {view: 'action'}}` silently drops an inherited `size: 'l'` — which the new test does encode, so I assume it is deliberate.

It matches the previous `ThemeProvider` behaviour, so there is no regression here. But composing providers is precisely what this PR enables, and "let me override one prop for this subtree" quietly losing the rest is an easy trap. Could this be spelled out explicitly in the docs? And it might be worth considering a one-level merge per component key instead.

src/components/theme/DefaultPropsProvider.tsx
```diff
@@ -127,18 +129,27 @@ export interface ComponentDefaultPropsMap {
...
+export function DefaultPropsProvider({
+    value,
     children,
 }: {
     value?: ComponentDefaultPropsMap;
     children: React.ReactNode;
 }) {
+    const parentDefaultProps = React.useContext(DefaultPropsContext);
+    const mergedDefaultProps = React.useMemo(() => {
+        if (!value) {
```
korvin89: 🤖 **AI generated**

❓ **question** — With the parent merge now living in the provider itself, is there still a way to *reset* inherited defaults for a subtree? `value={{}}` now inherits everything from the parent, so "render this branch without any defaults" doesn't seem expressible.

Is that intentional for now, or is an escape hatch planned (something like `value={null}` or a `reset` prop)?

src/components/theme/__tests__/useDefaultProps.test.tsx
```diff
@@ -76,6 +76,28 @@ describe('useDefaultProps', () => {
             });
             expect(result.current).toBe(props);
         });
+
+        it('merges nested providers by component and replaces defaults for the same component', () => {
```
korvin89: 🤖 **AI generated**

🟠 **should-fix** — A few gaps in coverage for the newly added logic:

- The `if (!value) return parentDefaultProps` branch (`DefaultPropsProvider.tsx:143-145`) is the only new code path in the file and no test exercises it — in particular a `<DefaultPropsProvider>` with no `value` nested inside one that has defaults, asserting the defaults pass straight through.
- The `ThemeProvider defaultProps` + nested `DefaultPropsProvider` interaction (and the reverse nesting) isn't covered, even though that composition is the main reason the merge moved into `DefaultPropsProvider`.
- The added test mirrors the existing `inner ThemeProvider defaultProps correctly override outer` case (lines 145-163) fairly closely, so the genuinely new scenarios are the ones above.

src/components/theme/DefaultPropsProvider.tsx
korvin89: 🤖 **AI generated**

🔵 **nit** — Could we add a `displayName` here? The neighbouring `ThemeProvider` sets one (`ThemeProvider.tsx:153`), and it helps in React DevTools and Storybook now that this component is public.

src/components/theme/__tests__/useDefaultProps.test.tsx
```diff
@@ -76,6 +76,28 @@ describe('useDefaultProps', () => {
...
+            const wrapper = ({children}: {children: React.ReactNode}) => (
```
korvin89: 🤖 **AI generated**

🔵 **nit** — Two small structural things about the new test:

- It declares an inline `wrapper`, while the file already has a `makeWrapper` helper (line 8). Extending that helper to take an outer and inner value would keep the file consistent.
- It sits under `describe('useDefaultProps') > describe('prop merging priority')`, but what it actually verifies is `DefaultPropsProvider` composition rather than the hook. A sibling `describe('DefaultPropsProvider')` block would be a more natural home.

korvin89 (commented):
🤖 **AI generated**

**Severity legend**

src/components/theme/DefaultPropsProvider.tsx
```diff
@@ -127,18 +129,27 @@ export interface ComponentDefaultPropsMap {
 
 const EMPTY: ComponentDefaultPropsMap = {};
 
-export const PrivateDefaultPropsContext = React.createContext<ComponentDefaultPropsMap>(EMPTY);
+export const DefaultPropsContext = React.createContext<ComponentDefaultPropsMap>(EMPTY);
 
-export function PrivateDefaultPropsProvider({
-    value = EMPTY,
+export function DefaultPropsProvider({
+    value,
```
korvin89: 🤖 **AI generated**

🔵 **nit** — The prop is named `value` here, while `ThemeProvider` calls the same object `defaultProps` (`ThemeProvider.tsx:41`). Since both are public now, having two names for one concept is a little confusing — worth aligning if there's no reason for the difference.

src/components/theme/DefaultPropsProvider.tsx
amje: Fixed

src/components/theme/index.ts
amje: Fixed

src/components/theme/DefaultPropsProvider.tsx
amje: Added note to the docs

src/components/controls/PasswordInput/PasswordInput.tsx
amje: Fixed

src/components/theme/DefaultPropsProvider.tsx
amje: Mention this in the docs

src/components/theme/DefaultPropsProvider.tsx
amje: I will add this in the future it there will be a request for that

src/components/theme/__tests__/useDefaultProps.test.tsx
amje: Added more tests

src/components/theme/__tests__/useDefaultProps.test.tsx
amje: Changed test structure

src/components/theme/DefaultPropsProvider.tsx
amje: Renamed

src/components/theme/DefaultPropsProvider.tsx
amje: Added

korvin89 (approved):
🤖 **AI generated**

Thanks for working through all the comments — the docs section, the exported props type, the `defaultProps` rename and the extra provider-composition tests all look good to me. LGTM.

PR #2800, opened 2026-08-28
fix(Modal): fix mobile styles

amje:


src/components/Modal/Modal.scss
```diff
@@ -86,11 +86,16 @@ $modal-transform-transition: transform $modal-transition-duration ease-out;
...
-        --g-modal-border-radius: 0px;
+        #{$block}__content {
+            margin: 0;
+            height: 100dvh;
+            min-height: 100svh;
+            max-height: initial;
+            width: 100vw;
+            min-width: initial;
+            max-width: initial;
+            border-radius: 0;
```
korvin89: 🤖 **AI generated**

🔴 **blocker** — `border-radius: 0` is set here, but the base rule's `clip-path: inset(0 round var(--g-modal-border-radius, #{$modal-default-border-radius}))` (line 45) isn't overridden, and it no longer receives `--g-modal-border-radius: 0px`. So the full-screen mobile content ends up clipped with a 5px rounded rect — or with whatever radius an app sets through the documented theming token (`docs/theming.md:179`) — while `border-radius` reports square corners.

I checked this in a browser against the real `Modal` → `Dialog` DOM at 390×844: computed `clip-path` on `.g-modal__content` goes from `inset(0px)` to `inset(0px round 5px)`.

Could we add `clip-path: inset(0);` to this rule? The snippet in my other comment covers it too.

src/components/Modal/Modal.scss
```diff
@@ -86,11 +86,16 @@ $modal-transform-transition: transform $modal-transition-duration ease-out;
     }
 
     &_mobile {
-        --g-modal-width: 100vw;
-        --g-modal-height: 100dvh;
-        --g-modal-min-height: 100svh;
-        --g-modal-margin: 0px;
-        --g-modal-border-radius: 0px;
+        #{$block}__content {
+            margin: 0;
```
korvin89: 🤖 **AI generated**

🔴 **blocker** — dropping `--g-modal-margin: 0px` breaks `Dialog` through **inheritance** rather than specificity, so the new higher-specificity rule doesn't cover it. On mobile `Dialog` always renders both `mobile` and `has-scroll` (`Dialog.tsx:167,169`), and `.g-dialog_has-scroll` computes `max-height: calc(100dvh - var(--g-modal-margin, 20px) * 2)` (`Dialog.scss:40-42`). Previously the mobile overlay declared that variable as `0px` and it inherited down to `.g-dialog` → `100dvh`. With the variable gone it falls back to `20px`, so the dialog caps at `100dvh - 40px` while its parent `.g-modal__content` is `100dvh` and `.g-dialog_mobile` asks for `height: 100%`.

Measured at 390×844: `.g-dialog` `max-height` `844px → 804px`, height `844 → 804`, leaving a 40px strip of `--g-color-base-modal` at the bottom of every mobile dialog with the footer floating above the screen edge.

Declaring the variables on the content element itself would fix this and my other comment in one go — a declaration on the element always beats an inherited value, no matter the ancestor's specificity:

```scss
&_mobile {
    #{$block}__content {
        --g-modal-margin: 0px;
        --g-modal-border-radius: 0px;

        margin: 0;
        /* … the rest as in the PR */
    }
}
```

I verified this variant in the browser: it restores every value to the pre-PR state while keeping what the PR fixes (the `Dialog` `maxWidth="s"` cap no longer applies — content width at a 600px viewport goes `480 → 600`).

src/components/Modal/Modal.scss
```diff
@@ -86,11 +86,16 @@ $modal-transform-transition: transform $modal-transition-duration ease-out;
     }
 
     &_mobile {
-        --g-modal-width: 100vw;
-        --g-modal-height: 100dvh;
-        --g-modal-min-height: 100svh;
-        --g-modal-margin: 0px;
-        --g-modal-border-radius: 0px;
+        #{$block}__content {
```
korvin89: 🤖 **AI generated**

🟠 **should-fix** — this is an undocumented change to a public contract: in mobile mode all eight `--g-modal-*` variables listed under "CSS API" in `README.md:60-71` become no-ops, and the only way around them is `!important` or a hand-rolled 0-3-0 selector. Someone building a bottom-sheet look (`--g-modal-height: auto` plus a top radius) would lose it silently.

Two ways out, and I'd love to hear which you intend:
- document in `README.md` / `README-ru.md` that the mobile presentation ignores the CSS API; or
- take the narrower fix and scope the offending `Dialog` rules instead — `.g-dialog__modal:not(.g-modal_mobile):has(…)` in `Dialog.scss:13-29` — which addresses the actual bug while leaving the variable contract intact for everyone else.

src/components/Modal/Modal.scss
```diff
@@ -86,11 +86,16 @@ $modal-transform-transition: transform $modal-transition-duration ease-out;
...
+            margin: 0;
+            height: 100dvh;
+            min-height: 100svh;
+            max-height: initial;
+            width: 100vw;
+            min-width: initial;
```
korvin89: 🤖 **AI generated**

🔵 **nit** — `min-width: initial` computes to `auto`, not `0`, so it quietly undoes the deliberate `min-width: 0` from `_has-scroll` (line 51). I confirmed the computed value goes `0px → auto` here.

It happens to be harmless today only because `overflow: auto` zeroes a flex item's automatic minimum size (I checked with 800px-wide content — no horizontal overflow either way), but that's an accident rather than a guarantee. `min-width: 0` would state the intent directly.

src/components/Modal/Modal.scss
```diff
@@ -86,11 +86,16 @@ $modal-transform-transition: transform $modal-transition-duration ease-out;
...
-        --g-modal-width: 100vw;
-        --g-modal-height: 100dvh;
-        --g-modal-min-height: 100svh;
-        --g-modal-margin: 0px;
-        --g-modal-border-radius: 0px;
+        #{$block}__content {
+            margin: 0;
+            height: 100dvh;
+            min-height: 100svh;
+            max-height: initial;
```
korvin89: 🤖 **AI generated**

🔵 **nit** — `max-height: initial` here and `max-width: initial` on line 96 read as "reset something" rather than stating a value; the rest of this file uses concrete keywords. `none` is identical and a bit clearer.

Functionally these are fine: `height: 100dvh` and `width: 100vw` are definite, so dropping the `_has-scroll` viewport clamps changes nothing.

korvin89 (commented):
🤖 **AI generated**

**Severity legend**

korvin89:
🤖 **AI generated**

🟠 **should-fix** — the fixed behavior isn't covered by a test. The existing mobile case (`Modal.visual.test.tsx:99-120`) mounts a bare `Modal` at 390px, so it passes identically before and after this change: nothing sets `--g-modal-max-width` there, and 390 < 480 anyway.

Would you consider adding a regression test with the consumer in the loop — `MobileProvider mobile __experimentalMobileModals` plus `<Dialog maxWidth="s" fullWidth open>` at a viewport wider than 480px (600×900 works), asserting `contentClientWidth === overlayClientWidth`? That's exactly the case that goes `480 → 600` with this change.

A mobile `Dialog` screenshot would additionally have caught the two blockers I left inline. There's currently no mobile `Dialog` coverage at all — `Dialog.visual.test.tsx` only runs at 1000×600.

korvin89:
🤖 **AI generated**

🔵 **nit** — pre-existing, from #2798 rather than this PR, but it concerns exactly the properties being overridden here, so it seems worth mentioning while we're in this file: `min-height: var(--g-modal-min-height, none)` (`Modal.scss:37`) and `min-width: var(--g-modal-min-width, none)` (line 41) use `none`, which isn't a valid value for `min-*` (only for `max-*`). When the variable is unset, the substitution produces an invalid declaration and the property falls back to its initial value (`auto`), so the declaration is effectively dead. `auto` as the fallback would say the same thing validly.

src/components/Modal/Modal.scss
amje: Fixed

src/components/Modal/Modal.scss
amje: Fixed

src/components/Modal/Modal.scss
amje: Reverted this change

src/components/Modal/Modal.scss
amje: Removed these lines

amje:
Rethink this fix and address issues. Should work now

src/components/Dialog/Dialog.scss
```diff
@@ -10,7 +10,7 @@
     flex-direction: column;
     width: var(--g-dialog-width, var(--_--width));
 
-    &__modal {
+    &__modal:not(.#{globalVariables.$ns}modal_mobile) {
```
korvin89: 🤖 **AI generated**

🔵 **nit** — this hardcodes `Modal`'s internal BEM modifier (`.g-modal_mobile`) into `Dialog`'s stylesheet. If that modifier is ever renamed, `Dialog` would break silently — no build error, just the desktop width constraints leaking back into the mobile presentation.

`Dialog` already knows this state on its own (`mobile`, `Dialog.tsx:91-92`) and fully controls the modal's className, so it could emit its own modifier instead — `b('modal', {mobile}, modalClassName)` at `Dialog.tsx:156` — and scope on that: `&__modal:not(#{variables.$block}__modal_mobile)`. That keeps the coupling inside `Dialog`. Entirely your call though — the current version works.

korvin89 (approved):
🤖 **AI generated**

Thanks for the quick turnaround! I re-checked this revision in a browser against the real `Modal` → `Dialog` DOM, and all seven points from my earlier review are addressed:

- Both blockers are gone: at 390×844 `clip-path` is back to `inset(0px)`, and the mobile dialog fills the full `100dvh` again with no 40px strip at the bottom.
- The narrower `Dialog`-side scoping fixes the original bug — content width at a 600px viewport goes `480 → 600` — while leaving the `--g-modal-*` CSS API contract intact, so no docs change is needed.
- The new visual test is a genuine regression test. I ran its four assertions against `main`, the first commit and this head: it fails on `main` (`480` vs `600`), fails on the first commit (dialog height `860` vs `900`, `clip-path: inset(0px round 5px)`), and passes here.
- `min-height` / `min-width` now fall back to a valid `auto`, with the same computed result as before.

Approving. One optional nit left inline.

src/components/Dialog/Dialog.scss
amje: Fixed

PR #2802, opened 2026-08-28
feat(Button): add icon size context

amje:


korvin89:
🤖 **AI generated**

**Severity legend**

src/components/Button/types.ts
```diff
@@ -0,0 +1,76 @@
+import type * as React from 'react';
```
korvin89, line 1: 🤖 **AI generated**

🔴 **blocker** — `npm run typecheck` currently fails with 10 errors across 6 files, and the "Verify Files" CI job is red. Moving the prop types out of `Button.tsx` left the existing importers pointing at a module that no longer exports them:

- `src/components/theme/PrivateDefaultPropsProvider.tsx:13` — `ButtonCommonProps`. This one is production code, and note that `ButtonCommonProps` isn't re-exported from `Button/index.ts` either, so it needs `from '../Button/types'`.
- `src/components/Palette/__tests__/Palette.test.tsx:4` — `ButtonSize`
- `src/components/Button/__tests__/cases.ts:3` — `ButtonProps`
- `src/components/Button/__tests__/Button.visual.test.tsx:5` — `ButtonProps`
- `src/components/Button/__stories__/Button.stories.tsx:16` — `ButtonButtonProps`, `ButtonLinkProps`, `ButtonSize`
- `src/components/Button/__stories__/ButtonViewShowcase.tsx:6` — `ButtonProps`

The two `TS7053` errors (`Button.stories.tsx:22`, `ButtonViewShowcase.tsx:95`) are just fallout from the same root cause. Either adding `export type * from './types';` to `Button.tsx` or updating the importers would fix it.

Worth noting that `npx jest` stays green here because Babel strips the types, so the unit suite won't catch this class of breakage.

One more thing worth a conscious decision: `Button.tsx` no longer exports any types at all, so anyone deep-importing from `.../components/Button/Button` will break too. Might be worth a mention in the release notes.

src/components/Button/ButtonIcon.tsx
```diff
@@ -17,6 +20,19 @@ function warnAboutPhysicalValues() {
 }
 
 export const ButtonIcon = ({side, className, children}: Props) => {
+    const buttonIconSize = React.useContext(ButtonIconSizeContext);
+
+    const content =
+        buttonIconSize && isIcon(children)
```
korvin89: 🤖 **AI generated**

🟠 **should-fix** — the new sizing only covers half of what `Button` accepts as an icon. `prepareChildren` treats `isIcon(item) || isSvg(item)` as an icon and wraps both in `Button.Icon` (`Button.tsx:176` and `Button.tsx:190`), but `ButtonIcon` only resizes when `isIcon(children)` is true. As a result `<Button size="xl"><Icon …/></Button>` and `<Button size="xl"><svg …/></Button>` end up with different icon sizes, which is probably surprising given both are supported inputs.

Would it make sense to handle the `isSvg` case as well (setting `width`/`height`), or alternatively to document the limitation explicitly?

src/components/Button/ButtonIcon.tsx
```diff
@@ -1,8 +1,11 @@
-import type * as React from 'react';
+import * as React from 'react';
```
korvin89, line 3: 🤖 **AI generated**

🟠 **should-fix** — `'use client'` seems to be missing here and in the new `src/components/Button/ButtonIconSizeContext.ts`. `ButtonIcon` now calls `React.useContext`, and the new module calls `React.createContext` at module scope; both are re-exported through `Button/index.ts:13` → `components/index.ts:13`, neither of which carries the directive. Before this PR `ButtonIcon` was hook-free and rendered fine in an RSC server component.

The repo does follow this convention pretty consistently — of all the files calling `createContext`, only `Accordion/AccordionContext.tsx` and `lab/FileDropZone/FileDropZone.Provider.tsx` lack it, while `Alert/AlertContext.tsx`, `theme/ThemeContext.ts`, `lab/Menu/MenuContext.ts`, `Dialog/DialogPrivateContext.ts` and `DropdownMenu/DropdownMenuContext.ts` all declare it. There's no lint rule enforcing it, so CI won't flag this.

src/components/Button/__tests__/Button.test.tsx
```diff
@@ -102,6 +104,76 @@ describe('Button', () => {
         expect(button).toContainElement(iconComponent);
     });
 
+    test.each(new Array<ButtonSize>('xs', 's', 'm', 'l', 'xl'))(
```
korvin89: 🤖 **AI generated**

🟠 **should-fix** — the new tests only exercise the `<button>` render path. The provider is duplicated across three branches — `Button.tsx:111` (polymorphic `component`), `Button.tsx:127` (`href` / `<a>`) and `Button.tsx:144` (`<button>`) — and two of them aren't pinned by anything, so dropping the provider from either would go unnoticed.

A few other cases that are currently uncovered, if you'd like to add them:
- a raw `<svg>` child (see the related comment on `ButtonIcon.tsx`);
- `Button.Icon` rendered outside any `Button` (context is `null` → children should be left untouched);
- the end-side icon path (`Button.tsx:217-221`).

One more small thing: the expectations at lines 120, 143 and 235 read from the same `BUTTON_ICON_SIZE_MAP` the implementation uses, so a wrong map wouldn't be caught. Hardcoding at least one expected pixel value would make the test bite.

src/components/Button/Button.tsx
```diff
@@ -175,7 +108,9 @@ const _Button = React.forwardRef(function Button<T extends ButtonCustomElementTy
                 ref: ref,
                 'aria-disabled': disabled ?? undefined,
             },
-            prepareChildren(children),
+            <ButtonIconSizeContext.Provider value={BUTTON_ICON_SIZE_MAP[size]}>
```
korvin89: 🤖 **AI generated**

🔵 **nit** — the same provider block is copy-pasted three times (lines 111, 127 and 144). Hoisting it once above the branches would keep a future fourth render path from silently forgetting it:

```tsx
const content = (
    <ButtonIconSizeContext.Provider value={BUTTON_ICON_SIZE_MAP[size]}>
        {prepareChildren(children)}
    </ButtonIconSizeContext.Provider>
);
```

src/components/Button/ButtonIcon.tsx
```diff
@@ -17,6 +20,19 @@ function warnAboutPhysicalValues() {
 }
 
 export const ButtonIcon = ({side, className, children}: Props) => {
+    const buttonIconSize = React.useContext(ButtonIconSizeContext);
+
+    const content =
```
korvin89: 🤖 **AI generated**

🔵 **nit** — the nested `??` + ternary is a little hard to follow, and it clones even when there's nothing to inject: when `size` is already set it re-passes the same value, and when `width`/`height` are set it passes `size: undefined` — either way a fresh element is created on every render for no benefit. Also, `buttonIconSize &&` is a falsy check on something typed `number | null`.

Something like this might read better and skip the needless clone:

```tsx
const shouldSize =
    buttonIconSize !== null &&
    isIcon(children) &&
    children.props.size === undefined &&
    children.props.width === undefined &&
    children.props.height === undefined;

const content = shouldSize
    ? React.cloneElement(children, {size: buttonIconSize})
    : children;
```

src/components/Button/Button.tsx
```diff
@@ -3,87 +3,27 @@
...
-    PolymorphicCustomElementType,
-    PolymorphicOverloadProps,
-} from '../utils/polymorphic';
+import type {PolymorphicOverloadProps} from '../utils/polymorphic';
 import {isPolymorphicComponentProps} from '../utils/polymorphic';
 
 import {ButtonIcon, getIconSide} from './ButtonIcon';
-import type {BUTTON_VIEWS} from './constants';
+import {ButtonIconSizeContext} from './ButtonIconSizeContext';
+import {BUTTON_ICON_SIZE_MAP} from './constants';
```
korvin89, line 16: 🤖 **AI generated**

🔵 **nit** — now that the context handles this, the manual `BUTTON_ICON_SIZE_MAP[size]` lookups for icons that live inside a `Button` are redundant: `src/components/lab/Menu/MenuTrigger.tsx:23`, `src/components/lab/Menu/MenuItem.tsx:115`, `src/components/Button/__stories__/Button.stories.tsx:22`, `src/components/Button/__stories__/ButtonViewShowcase.tsx:95`.

Leaving them means the codebase keeps demonstrating exactly the pattern this PR is trying to remove — and the two story files are already in the list of files that need touching for the typecheck fix anyway.

src/components/Button/ButtonIcon.tsx
```diff
@@ -17,6 +20,19 @@ function warnAboutPhysicalValues() {
...
+        buttonIconSize && isIcon(children)
+            ? React.cloneElement(children, {
```
korvin89: 🤖 **AI generated**

❓ **question** — what's the intended precedence relative to `DefaultPropsProvider`? `Icon` calls `useDefaultProps('Icon', …)` (`Icon.tsx:53`), and `useDefaultProps` gives explicitly-passed props priority over the configured defaults. So a global `defaultProps={{Icon: {size: 24}}}` used to apply to icons inside buttons and is now silently overridden by the button-derived size. Is "Button wins" the deliberate choice here?

src/components/Button/ButtonIcon.tsx
```diff
@@ -1,8 +1,11 @@
-import type * as React from 'react';
+import * as React from 'react';
 
 import {block} from '../utils/cn';
+import {isIcon} from '../utils/common';
```
korvin89: 🤖 **AI generated**

❓ **question** — `isIcon` is `isOfType(Icon)` (`utils/common.ts:13`), which matches not just by identity but also by `displayName === 'Icon'` (`isOfType.ts:20-21`). There's an example right in this repo: `Button.stories.tsx:24` sets `IconWithSize.displayName = 'Icon'` on a component whose `size` prop is a `ButtonSize` (`'xs' | 's' | …`), not a number.

That means `Button` will now inject a numeric `size` into arbitrary user components that merely share the display name, with no type checking involved. Is that an acceptable risk, or would an identity-only check be safer in this particular spot?

src/components/Button/ButtonIcon.tsx
```diff
@@ -17,6 +20,19 @@ function warnAboutPhysicalValues() {
 }
 
 export const ButtonIcon = ({side, className, children}: Props) => {
+    const buttonIconSize = React.useContext(ButtonIconSizeContext);
```
korvin89, line 33: 🤖 **AI generated**

❓ **question** — `ClipboardButton` keeps its own map at `src/components/ClipboardButton/ClipboardButton.tsx:44`: `{xs: 12, s: 16, m: 16, l: 16, xl: 20}`, which diverges from `BUTTON_ICON_SIZE_MAP` at `s` (16 vs 14). Since the context isn't exported and `ClipboardIcon` isn't an `Icon`, the built-in icon stays 16 while a user-supplied `icon={<Icon …/>}` at the same `size="s"` will now get 14.

Would it make sense to reconcile the two maps, and/or to export the context (or a small `useButtonIconSize` hook) so custom icon wrappers can participate in the same sizing?

src/components/Button/ButtonIcon.tsx
```diff
@@ -17,6 +20,19 @@ function warnAboutPhysicalValues() {
...
+
+    const content =
+        buttonIconSize && isIcon(children)
+            ? React.cloneElement(children, {
+                  size:
+                      children.props.size ??
+                      (children.props.width === undefined && children.props.height === undefined
```
korvin89: 🤖 **AI generated**

❓ **question** — when only `width` is given (no `height`, no `size`), the code passes `size: undefined`, so the height falls back to the SVG `viewBox` rather than the button-derived size. Would passing `size: buttonIconSize` and letting `width`/`height` override the relevant axis be closer to what you want? That way the unspecified dimension stays button-sized.

korvin89:
🤖 **AI generated**

🟠 **should-fix** — (this one is about files outside the diff, so I couldn't leave it inline)

The docs haven't been updated for the new behavior. The whole "Icons" section — `src/components/Button/README.md:173-221` and `src/components/Button/README-ru.md:173-221` — still teaches `<Icon data={Gear} size={18} />`, which now actively opts *out* of the automatic sizing this PR adds, and the automatic sizing itself isn't documented anywhere.

Since getting rid of that manual `size` is the point of the change, it would be great to update the examples and add a short note (the props table row for `children` at `README.md:435` might be a good spot too).

src/components/Button/types.ts
amje: Fixed

src/components/Button/ButtonIcon.tsx
amje: Fixed

src/components/Button/ButtonIcon.tsx
amje: Fixed

src/components/Button/__tests__/Button.test.tsx
amje: Added more tests

src/components/Button/Button.tsx
amje: Done

src/components/Button/ButtonIcon.tsx
amje: Fixed

src/components/Button/Button.tsx
amje: Fixed all except `src/components/lab/Menu/MenuItem.tsx:115`. There `Icon` is not a child of the `Button`

src/components/Button/ButtonIcon.tsx
amje: The order of applying size is now:
1. Local `Icon`'s size/width/height props
2. Size from `Button`
3. Default prop for `Icon`

src/components/Button/ButtonIcon.tsx
amje: Added an option to disable displayName check in the helper and use it for ButtonIcon

src/components/Button/ButtonIcon.tsx
amje: Removed own map for ClipboardButton, the component now uses common map

src/components/Button/Button.tsx
korvin89: 🤖 **AI generated**

Agreed, and thanks for checking that one — you're right, `MenuItem.tsx:115` renders the arrow into `<div className={b('arrow')}>` rather than a `Button.Icon`, so the context never reaches it and the explicit map is the correct thing there. Same goes for the `lab/Menu` stories. My original list was too broad on that point. Resolving.

src/components/Button/ButtonIcon.tsx
korvin89: 🤖 **AI generated**

Thanks for spelling out the order — that reads well, and `should prefer button icon size over Icon default size` pins it, so a future change to `useDefaultProps` can't silently flip the precedence. Resolving.

src/components/Button/ButtonIcon.tsx
korvin89: 🤖 **AI generated**

This one looks addressed in `0f6cdc1d` even though the thread wasn't updated: the condition became `width === undefined || height === undefined`, so an icon with only `width` now takes the button-derived size for its height, and `should use button size for unspecified icon dimension` covers both directions. Exactly what I was after — resolving.

korvin89:
🤖 **AI generated**

❓ **question** — is the current red "Visual Tests" run expected, or still to be regenerated?

`0f6cdc1d` updates the DropdownMenu (`s`, `xl`) and Toaster snapshots, but the run on that commit now fails on 5 different ones — all `src/components/controls/PasswordInput/__tests__/PasswordInput.visual.test.tsx` (`smoke`, `smoke with error`, `smoke inside error placement tooltip`, `smoke reveal button tooltip`, `smoke copy button tooltip`). Other branches built around the same time pass the suite, so it does look specific to this PR.

The trigger appears to be the `ClipboardButton` map unification: `BUTTON_ICON_SIZE_MAP.s` is `14` where the old local `ButtonSizeToIconSize.s` was `16`. `PasswordInput` at its default `size="m"` passes `actionButtonSize: 's'` to both action buttons (`src/components/controls/PasswordInput/utils.ts`), so the copy icon goes 16 → 14 while the reveal icon next to it stays at the hardcoded `iconSize = 16` from the same helper.

So there are arguably two things here — the snapshots to regenerate, and the fact that the two icons in that control no longer match. Would it make sense to bring `getActionButtonSizeAndIconSize` in line with `BUTTON_ICON_SIZE_MAP` (or drop its `iconSize` entirely now that `Button` derives it), and regenerate afterwards? Happy to be told this is intentional and just needs the snapshot refresh.

src/components/Button/ButtonIcon.tsx
amje: Yes

src/components/Button/Button.tsx
```diff
@@ -3,87 +3,27 @@
...
+import type {
+    ButtonButtonProps,
+    ButtonComponentProps,
+    ButtonCustomElementType,
+    ButtonLinkProps,
+    ButtonProps,
+} from './types';
```
ogonkov, line 23: That would be nice to have types moving in separate PR, to keep changes here more focused

amje: What the point of making it separate? It's a part of needed changes

amje:
> 🤖 **AI generated**
> 
> ❓ **question** — is the current red "Visual Tests" run expected, or still to be regenerated?
> 
> `0f6cdc1d` updates the DropdownMenu (`s`, `xl`) and Toaster snapshots, but the run on that commit now fails on 5 different ones — all `src/components/controls/PasswordInput/__tests__/PasswordInput.visual.test.tsx` (`smoke`, `smoke with error`, `smoke inside error placement tooltip`, `smoke reveal button tooltip`, `smoke copy button tooltip`). Other branches built around the same time pass the suite, so it does look specific to this PR.
> 
> The trigger appears to be the `ClipboardButton` map unification: `BUTTON_ICON_SIZE_MAP.s` is `14` where the old local `ButtonSizeToIconSize.s` was `16`. `PasswordInput` at its default `size="m"` passes `actionButtonSize: 's'` to both action buttons (`src/components/controls/PasswordInput/utils.ts`), so the copy icon goes 16 → 14 while the reveal icon next to it stays at the hardcoded `iconSize = 16` from the same helper.
> 
> So there are arguably two things here — the snapshots to regenerate, and the fact that the two icons in that control no longer match. Would it make sense to bring `getActionButtonSizeAndIconSize` in line with `BUTTON_ICON_SIZE_MAP` (or drop its `iconSize` entirely now that `Button` derives it), and regenerate afterwards? Happy to be told this is intentional and just needs the snapshot refresh.

Fixed PasswordInput view and update tests

korvin89 (approved):

src/components/Button/Button.tsx
ogonkov: You got changes that relate only to changing imports, that is not directly relaty to primary goal of PR

amje: So? It's a normal refactoring while adding a new feature. Changing imports does not affect public API at all. It doesn't matter whether it would be a separate commit or within the current commit. I see no harm to do so

src/components/Button/ButtonIcon.tsx
```diff
@@ -17,6 +25,27 @@ function warnAboutPhysicalValues() {
 }
 
 export const ButtonIcon = ({side, className, children}: Props) => {
+    const buttonIconSize = React.useContext(ButtonIconSizeContext);
+
+    let content = children;
+
+    if (buttonIconSize !== null) {
+        if (
+            isIcon(children) &&
```
ogonkov, line 40: user explicitly use `<Button.Icon>` for components that he believe is icon, i think we should try resize all children anyway, even if doesn't have `Icon` name

src/components/Button/ButtonIcon.tsx
```diff
@@ -17,6 +25,27 @@ function warnAboutPhysicalValues() {
...
+        ) {
+            content = React.cloneElement(children, {size: buttonIconSize});
+        } else if (isSvg(children)) {
+            const width = children.props.width ?? buttonIconSize;
+            const height = children.props.height ?? buttonIconSize;
+
+            if (width !== children.props.width || height !== children.props.height) {
+                content = React.cloneElement(children, {width, height});
+            }
+        }
```
ogonkov, line 52: May be we can add clone for children with width/height if we not sure what type of component user is used?

src/components/Button/ButtonIcon.tsx
amje: Implemented render function for custom icons

src/components/Button/ButtonIcon.tsx
amje: Implemented render function for custom icons

src/components/Button/__stories__/ButtonViewShowcase.tsx
```diff
@@ -89,11 +88,7 @@ export function ButtonViewShowcase(args: ButtonProps) {
                         className={b('grid-cell')}
                     >
                         <Button {...props}>
-                            <Icon
-                                key="icon"
-                                data={Gear}
-                                size={BUTTON_ICON_SIZE_MAP[args.size || 'm']}
-                            />
+                            <Icon key="icon" data={Gear} />
```
ogonkov: not blocker but i guess `key` is redundant here

src/components/ClipboardButton/ClipboardButton.tsx
```diff
@@ -62,7 +60,7 @@ const ClipboardButtonComponent = (props: ClipboardButtonComponentProps) => {
 
     const buttonIcon = (
         <Button.Icon className={b('icon')}>
-            {icon ?? <ClipboardIcon size={ButtonSizeToIconSize[size]} status={status} />}
+            {icon ?? <ClipboardIcon size={BUTTON_ICON_SIZE_MAP[size]} status={status} />}
```
ogonkov: can we use signature with children as function here?

src/components/lab/Menu/MenuTrigger.tsx
```diff
@@ -18,10 +17,7 @@ export const MenuTrigger = React.forwardRef<HTMLButtonElement, MenuTriggerProps>
                 {children ? (
                     children
                 ) : (
-                    <Icon
-                        data={icon === 'vertical' ? EllipsisVertical : Ellipsis}
-                        size={BUTTON_ICON_SIZE_MAP[size]}
```
ogonkov, line 23: why icon size changed on screenshots, if we already got correct sizes here?

amje: Screenshots are changed for DropdownMenu, this is different component

src/components/Button/__stories__/ButtonViewShowcase.tsx
amje: Removed key

src/components/ClipboardButton/ClipboardButton.tsx
amje: Used function

PR #2809, opened 2026-09-01
chore: improve harness

NikitaCG:
This PR improves the repository harness and documentation for `@gravity-ui/uikit`.

  The updated harness directs agents to authoritative documentation before source code, defines repository invariants, and
  introduces a change-aware validation policy based on the npm scripts declared in `package.json`.

  The repository-overview eval shows that the updated harness preserves answer quality while substantially reducing token
  usage, execution time, tool calls, and repository exploration.

  ## Workflow

  ### 1. Install the eval skill

  Add the Gravity UI skills marketplace:

  ```bash
  codex plugin marketplace add gravity-ui/skills-dev

  Then:

  1. Open /plugins.
  2. Select the Gravity UI marketplace.
  3. Install gravity-ui-dev-skills.
  4. Start a new Codex session so the installed skills are available.

  Initialize the portable repository-overview eval using the repository-overview-eval-init skill.

  The eval is installed at:

  .agents/evals/repository-overview/

  Generated artifacts are stored under the ignored directory:

  .agents/evals/repository-overview/.eval-artifacts/

  ### 2. Run the baseline

  Run three attempts before changing the repository harness:

  node .agents/evals/repository-overview/run.mjs \
    --host codex \
    --repo . \
    --model gpt-5.6-sol \
    --effort high \
    --repeat 3

  Baseline artifact:

  .agents/evals/repository-overview/.eval-artifacts/runs/2026-09-01T10-59-42.171Z-codex

  ### 3. Improve the harness

  The following changes were made:

  - Reworked AGENTS.md into a compact, English-only repository harness.
  - Added a documentation-first pre-response gate:
      1. Read docs/navigation.md.
      2. Read docs/architecture.md.
      3. Read the relevant domain guide or component README.
      4. Inspect source code only after consulting the documentation.

  - Added explicit repository invariants for:
      - public imports and entrypoints;
      - stable, legacy, and unstable APIs;
      - component co-location;
      - semantic design tokens;
      - RTL and i18n;
      - npm-only workflows.

  - Added change-aware validation rules.
  - Required agents to use npm scripts from package.json so local validation remains consistent with CI.
  - Prohibited ad hoc shell, Node.js, or one-off validators when repository scripts are available.
  - Added English and Russian architecture and navigation guides.
  - Added missing Russian layout and typography guides.
  - Expanded the Russian theming guide.
  - Updated the English and Russian README documentation indexes.
  - Excluded .agents/ from ESLint and Prettier while keeping the portable eval available for version control.

  ### 4. Run the candidate

  Run the same eval again after changing the harness.

  The host, model, effort, CLI version, scenario, runner, and repeat count must remain identical:

  node .agents/evals/repository-overview/run.mjs \
    --host codex \
    --repo . \
    --model gpt-5.6-sol \
    --effort high \
    --repeat 3

  Candidate artifact:

  .agents/evals/repository-overview/.eval-artifacts/runs/2026-09-01T13-30-20.917Z-codex

  ### 5. Compare the results

  node .agents/evals/repository-overview/compare.mjs \
    --baseline .agents/evals/repository-overview/.eval-artifacts/runs/2026-09-01T10-59-42.171Z-codex \
    --candidate .agents/evals/repository-overview/.eval-artifacts/runs/2026-09-01T13-30-20.917Z-codex

  ## Evaluation environment

  The baseline and candidate used the same environment:

  - Host: codex
  - Model: gpt-5.6-sol
  - Effort: high
  - CLI: codex-cli 0.152.0
  - Scenario version: 5
  - Runner version: 4
  - Attempts: 3 per run

  ## Results

   Metric                         Baseline     Candidate         Change
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━  ━━━━━━━━━━━━  ━━━━━━━━━━━━━
   Successful attempts                 3/3           3/3      unchanged
  ───────────────────────────  ────────────  ────────────  ─────────────
   Quality score                       1.0           1.0      unchanged
  ───────────────────────────  ────────────  ────────────  ─────────────
   Route score                         1.0           1.0      unchanged
  ───────────────────────────  ────────────  ────────────  ─────────────
   Median tokens                   396,065       210,364         −46.9%
  ───────────────────────────  ────────────  ────────────  ─────────────
   Median duration              168,820 ms    137,466 ms         −18.6%
  ───────────────────────────  ────────────  ────────────  ─────────────
   Median duration, readable          2:49          2:17    −32 seconds
  ───────────────────────────  ────────────  ────────────  ─────────────
   Median tool calls                    26            19         −26.9%
  ───────────────────────────  ────────────  ────────────  ─────────────
   Median referenced paths              68            33         −51.5%
  ───────────────────────────  ────────────  ────────────  ─────────────
   Median failed tool calls              1             0          −100%
  ───────────────────────────  ────────────  ────────────  ─────────────
   Repository mutations                  0             0      unchanged

  The token and duration ranges also did not overlap:

  - Baseline tokens: 332,867–531,630
  - Candidate tokens: 167,625–237,994
  - Baseline duration: 160,511–194,145 ms
  - Candidate duration: 106,122–146,164 ms

  The tool-call and referenced-path ranges did not overlap either.

  ## Navigation behavior

  All three candidate attempts followed the intended documentation-first route:

  1. docs/navigation.md
  2. docs/architecture.md
  3. relevant repository documentation
  4. package and source files

  All candidate responses also included the required Consulted docs section.

  The baseline attempts explored the repository more broadly and repeatedly revisited package, source, configuration, and
  documentation files.

  ## Conclusion

  The new harness produced a clear measurable improvement:

  - answer quality remained unchanged at the maximum score;
  - token usage decreased by approximately 47%;
  - execution time decreased by approximately 19%;
  - tool usage decreased by approximately 27%;
  - repository exploration was reduced by approximately 52%;
  - failed tool calls were eliminated;
  - all attempts remained read-only;
  - agents consistently followed the intended documentation-first route.

  This indicates that the new navigation and architecture documentation gives agents enough authoritative context to
  understand the repository without broad source exploration.


korvin89:
🤖 **AI generated**

**Severity legend**

docs/navigation.md
```diff
@@ -0,0 +1,62 @@
+# Repository navigation
```
korvin89: 🤖 **AI generated**

🔴 **blocker**

I think these new guides end up in the published npm tarball, which probably isn't intended. `gulpfile.js:109` runs `copy-docs` → `utils.buildDocs()`, whose default config collects **every** `*.md` under `docs/` into `build/docs/guides/` and lists it in `build/docs/INDEX.md` — and `package.json` publishes `build`.

I ran `buildDocs()` against this branch to confirm: `guides/navigation.md`, `guides/navigation-ru.md`, `guides/architecture.md` and `guides/architecture-ru.md` are all generated, and the English ones appear in `INDEX.md`. That's the file `README.md:168` points agents at inside consumer projects (`node_modules/@gravity-ui/uikit/build/docs/INDEX.md`), so an agent working in someone else's app would be told to "Prepare a contribution", run `npm run lint:styles`, inspect `src/components/…`, and never update snapshots without approval.

There's a second symptom: `rewriteDocLinks` unwraps links to files that aren't shipped, so in the tarball the contribution row degrades to plain text — `README development setup, contributor overview, CLA, and PR guide` — with no links at all.

Would you consider moving these two guides out of `docs/`? `contribute/` isn't published, so `contribute/navigation.md` / `contribute/architecture.md` would solve it. Passing a custom config with an `exclude` to `buildDocs` would work too.

docs/layout.md
```diff
@@ -1,5 +1,7 @@
 # Layout components and spacings
 
+[English](layout.md) | [Русский](layout-ru.md)
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

The language switcher becomes the document's first paragraph, and `buildDocs` uses that paragraph as the guide's description in the published `build/docs/INDEX.md`.

I built the docs on both `main` and this branch to compare. Before:

```
- [Layout components and spacings](./guides/layout.md) — This guide covers UIKit's layout foundations: a shared spacing scale (--g-spacing-*, used everywhere via tokens and props) and a responsive grid (Container/Row/Col)…
```

After:

```
- [Layout components and spacings](./guides/layout.md) — English | Русский
```

`docs/typography.md:3` has the same effect. These are two of the three guides `README.md` lists under "Useful docs", so the index gets noticeably less useful for consumer-side agents.

Would it work to put the switcher after the intro paragraph instead? That keeps the description intact while still offering the language links.

.agents/evals/repository-overview/scenario.json
```diff
@@ -0,0 +1,71 @@
...
+  "id": "repository-overview",
+  "version": 5,
+  "prompt": "Explain this repository: what applications or packages it contains, how its architecture is organized, which technology stack it uses, and which important invariants developers must follow. Inspect the repository read-only. Do not use the eval itself or previous artifacts in .agents/evals/repository-overview as sources of information about the repository. Support substantive claims with relative file paths. Structure the answer using the sections \"Applications and packages\", \"Architecture\", \"Technology stack\", and \"Invariants\".",
+  "assertions": [
+    {
+      "id": "answer-applications",
+      "group": "quality",
+      "type": "answer-regex",
+      "description": "The answer describes applications or packages",
+      "pattern": "(?:приложени|пакет|applications?|packages?)"
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

I think the quality assertions can't fail, which would mean "Quality score 1.0 unchanged" doesn't tell us much about answer quality.

The prompt on line 4 already requires the answer to use the sections "Applications and packages", "Architecture", "Technology stack" and "Invariants". The four quality assertions then regex for `applications?|packages?`, `architecture`, `stack` and `invariants?` — exactly the headings the prompt mandates. `answer-cites-sources` only needs two real paths, which any answer mentioning `package.json` and `README.md` satisfies. So any structurally compliant answer scores 1.0 regardless of whether its content is correct.

That matters here specifically: the main risk behind a −47% token drop is that the agent stopped reading enough to stay accurate, and this is the metric that would have caught it.

`compare.mjs` agrees, incidentally — `bothScoresPerfect()` emits "The assertions did not discriminate between runs" with a `stronger-assertions` recommendation whenever both sides score 1.0, so the attached `comparison.md` should already carry that warning.

The eval's own `README.md` ("Adapt assertions to a repository") suggests the fix — repository-specific facts. For uikit, something like the `unstable_` naming convention, `--g-color-private-*` being the implementation layer, `files: ["build", "styles"]`, or the four built-in themes would make the score mean something.

.agents/evals/repository-overview/scenario.json
```diff
@@ -0,0 +1,71 @@
...
+    },
+    {
+      "id": "answer-cites-sources",
+      "group": "quality",
+      "type": "answer-existing-paths",
+      "description": "The answer cites at least two real repository files",
+      "minimum": 2
+    },
+    {
+      "id": "route-readme",
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

There's no route assertion covering the documentation-first route this PR introduces. The `route` group only checks the root `README` and a package manifest, plus that the eval's own directory isn't used as a source — and both the README and `package.json` are visited by the baseline and the candidate alike.

That means "Route score 1.0 unchanged" is equally compatible with the candidate never opening the new docs, so it doesn't back up "All three candidate attempts followed the intended documentation-first route" from the description. The recorded path sequences do show it — it just isn't asserted, so nothing would catch a regression later.

I'd suggest adding `trace-path-if-present` assertions for `^docs/navigation\.md$` and `^docs/architecture\.md$`, and possibly `max-tool-calls` / `max-path-revisits` budgets, so the route score actually measures the behavior the harness is meant to produce.

.agents/evals/repository-overview/compare.mjs
```diff
@@ -0,0 +1,409 @@
...
+      recommendations: experimentRecommendations.sort((left, right) => priorityRank(left.priority) - priorityRank(right.priority)),
+    },
+  };
+}
+
+function runSummary(result) {
+  const attempts = result.attempts ?? [];
+  const total = attempts.length || result.repeat || 0;
+  const successful = attempts.length
+    ? attempts.filter((attempt) => attempt.success !== false && attempt.metrics?.degraded !== true && (attempt.metrics?.hostErrorCount ?? 0) === 0).length
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

`runSummary()` reads `attempt.metrics?.degraded`, but `run.mjs:140` writes `degraded` at the attempt root — `metrics` only carries `malformedJsonLines` and `hostErrorCount`. The clause is therefore always true and the degraded check never fires.

I reproduced it with synthetic `result.json` fixtures: three candidate attempts with `degraded: true` and `malformedJsonLines: 42` (the parser lost events, so tokens and tool calls are undercounted) still produce `status: improved`, `experiment.validity: valid`, and no experiment findings. Reading `attempt.degraded` instead turns the same input into `inconclusive` / `invalid`.

Suggested fix:

```js
const successful = attempts.length
  ? attempts.filter((attempt) =>
      attempt.success !== false
      && attempt.degraded !== true
      && (attempt.metrics?.hostErrorCount ?? 0) === 0
      && (attempt.metrics?.malformedJsonLines ?? 0) === 0,
    ).length
  : result.aggregate?.successfulAttempts ?? total;
```

.agents/evals/repository-overview/lib/codex.mjs
```diff
@@ -0,0 +1,44 @@
...
+export function normalize(events) {
+  let response = '';
+  let usage = normalizeUsage();
+  let reportedCostUsd = null;
+  const trace = [];
+  const hostErrors = [];
+  for (const event of events) {
+    const item = event.item ?? {};
+    if (event.type === 'item.completed' && item.type === 'agent_message') {
+      response = item.text ?? textFromContent(item.content) ?? response;
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

The `?? response` fallback here can't be reached: `textFromContent()` (`lib/shared.mjs:18`) always returns a string — `''` for anything that is neither a string nor an array — so it's never nullish.

The effect is that an `agent_message` event without `text` and without an array `content` overwrites the answer collected so far with an empty string, after which `run.mjs:120` marks the attempt unsuccessful because `response.trim()` is falsy. Quick repro:

```js
normalize([
  {type: 'item.completed', item: {type: 'agent_message', text: 'real answer'}},
  {type: 'item.completed', item: {type: 'agent_message'}},
]).response // => ""
```

Switching `??` to `||` would make the fallback behave the way it reads.

.agents/evals/repository-overview/lib/opencode.mjs
```diff
@@ -0,0 +1,39 @@
...
+        part.state?.input ?? part.input ?? event.input,
+        part.state?.output ?? part.output,
+        null,
+        {status: part.state?.status ?? null, exitCode: null},
+      ));
+      if (part.state?.status === 'error') hostErrors.push(part.state.error ?? `OpenCode tool ${part.tool ?? part.name} failed`);
+    }
+    if (event.type === 'error') hostErrors.push(event.error?.message ?? event.message ?? 'Unknown OpenCode error');
+    if (event.type === 'step_finish' || part.type === 'step-finish') {
+      usage = addUsage(usage, normalizeUsage(part.tokens ?? event.tokens ?? event.usage));
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

The disjoint-counter correction exists only for the Claude adapter (`lib/claude.mjs:56-68`); OpenCode goes straight through `normalizeUsage`.

OpenCode reports `tokens` as `{input, output, cache: {read, write}}`, and for Anthropic models `input` excludes the cached tokens. In that case `total = input + output` (`lib/shared.mjs:57`) undercounts by the entire cache, and `estimateCost` (`lib/shared.mjs:214`) clamps `cached` to `Math.min(cached, input)` and drives `uncached` to zero, so the dollar estimate is off as well.

Related: `aggregate.uncachedInputTokens` (`run.mjs:187`) subtracts only `cachedInput`, so for the Claude adapter — where `input` was redefined as uncached + cache-read + cache-creation — cache-creation tokens get reported as "uncached".

It might be cleaner to move the normalization into a single helper with an explicit "does `input` already include the cache" flag and apply it for every host.

AGENTS.md
```diff
@@ -1,15 +1,77 @@
...
+  ultimately the package entrypoint.
+- Use semantic `--g-color-*` tokens in component and application styles; private palette tokens are
+  implementation details. Prefer spacing, typography, and radius tokens over hard-coded values.
+- Preserve RTL and i18n behavior: use logical CSS properties where direction matters and the
+  existing configuration/i18n patterns for built-in text.
+- Use npm only. The supported local baseline is Node.js 20.19+ and npm 9+; restore dependencies
+  with `npm ci` when required checks cannot run because the installation is incomplete.
+- Do not change public exports or TypeScript API shape unless the task explicitly requires it.
+
+## Testing and validation
```
korvin89, line 49: 🤖 **AI generated**

🟠 **should-fix**

The validation policy is now stated three times — here, in `docs/navigation.md:38`, and in `docs/navigation-ru.md` — in near-identical wording, and the copies have already drifted within this PR: `docs/navigation.md` names the scripts explicitly (`npm run lint`, `npm test`) and adds a paragraph about restoring dependencies with `npm ci`, neither of which appears here.

Since `AGENTS.md:8` already routes readers to `docs/navigation.md`, would it make sense to keep the gate and the invariants here and link to the validation matrix rather than restating it? That way there's a single place to update.

README.md
```diff
@@ -66,7 +66,14 @@ A SCSS [mixins](styles/mixins.scss) file with useful helpers is also available f
 
 Read more:
 
+- [Repository navigation](docs/navigation.md) — find the authoritative docs, source area, and
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

These two entries sit in the "Guides" section that follows "Usage" and "Styles", which reads as the section for people consuming the package from npm. "Repository navigation — find the authoritative docs, source area, and validation for a task" and "Repository architecture — … build, and tests" are contributor-facing, so they feel a little out of place there (`README-ru.md:69` has the same pairing).

Combined with the tarball issue in the other comment, I'd suggest linking them from `AGENTS.md` and `contribute/README.md` and leaving the domain guides here.

.agents/evals/repository-overview/run.mjs
```diff
@@ -0,0 +1,286 @@
+#!/usr/bin/env node
```
korvin89: 🤖 **AI generated**

❓ **question**

Would you consider landing the eval harness as its own PR and keeping this one to the documentation and `AGENTS.md` changes?

A few things prompted the question:

- Nothing in the repository references the directory except the three ignore entries this PR adds to `eslint.config.mjs`, `.prettierignore` and `.gitignore`. No workflow runs it and there's no npm script, so a contributor is unlikely to discover it.
- The measurement itself isn't committed — `.eval-artifacts/` is gitignored, so the repository gets the instrument but no reproducible record of the run, and a re-run on a different model version won't reproduce the numbers either.
- It's ~1150 lines of JavaScript carved out of ESLint and Prettier with no tests, and this review already surfaced three real defects in it (the other comments on `compare.mjs`, `lib/codex.mjs` and `lib/opencode.mjs`).
- It's a copy of the upstream harness from `gravity-ui/skills-dev`, installable on demand via the `repository-overview-eval-init` skill, so the vendored copy will drift with no update path.

The documentation half stands on its own and I'd be glad to see it land — I checked the new guides against the repository and the paths, tokens, entrypoints, build steps and test configuration they describe all match, and the Russian pairs line up with the English originals heading for heading and code block for code block.

If the harness should live here long-term, a follow-up PR could tailor `scenario.json` to uikit, run it through lint and Prettier, add an npm script plus a line in `navigation.md` for discoverability, and commit a baseline artifact or the comparison report.

.agents/evals/repository-overview/scenario.json
```diff
@@ -0,0 +1,71 @@
+{
+  "id": "repository-overview",
+  "version": 5,
+  "prompt": "Explain this repository: what applications or packages it contains, how its architecture is organized, which technology stack it uses, and which important invariants developers must follow. Inspect the repository read-only. Do not use the eval itself or previous artifacts in .agents/evals/repository-overview as sources of information about the repository. Support substantive claims with relative file paths. Structure the answer using the sections \"Applications and packages\", \"Architecture\", \"Technology stack\", and \"Invariants\".",
```
korvin89: 🤖 **AI generated**

❓ **question**

How much do you think the result generalizes beyond this one scenario?

`repository-overview` asks precisely the question `docs/navigation.md` and `docs/architecture.md` were written to answer, so a large token drop on it is close to expected. For a narrow task — "fix a bug in Select", say — the pre-response gate in `AGENTS.md:17` instead mandates two extra document reads before any source file is opened, which pushes in the opposite direction.

Would it be worth adding one narrowly scoped scenario before treating the gate as a net win? Happy to be wrong here if you've already looked at focused tasks separately.

AGENTS.md
```diff
@@ -1,15 +1,77 @@
-## How should the AI use the Gravity UI design system?
+# UIKit repository guide
```
korvin89, line 1: 🤖 **AI generated**

❓ **question**

Was dropping the "Example prompts for AI tools (Builder.io, Cursor, etc.)" section, along with "Usage in other repos" and "Naming across repos", intentional? The description doesn't mention it.

I don't think anything is actually lost: `AGENTS.md` isn't published (`files` is `["build", "styles"]`), and `README.md:135` covers the consumer-facing guidance in its "For AI agents" section. But the file's audience effectively shifts from "an agent working in a consumer project" to "an agent contributing to this repository", which seems worth calling out in the description in case anything external links to those examples.

AGENTS.md
```diff
@@ -1,15 +1,77 @@
...
+  existing configuration/i18n patterns for built-in text.
+- Use npm only. The supported local baseline is Node.js 20.19+ and npm 9+; restore dependencies
+  with `npm ci` when required checks cannot run because the installation is incomplete.
+- Do not change public exports or TypeScript API shape unless the task explicitly requires it.
+
+## Testing and validation
+
+Use a change-aware fast loop and require every applicable check to pass:
+
+- Use the applicable npm scripts declared in `package.json` as the validation entrypoints. They
```
korvin89: 🤖 **AI generated**

🔵 **nit**

Small tension inside this section: the first bullet rules out "ad hoc shell, Node.js, or other one-off validators in place of repository scripts", while the bullets right after ask for `git diff --check`, ESLint on changed files, and exact co-located Jest files — none of which are npm scripts. This PR also adds `node .agents/evals/repository-overview/run.mjs` and an `EVAL.md` that tells agents to run it.

Maybe rephrase along the lines of "prefer the npm scripts; targeted `eslint <file>` / `jest <file>` invocations are fine as a faster subset"?

package-lock.json
```diff
@@ -14835,9 +14835,9 @@
       "license": "BSD-3-Clause"
     },
     "node_modules/ignore": {
-      "version": "7.0.7",
-      "resolved": "https://registry.npmjs.org/ignore/-/ignore-7.0.7.tgz",
-      "integrity": "sha512-dML0wP6oak21rsNYCJpJB6O1BJIEwNpGrTw0URPfAk4hm0e3pRfCtzkfB6olBcXcVlU2rouCyz7lCyRB0OMVCA==",
+      "version": "7.0.8",
```
korvin89, line 14838: 🤖 **AI generated**

🔵 **nit**

There's an unrelated `ignore` 7.0.7 → 7.0.8 bump in the lockfile. Given the recent `chore: fix package lock (#2805)` on main, it's probably worth dropping to keep the diff focused.

docs/layout.md
NikitaCG: good point ,fix it

docs/navigation.md
NikitaCG: done, fix it

AGENTS.md
NikitaCG: Yes, this was intentional. AGENTS.md is now a repository-local harness
  > for agents contributing to UIKit rather than guidance for agents
  > consuming the npm package

AGENTS.md
NikitaCG: fix it

AGENTS.md
NikitaCG: fix

README.md
NikitaCG: fix

NikitaCG:
I pick all comments for eval and I want to improve eval in another PR in skills-dev repo

NikitaCG:
when we install eval in next time, baseline also changed, but we add custom scenario and can improve repo harness on these scenarios as you requested

NikitaCG:
But even now, with such inaccuracies in the current eval, you can see the benefit of Harness.

korvin89 (approved):

PR #2812, opened 2026-09-01
fix(TextArea): skip empty auto-resize measurements

### LINKED ISSUE #2811: TextArea: skip auto-resize measurements for empty values
### Problem

`TextArea` runs auto-resize geometry reads and attaches `ResizeObserver` even when its value is empty. Empty controls do not need height calculation, and repeated empty instances add avoidable layout work.

Scrollbar geometry is also read when the clear control is hidden, although the resulting modifier only affects clear-button spacing.

### Expected behavior

- Empty auto-growing textareas reset inline height and skip geometry reads and resize observation.
- Non-empty values keep current auto-resize behavior.
- Explicit `rows` keep fixed-height behavior.
- Scrollbar detection runs only while the clear control is visible.

daniil-filippov-dev:
Closes #2811.

## Changes

- Skip auto-resize geometry reads and ResizeObserver for empty values.
- Reset stale inline height when an auto-growing textarea becomes empty.
- Limit scrollbar measurement and modifier state to visible clear controls.
- Preserve non-empty auto-resize, default values, and explicit rows behavior.

## Tests

- Focused TextArea Jest suite: 29 passed.
- Scoped ESLint: passed.
- Scoped Prettier check: passed.
- Full TypeScript check: passed.


src/components/controls/TextArea/TextAreaControl.tsx
```diff
@@ -56,13 +56,19 @@ export function TextAreaControl(props: Props) {
     const innerControlRef = React.useRef<HTMLTextAreaElement>(null);
     const handleRef = useForkRef(controlRef, innerControlRef);
     const textareaRows = Math.max(rows || minRows, 1);
-    const innerValue = value || innerControlRef?.current?.value;
+    const innerValue = value ?? innerControlRef.current?.value ?? defaultValue;
+    const shouldAutoResize = !rows && Boolean(innerValue);
```
korvin89: 🤖 **AI generated**

🔴 **blocker**

I think this one needs a look before merge: skipping the measurement for empty values also drops the placeholder from the height calculation, and a wrapping placeholder now gets clipped.

The placeholder does contribute to `textarea.scrollHeight`, so on `main` an empty auto-growing textarea grew to fit it. With `shouldAutoResize === false` the height falls back to `rows` (1 by default), so only the first line stays visible and the rest is scrolled out of view.

I verified this in Chromium via a component test — empty value, no `rows`, 140px-wide container:

| | `main` | this branch |
|---|---|---|
| textarea height | 106px | 26px |
| `scrollHeight` / `clientHeight` | 106 / 106 | 106 / **26** |

This is the default configuration (no `rows`, no `minRows`), so it would affect every consumer whose placeholder is long enough to wrap. The visual snapshots don't catch it because every case in `cases.tsx` uses the single-line `'Placeholder'`.

Would it be possible to keep the measurement when the control is empty but has a placeholder? Something like `!rows && Boolean(innerValue || placeholder)` would cover it, although it does give up part of the optimization for placeholder-ed textareas.

src/components/controls/TextArea/TextAreaControl.tsx
```diff
@@ -113,7 +119,7 @@ export function TextAreaControl(props: Props) {
             ref={handleRef}
             style={{
                 ...controlProps.style,
-                height: rows ? 'auto' : undefined,
+                height: shouldAutoResize ? undefined : 'auto',
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

Rendering `height: 'auto'` for empty values also changes how the control behaves as a flex item, and I suspect that's an unintended side effect.

`.g-text-area__content` is `display: flex` with `height: inherit`, and `.g-text-area__control` gets a definite height from the `input-control` mixin. While that height was definite, `align-items: stretch` had nothing to do; with `height: auto` the control stretches to the wrapper.

Measured in Chromium, `<TextArea style={{height: 200}} />` without `rows`:

| | `main` | this branch |
|---|---|---|
| empty | 26px | **198px** |
| with a value | 26px | 26px |

So the textarea fills the wrapper while empty and collapses to 26px on the first character. The documented pattern from the README (`rows={1}` plus `style`) is unaffected — that path rendered `height: 'auto'` before this PR too, and measures 198px on both sides.

Worth noting this one is invisible in a screenshot, since the visible border belongs to `__content`, which keeps the wrapper height either way — it would need an explicit assertion rather than a snapshot if you want it covered.

src/components/controls/TextArea/TextAreaControl.tsx
```diff
@@ -56,13 +56,19 @@ export function TextAreaControl(props: Props) {
     const innerControlRef = React.useRef<HTMLTextAreaElement>(null);
     const handleRef = useForkRef(controlRef, innerControlRef);
     const textareaRows = Math.max(rows || minRows, 1);
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

For empty values the height now comes from the native `rows` attribute, which doesn't know about `maxRows` — while the measuring branch below still clamps with `Math.min(minRows, maxRows || Infinity)` (line 97). The two paths end up disagreeing.

Verified in Chromium with `<TextArea minRows={5} maxRows={3} />`:

| | `main` | this branch |
|---|---|---|
| empty | 58px (3 rows) | **90px (5 rows)** |
| with a value | 58px (3 rows) | 58px (3 rows) |

So the control renders 5 rows tall while empty and snaps back to 3 on the first character. It only shows up when `maxRows < minRows`, which is arguably a misconfiguration, but the component used to handle it consistently. Could the `rows` attribute be clamped the same way the measurement is?

korvin89 (commented):
🤖 **AI generated**

**Severity legend**

src/components/controls/TextArea/TextAreaControl.tsx
```diff
@@ -56,13 +56,19 @@ export function TextAreaControl(props: Props) {
...
-    const innerValue = value || innerControlRef?.current?.value;
+    const innerValue = value ?? innerControlRef.current?.value ?? defaultValue;
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

`innerValue` reads `innerControlRef.current?.value` during render, and after this change that read decides the rendered output — the inline `style.height` on line 122 and whether `useResizeObserver` subscribes on line 108. Before, it only fed the effect's closure.

React explicitly discourages reading `ref.current` during render, and nothing schedules a re-render when the DOM value changes outside of React state (a consumer writing to `controlRef.current.value`, for example). The blast radius grew from "the next measurement uses a stale string" to "the committed height and the observer subscription are wrong".

`TextArea` already owns the authoritative value in `inputValue` (`TextArea.tsx:75`) but only spreads `{...props}` down (line 174). Would it be worth passing it explicitly? It would need to be a separate prop rather than `value` — passing it as `value` would make the textarea controlled and break the `defaultValue` path.

src/components/controls/TextArea/__tests__/TextArea.test.tsx
```diff
@@ -259,4 +259,140 @@ describe('TextArea', () => {
             expect(clearButton).toBeInTheDocument();
         });
     });
+
+    describe('auto resize', () => {
```
korvin89, line 263: 🤖 **AI generated**

🟠 **should-fix**

Thanks for the thorough suite here. One path isn't covered though: the empty → non-empty transition, i.e. typing the first character into an initially empty auto-growing textarea. That's exactly what this PR introduces — `shouldAutoResize` flips `false → true`, React drops the inline height, the effect measures, the observer attaches.

The suite covers static renders and the non-empty → empty direction, but not this one. The uncontrolled case would be especially valuable, since it relies entirely on the render-time ref read.

I checked that it does work today (`height` goes `"auto"` → `"64px"` and `observe` is called once, in both controlled and uncontrolled mode), so this is purely about locking it in.

src/components/controls/TextArea/__tests__/TextArea.test.tsx
```diff
@@ -259,4 +259,140 @@ describe('TextArea', () => {
...
+            clientHeightSpy.mockClear();
+            observe.mockClear();
+            disconnect.mockClear();
+
+            fireEvent.click(clearButton);
+
+            expect(input).toHaveAttribute('rows', '3');
+            expect(input.style.height).toBe('auto');
+            // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
+            expect(container.querySelector('.g-text-area__clear')).not.toBeInTheDocument();
```
korvin89, line 391: 🤖 **AI generated**

🟠 **should-fix**

This test checks that the clear button disappears and that no new measurements happen, but not the thing the `TextArea.tsx` change actually guarantees: that the root loses `g-text-area_has-scrollbar` when the clear control is hidden, regardless of geometry.

Could you add `expect(container.firstElementChild).not.toHaveClass('g-text-area_has-scrollbar')`? Right now nothing pins the new behaviour on lines 136-142 and 167.

src/components/controls/TextArea/TextAreaControl.tsx
```diff
@@ -56,13 +56,19 @@ export function TextAreaControl(props: Props) {
...
+    const shouldAutoResize = !rows && Boolean(innerValue);
 
     const resizeHeight = React.useCallback(() => {
         const control = innerControlRef?.current;
         const parent = control?.parentElement;
 
         if (control && parent && !rows) {
+            if (!innerValue) {
+                control.style.height = 'auto';
```
korvin89: 🤖 **AI generated**

🔵 **nit**

This line looks redundant: the branch is only reachable when `!rows && !innerValue`, which is exactly when the render already applied `height: 'auto'` on line 122 from the same `innerValue` snapshot. I removed it locally and the whole suite stayed green.

It might be load-bearing in one narrow race — a callback from the not-yet-disconnected observer firing between commit and the passive effect with a stale non-empty closure. If that's the intent, a one-line comment would be great; otherwise someone will delete it during a future refactor.

src/components/controls/TextArea/__tests__/TextArea.test.tsx
```diff
@@ -259,4 +259,140 @@ describe('TextArea', () => {
...
+            expect(scrollHeightSpy).not.toHaveBeenCalled();
+            expect(clientHeightSpy).not.toHaveBeenCalled();
+            expect(observe).not.toHaveBeenCalled();
+        });
+
+        test('auto resizes a non-empty default value', () => {
+            render(<TextArea defaultValue="value" />);
+
+            expect(getComputedStyleSpy).toHaveBeenCalled();
+            expect(observe).toHaveBeenCalledWith(screen.getByRole('textbox'), {box: undefined});
```
korvin89: 🤖 **AI generated**

🔵 **nit**

`{box: undefined}` pins an implementation detail of `useResizeObserver` (it calls `observer.observe(element, {box})`). If that hook ever stops forwarding the options object, these two assertions break for an unrelated reason. `toHaveBeenCalledWith(input, expect.anything())` or `toHaveBeenCalledTimes(1)` would convey the same intent.

src/components/controls/TextArea/__tests__/TextArea.test.tsx
```diff
@@ -259,4 +259,140 @@ describe('TextArea', () => {
...
+        const disconnect = jest.fn();
+
+        beforeEach(() => {
+            originalResizeObserver = global.ResizeObserver;
+            global.ResizeObserver = class implements ResizeObserver {
+                disconnect = disconnect;
+                observe = observe;
+                unobserve() {}
+            };
+            getComputedStyleSpy = jest.spyOn(window, 'getComputedStyle').mockReturnValue({
```
korvin89: 🤖 **AI generated**

🔵 **nit**

Two small things about this mock, if you feel like tightening it up.

The spy replaces `getComputedStyle` for the whole document, and the stub returns `undefined` for any property other than the three listed — so anything else in the tree (RTL matchers, `ClearButton`) silently gets a broken `CSSStyleDeclaration`.

And `expect(getComputedStyleSpy).not.toHaveBeenCalled()` asserts that *nothing in the rendered tree* touched `getComputedStyle`, which is broader than "TextArea skipped its measurement". It's also coupled to the query choice — `screen.getByRole('textbox')` itself calls `window.getComputedStyle` five times through dom-testing-library's visibility check, which is presumably why these tests have to go through `container.querySelector` with the extra eslint-disable lines.

Scoping the spy to the textarea element, plus asserting the observable outcome (`input.style.height`), would be less brittle.

korvin89:
🤖 **AI generated**

🟠 **should-fix** — visual coverage for the empty state

Following up on the sizing comments in the review: `src/components/controls/TextArea/__tests__/TextArea.visual.test.tsx` currently can't catch any of them. `createSmokeScenarios` varies one prop at a time from `defaultProps`, so `minRows` and `maxRows` never appear together; `placeholder` lives in `defaultProps` and is always the single-line `'Placeholder'`; and `style` isn't in the matrix at all.

Rather than extending `commonPropCases` (which would regenerate the three existing snapshots), a dedicated test would do it:

```tsx
const longPlaceholder =
    'A placeholder long enough to wrap onto several lines inside a narrow text area';

test('empty sizing', async ({mount, expectScreenshot}) => {
    const sizingScenarios: Array<[string, TextAreaProps]> = [
        ['long placeholder, empty', {placeholder: longPlaceholder}],
        ['long placeholder, with value', {placeholder: longPlaceholder, value: 'Value'}],
        ['minRows > maxRows, empty', {...defaultProps, minRows: 5, maxRows: 3}],
        ['minRows > maxRows, with value', {...defaultProps, minRows: 5, maxRows: 3, value: 'Value'}],
        ['outer height, empty', {...defaultProps, style: {height: 120}}],
        ['outer height, with value', {...defaultProps, style: {height: 120}, value: 'Value'}],
        ['outer height with rows, empty', {...defaultProps, style: {height: 120}, rows: 1}],
    ];

    await mount(
        <div>
            {sizingScenarios.map(([title, props]) => (
                <div key={title}>
                    <h4>{title}</h4>
                    <div style={{width: 140}}>
                        <TextArea {...props} />
                    </div>
                </div>
            ))}
        </div>,
    );

    await expectScreenshot({themes: ['light']});
});
```

The empty/with-value pairs matter — the height jump on the first character is only visible when both states sit in the same snapshot.

I ran it: with the snapshot generated on this branch it passes, and restoring only the two source files from `main` makes it fail with `488×1952` vs `488×1984` and 23132 differing pixels — it does capture the clipped placeholder and the `minRows`/`maxRows` jump.

One caveat: if you add this test, please generate the snapshot **after** fixing the placeholder issue, otherwise it locks in the clipped rendering as the expected result. It also does not cover the flex-stretch case, for the reason given in that comment.

src/components/controls/TextArea/TextAreaControl.tsx
```diff
@@ -52,22 +53,30 @@ export function TextAreaControl(props: Props) {
...
+        inputValue,
     } = props;
     const innerControlRef = React.useRef<HTMLTextAreaElement>(null);
     const handleRef = useForkRef(controlRef, innerControlRef);
-    const textareaRows = Math.max(rows || minRows, 1);
-    const innerValue = value || innerControlRef?.current?.value;
+    const textareaRows = rows
+        ? Math.max(rows, 1)
+        : Math.max(Math.min(minRows, maxRows || Infinity), 1);
+    const shouldAutoResize = !rows && Boolean(inputValue || placeholder);
```
korvin89: 🤖 **AI generated**

🔴 **blocker**

Thanks for the quick turnaround — the four source-level points are genuinely fixed. This one looks new, introduced by this commit.

Changing the reset on line 71 from `control.style.height = 'auto'` to `= ''`, together with the render going back to `height: rows ? 'auto' : undefined`, leaves the empty / no-placeholder / no-`rows` path with no inline height at all. `.g-text-area__control` has a **definite** height from the `input-control` mixin (`height: calc(#{$m-height} - border * 2)`), and a definite `height` overrides the `rows` attribute — `rows` only drives intrinsic sizing.

Measured on `042bb25a` against `main`:

| | `main` | `042bb25a` |
|---|---|---|
| `<TextArea minRows={3} />` (empty, no placeholder) | 58px | **26px** |
| `<TextArea minRows={5} maxRows={3} />` (empty, no placeholder) | 58px | **26px** |
| `<TextArea minRows={3} placeholder="P" />` (empty) | 58px | 58px |

So an empty textarea with `minRows > 1` and no placeholder collapses from `minRows` rows down to one. To confirm the mechanism I set `style.height = 'auto'` on that same element at runtime and it measured 58px, while as rendered `getComputedStyle(...).height` reports `26px`.

This also masks the `maxRows` fix: in the second row above the `rows` attribute is correctly `3`, but the height stays 26px either way — which is why I left that thread open.

Adding `minRows` to the condition would cover it, something like `!rows && Boolean(inputValue || placeholder || minRows > 1)`.

While you're here, it might be worth re-weighing the original goal from #2811: now that `placeholder` (and probably `minRows`) has to stay in the condition, the measurement is only skipped for empty single-row textareas with no placeholder, which is a fairly small slice of real usage.

src/components/controls/TextArea/__tests__/TextArea.test.tsx
```diff
@@ -259,4 +259,196 @@ describe('TextArea', () => {
...
+            expect(input).toHaveAttribute('rows', '3');
+            expect(input?.style.height).toBe('');
+            expect(isCalledFor(getComputedStyleSpy, input)).toBe(false);
+            expect(offsetHeightSpy).not.toHaveBeenCalled();
+            expect(scrollHeightSpy).not.toHaveBeenCalled();
+            expect(clientHeightSpy).not.toHaveBeenCalled();
+            expect(observe.mock.calls.map(([element]) => element)).not.toContain(input);
+        });
+
+        test('keeps native minRows height for an uncontrolled empty value', () => {
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

This test renders exactly the configuration that regressed (`minRows={3}`, empty, no placeholder) and passes, because jsdom has no layout — it can only observe `rows` and `style.height`, not the height the browser computes. In Chromium this control renders 26px, not the three rows the test name promises.

Could this assertion move to `TextArea.visual.test.tsx`, next to the two Playwright checks you added there? Comparing `textarea.evaluate((el) => el.getBoundingClientRect().height)` against the same control with a placeholder would catch it. The snapshots don't cover it either, since `defaultProps` there always sets `placeholder: 'Placeholder'`.

korvin89 (commented):
🤖 **AI generated**

I re-checked every comment from the previous review against `042bb25a` in Chromium.

Fixed and verified, so I'm resolving those threads: the wrapping placeholder, the flex stretch, the render-time ref read, and the four test points. Thanks for the thorough follow-up — the `HTMLTextAreaElement`-scoped `getComputedStyle` spy and the `isCalledFor` helper are a nice improvement.

I've left the `maxRows` thread open: the `rows` attribute is clamped correctly now, but that configuration still changes height on the first character for a different reason. Details in the comment below.

korvin89 (approved):
🤖 **AI generated**

Re-verified `dcd03b79` in Chromium against `main`. All 13 sizing configurations I probed now render identically: plain empty, `minRows` with and without a placeholder, the `minRows`/`maxRows` inversion, `maxRows` below `minRows`, an outer height on the wrapper, a wrapping placeholder, fixed `rows`, `view="clear"` and `size="l"`. The only thing that differs is the `rows` attribute itself, which now reflects the effective row count — the rendered height is unchanged.

Using `textareaRows > 1` rather than `minRows > 1` is the better call: `<TextArea minRows={3} maxRows={1} />` keeps the single-row height and still skips the measurement.

I also ran the new `keeps empty minRows height without a placeholder` check against the previous commit — it fails there with `Expected: > 26, Received: 26`, so it is a real regression guard rather than a passing formality.

One thing worth keeping in mind rather than acting on now: with `placeholder` and `textareaRows > 1` both in the condition, the measurement is only skipped for an empty single-row textarea with no placeholder. The correctness is right, but the performance win is much smaller than the original patch implied — might be worth a note on #2811.

Thanks for the quick iterations on this.

PR #2823, opened 2026-09-09
feat!: add virtualizer and hello-pangea-dnd entry points

korvin89:
🤖 AI generated

## What's added

Two entry points, alongside the existing `/unstable`, `/legacy`, `/server`, `/i18n`, `/toaster-singleton` and `/styles/*`:

| Entry point | Exports | Source |
| --- | --- | --- |
| `@gravity-ui/uikit/virtualizer` | `Virtualizer`, `ListVirtualizer`, `VirtualizerProps`, `VirtualizerItem`, `VirtualizerApi`, `ScrollAlignment`, `ListVirtualizerProps` | `src/virtualizer.ts` ← `src/components/Virtualizer/` |
| `@gravity-ui/uikit/hello-pangea-dnd` | `useListHelloPangeaDnd`, `UseListHelloPangeaDndOptions`, `UseListHelloPangeaDndResult` | `src/hello-pangea-dnd.ts` ← `src/components/HelloPangeaDnd/` |

## Moved

- `src/components/lab/Virtualizer/` → `src/components/Virtualizer/` (`Virtualizer`, `ListVirtualizer`, the private `useLoadMore`)
- `src/components/lab/List/useListHelloPangeaDnd.ts` and its test → `src/components/HelloPangeaDnd/`

`VirtualizerProps` and the row type (`VirtualizerItem`, previously the private `Item`) are exported now — a public entry point without the types of its props is incomplete. Neither directory is re-exported from `src/components/index.ts`.

## Dependencies

`@tanstack/react-virtual` moves from `dependencies` to `peerDependencies` + `peerDependenciesMeta.optional`, and stays in `devDependencies` for the build, the tests and Storybook. It is imported by `src/components/Virtualizer/Virtualizer.tsx` only, so the main entry point does not need it.

`@hello-pangea/dnd`, `react-window` and `react-virtualized-auto-sizer` stay in `dependencies`: the `List` and the `TableColumnSetup` of the main entry point import them at runtime, and an optional peer would make `npm i @gravity-ui/uikit` a silently broken install. Each of them becomes an optional peer in the release that moves its last consumer out of the main entry.

## Guards

`src/__tests__/entry-isolation.test.ts` builds two import graphs of the sources — runtime (`import type` skipped, those lines are erased from the emit) and types (kept: a consumer with `skipLibCheck: false` must not reference the types of a package that is not installed). For each of the six shared entry points it asserts that neither graph reaches `@tanstack/react-virtual` or a file under `src/components/Virtualizer/` and `src/components/HelloPangeaDnd/`. Verified by hand: an `export * from './Virtualizer'` in `src/components/index.ts` fails four of the tests.

`no-restricted-imports` in `eslint.config.mjs` is the local guard for the same rule, deep imports of both packages included.

## Docs

- `README.md` and `README-ru.md`: an "Entry points" section listing all nine subpaths and what has to be installed
- `contribute/architecture.md` and `contribute/architecture-ru.md`: two new rows in the public surface table
- `src/components/Virtualizer/README.md` and `src/components/HelloPangeaDnd/README.md` — both pass `gravity-readme --component` and ship as AI docs
- `src/components/lab/List/README.md`: the imports in the virtualization and drag-and-drop sections

## Breaking changes

- removed from `@gravity-ui/uikit/unstable`: `unstable_ListVirtualizer`, `unstable_ListVirtualizerProps`, `unstable_useListHelloPangeaDnd`, `unstable_UseListHelloPangeaDndOptions`, `unstable_UseListHelloPangeaDndResult` — import them from the new entry points instead
- `@tanstack/react-virtual` has to be installed alongside the package when importing from `@gravity-ui/uikit/virtualizer`

## Checks

`typecheck`, `lint`, `jest` (1478 tests), `build`, `build-storybook`, `npm pack --dry-run`, `gravity-readme` for both new READMEs.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -8,6 +8,12 @@ import storybookPlugin from 'eslint-plugin-storybook';
 import testingLibraryPlugin from 'eslint-plugin-testing-library';
 import globals from 'globals';
 
+const TANSTACK_MESSAGE =
+    'Belongs to the @gravity-ui/uikit/virtualizer entry point. Every other import needs an eslint-disable with a reason.';
+
+const DND_MESSAGE =
+    'Belongs to the @gravity-ui/uikit/hello-pangea-dnd entry point. Every other import needs an eslint-disable with a reason.';
+
 export default defineConfig([
     ...baseConfig,
     ...clientConfig,
@@ -96,6 +102,27 @@ export default defineConfig([
                 ...globals.node,
                 ...globals.jest,
             },
+        },
+    },
+    // The packages of the dedicated entry points are restricted everywhere: an import outside of
+    // them is an explicit eslint-disable, so the places that carry one are easy to list.
+    // `paths` matches the exact specifier, `patterns` the deep imports of the same package.
+    {
+        files: ['src/**/*.{ts,tsx}'],
+        rules: {
+            'no-restricted-imports': [
+                'error',
+                {
+                    paths: [
+                        {name: '@tanstack/react-virtual', message: TANSTACK_MESSAGE},
+                        {name: '@hello-pangea/dnd', message: DND_MESSAGE},
+                    ],
+                    patterns: [
+                        {group: ['@tanstack/react-virtual/*'], message: TANSTACK_MESSAGE},
+                        {group: ['@hello-pangea/dnd/*'], message: DND_MESSAGE},
+                    ],
+                },
+            ],
         },
     },
     {
--- a/src/components/HelloPangeaDnd/index.ts
+++ b/src/components/HelloPangeaDnd/index.ts
@@ -1,0 +1,5 @@
+export {useListHelloPangeaDnd} from './useListHelloPangeaDnd';
+export type {
+    UseListHelloPangeaDndProps,
+    UseListHelloPangeaDndResult,
+} from './useListHelloPangeaDnd';
--- a/README-ru.md
+++ b/README-ru.md
@@ -73,6 +73,25 @@ import '@gravity-ui/uikit/styles/styles.css';
 - [Серверный рендеринг (SSR)](docs/server-side-rendering-ru.md) — генерация корневого CSS-класса на сервере
 - [Интернационализация (I18N)](docs/i18n-ru.md) — язык встроенных текстов компонентов
 
+### Точки входа
+
+| Точка входа                           | Что там                                     | Когда брать                                                                          |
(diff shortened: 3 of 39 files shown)


src/components/HelloPangeaDnd/index.ts
```diff
@@ -0,0 +1,5 @@
+export {useListHelloPangeaDnd} from './useListHelloPangeaDnd';
+export type {
+    UseListHelloPangeaDndOptions,
```
amje: Let's rename it to `UseListHelloPangeaDndProps` as in all other our hooks

eslint.config.mjs
```diff
@@ -98,6 +111,48 @@ export default defineConfig([
...
+            'src/**/__tests__/**/*.{ts,tsx}',
+        ],
+        ignores: ['src/components/Virtualizer/**/*.{ts,tsx}'],
+        rules: {
+            'no-restricted-imports': [
+                'error',
+                {paths: [TANSTACK_RESTRICTION], patterns: TANSTACK_PATTERNS},
+            ],
+        },
+    },
```
amje: Too much rules here. Maybe just restrict those packages everywhere and ingore comment it in places where needed?

amje (changes requested):

src/components/HelloPangeaDnd/index.ts
korvin89: 🤖 AI generated

Renamed to `UseListHelloPangeaDndProps` — checked the other hooks and `*Props` is indeed the convention (`UseAsyncActionHandlerProps`, `UseCollapseChildrenProps`, `UseFocusWithinProps`). Also renamed the prose in both READMEs from "The options:" to "The props:" so the docs match.

eslint.config.mjs
korvin89: 🤖 AI generated

Agreed, done — both packages are now restricted across `src/**` by a single block, and the 20 legitimate imports carry an `// eslint-disable-next-line no-restricted-imports` each. The config lost 28 lines and, more importantly, the three-block layering: it used to rely on flat config replacing (not merging) the rule options, so every block had to repeat the restrictions that still applied to it — easy to get wrong on the next change.

The list of places that legitimately import them is now greppable by the directive instead of being encoded in glob patterns. Verified: no unused directives, and removing one brings the error back.

amje (approved):

Issue #10, opened 2021-12-17
Demos and examples 


ierehon1905:
Hi, thank you for your awesome library.

I wondered if you could make more examples and a demo, maybe in form of GitHub pages. I think it s pretty important for a UI library to have a showcase.

amje:
Hi, there. We are working on that. Stay tuned!)

amje:
We've recently released docs section in our site: https://gravity-ui.com/components/uikit

PR #2826, opened 2026-09-11
feat(Select)!: migrate to the new List

korvin89:
🤖 AI generated

The options of `Select` are drawn by the new list instead of the old `List`. Second PR of three: the core this needs is #2825, the documentation follows in the next one.

## Why

The old `List` is the last thing holding `react-window` in the bundle of the main entry point, and its keyboard and ARIA live apart from the ones the new core implements. The Select is the first big consumer to move: it gets the keyboard, the ARIA and the row view of the core, and everything it used to carry itself — the quick search, the tick icon, the scroll to the selected option, the virtualization by a threshold — is either the core's job now or gone.

## What the popup looks like from the outside

- the trigger (or the filter input) keeps the focus, and the active option is pointed at with `aria-activedescendant`. A click on an option no longer takes the focus away, so the filter keeps working after it;
- a group is a section: its header is a caption rather than an option, it names the options under it with `aria-describedby`, and a group with an empty label draws a separating line;
- the rows are the rows of the new design: inset from the edges of the popup with a rounded highlight, and in `multiple` the check mark stands in a column of its own.

## What changes for a consumer

| Was | Now | Why |
| --- | --- | --- |
| `virtualizationThreshold` | wrap the `Select` in `ListVirtualizer` from `@gravity-ui/uikit/virtualizer` | virtualization is a decision of the consumer, not a number of options; the entry point keeps `@tanstack/react-virtual` out of an install that does not need it |
| `SelectOption.text` | `getOptionText` on the `Select` and on `useSelectOptions` | one getter feeds the trigger, the filter and the search, instead of a field repeated on every option |
| the row id was `${popupId}-item-${index}` | `${popupId}-item-${encodeURIComponent(value)}` | the row is identified by the option, so the id survives filtering and reordering |
| a group header was an option | `role="presentation"` | a caption is not something you can pick; it also drops out of the count of options |
| the class names of the rows (`__group-label`, `__tick-icon`, `__option-default-label_disabled`, `.g-list__item`) | gone | the markup of a row belongs to the list and its view now |
| the search by the first letters matched a substring, kept the query for two seconds and let `Backspace` edit it | matches a prefix, keeps the query for a second, cycles on a repeated letter | it is the search of the core, shared with every list |
| the popup was as wide as the control | it can be wider — the width follows the content | the row is wider than it was: the check mark has its own column and the rows are inset. `width="fit"` pins the popup to the control as before |
| `renderFilter` gave `value`, `onChange`, `onKeyDown` | `inputProps`, which also carries `role`, `aria-controls`, `aria-activedescendant`, `aria-expanded` | the old three are still there and still work, but only `inputProps` makes the filter a combobox for a screen reader |

`SelectQa.ACTIVE_ITEM` is new, and its value is the old `list-active-item` — the e2e tests of consumers that look for the active option keep working. Above 150 options the Select writes a development warning pointing at `ListVirtualizer`.

## Two things worth a look in the diff

**The active option and the scroll to it.** The choice is made anew on every opening — the selected option, otherwise the first one that can be active — and an option the filter takes away falls back to the first navigable one without coming back when the filter is cleared. Bringing it into view is a short watch over the frames: the virtualizer keeps correcting the total height while it measures the rows, the row drifts along with it, and the list re-aligns until the row stands still for three frames in a row.

**The loading row.** It asks for more options through an `IntersectionObserver`, and the element has to reach the observer through state: a row of the new list renders once and is memoized after that, so an element caught by a ref never arrives. Without it `onLoadMore` never fires; there is a test that fails if the element stops being observed.

## Checks

`typecheck`, `eslint`, `stylelint`, `prettier`, `jest` (1386 tests), `playwright` (513 visual tests) — run on this branch alone, in a separate worktree.


src/components/Select/components/SelectList/SelectList.tsx
```diff
@@ -29,135 +43,262 @@ type SelectListProps = {
...
+            onLoadMore,
+            id,
+            labelledBy,
+            focusOwner,
+            activeItemId,
+            onActiveItemUpdate,
+        } = props;
+
+        const getItemTextValue = React.useCallback(
+            (node: SelectListNode) => getSelectListNodeText(node, getOptionText),
```
jhoncool, line 88: **#1 [P2] Do not pass the internal loading row to `getOptionText`**

`getItemTextValue` is called for every List node, including the internal `LOADING_OPTION`. With `loading` and a domain getter such as `option => option.data!.label`, opening the popup throws because the sentinel has no `data`. Please return an internal text value for the loading row without invoking the consumer callback, and add coverage for `loading + getOptionText`.

AI generated

src/components/Select/utils.tsx
```diff
@@ -53,64 +71,129 @@ export const getFlattenOptions = (options: SelectOptions): FlattenOptions => {
...
+export const buildSelectListNodes = (
+    flattenOptions: FlattenOption[],
+    loading?: boolean,
+): SelectListNode[] => {
+    const nodes: SelectListNode[] = [];
+    let currentGroup: SelectGroupNode | undefined;
+
+    flattenOptions.forEach((option, index) => {
+        if (isSelectGroupTitle(option)) {
+            currentGroup = {...option, id: `__group_${index}`, options: []};
```
jhoncool, line 130: **#2 [P2] Keep section IDs disjoint from option values**

A valid option value such as `__group_0` collides with the generated ID of the first section. `List.flattenItems` then overwrites the section in `rowById`, so the header disappears and the option is rendered twice with duplicate React and DOM IDs. Please generate section IDs that are guaranteed not to collide with any `option.value`, or separate internal row identity from the option DOM-ID contract.

AI generated

src/components/Select/utils.tsx
```diff
@@ -53,64 +71,129 @@ export const getFlattenOptions = (options: SelectOptions): FlattenOptions => {
...
+            nodes.push(currentGroup);
+            return;
+        }
+
+        if (currentGroup) {
```
jhoncool, line 135: **#3 [P2] Preserve the actual boundary of each `SelectOptionGroup`**

`currentGroup` is never cleared after the original group's options. For the publicly valid input `[group, standaloneOption]`, the standalone option becomes a child of the preceding section and receives that header's `aria-describedby`. Please preserve group membership while flattening/filtering, or build the List tree from the structured options rather than inferring boundaries from headers.

AI generated

src/components/Select/hooks/useAlignActiveOption.ts
```diff
@@ -0,0 +1,123 @@
...
+                return;
+            }
+
+            request = requestAnimationFrame(tick);
+        };
+
+        tick();
+
+        return () => cancelAnimationFrame(request);
+    }, [activeItemId, alignActiveRow]);
```
jhoncool, line 122: **#4 [P2] Restart alignment when rows or their geometry change**

This effect does not rerun when `rows` or `getItemHeight` change while `activeItemId` stays the same. Inserting rows before the active option can therefore move the keyboard cursor outside the viewport without scheduling another alignment frame. Please include a stable rows/layout version in the effect trigger and restart the frame watch after such changes.

AI generated

src/components/Select/components/SelectList/SelectList.tsx
```diff
@@ -29,135 +43,262 @@ type SelectListProps = {
...
+                {getItemProps, getItemViewProps}: ListItemHelpers,
+            ) => {
+                const isItemActive = ctx.state.active;
+                const itemHeight = getItemHeight(ctx.item, ctx.index);
+                const isSection = ctx.kind === 'section';
+                // The row view sizes itself: an inline height is only what the consumer asked for
+                const hasCustomHeight = Boolean(isSection ? getOptionGroupHeight : getOptionHeight);
+                // minHeight too: the view has a minimum of its own for every size, and the number
+                // the consumer returned has to be the height of the row exactly
+                const style = hasCustomHeight
```
jhoncool, line 161: **#5 [P2] Apply `getOptionGroupHeight` to the outer border box**

For the default renderer this inline height lands on a content-box `List.SectionHeader`; its vertical padding is then added on top. A requested 32px row becomes 40px for the first header and 48px for later headers, while the virtualizer still estimates 32px. With `measure={false}` rows can overlap. Please make the group row border-box or fix the height on an outer wrapper.

AI generated

src/components/Select/Select.tsx
```diff
@@ -111,9 +107,16 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function
...
-    const listRef = React.useRef<List<FlattenOption>>(null);
     const handleControlRef = useForkRef(ref, controlRef);
 
+    // One owner for two elements: the trigger and the filter input. The list is mounted only while
+    // the popup is open, so the owner is connected exactly then
+    const focusOwner = useListFocusOwner();
+    // Virtualization is opt-in from the outside: <ListVirtualizer> around the Select. The context
+    // reaches the list through the portal of the popup; the Select needs to know about it for the
+    // width of the popup and for the modifier of the list
+    const virtualized = React.useContext(ListVirtualizationContext) !== null;
```
jhoncool, line 118: **#6 [P2] Do not inherit virtualization from an unrelated ancestor List**

Any Select rendered inside a row of another `ListVirtualizer` sees this context and silently virtualizes its own popup; React context also crosses the popup portal. It consequently inherits settings such as `measure={false}` and `overscan` from the outer list, with no opt-out, which can break custom Select rows of variable height. Please scope the context to the intended List/Select consumer and add a nested-list regression test.

AI generated

src/components/Select/types.ts
```diff
@@ -106,6 +109,12 @@ export type SelectProps<T = any> = AriaLabelingProps &
         renderEmptyOptions?: ({filter}: {filter: string}) => React.ReactElement;
         renderPopup?: SelectRenderPopup;
         renderCounter?: SelectRenderCounter;
+        /**
+         * The text of an option: the trigger shows it for a selected option, the filter matches it
+         * and the search by the first letters looks it up. Defaults to the content of the option
+         * when that is a string, otherwise to its value — `getSelectOptionText` is that default
+         */
+        getOptionText?: (option: SelectOption<T>) => string;
```
jhoncool, line 117: **#7 [Standards] Update the authoritative Select documentation before integration**

This public API addition accompanies removal of `option.text` and `virtualizationThreshold`, but the component README still documents built-in threshold virtualization, lists the removed prop, and does not explain `getOptionText` or the migration path. The PR description explicitly defers docs to the next stacked PR, so this need not block this layer, but the documentation update should be required before the stack reaches `main`.

AI generated

src/components/Select/constants.ts
```diff
@@ -12,27 +12,55 @@ export const selectListBlock = block('select-list');
...
     xl: 36,
 };
 
-export const MOBILE_ITEM_HEIGHT = 32;
+/** A row on mobile is a row of size `l` */
+export const MOBILE_SIZE: SelectSize = 'l';
 
-export const GROUP_ITEM_MARGIN_TOP = 5;
+/**
+ * The heights of a section header (ListSectionHeader): the vertical padding of `--g-spacing-1`
```
jhoncool, line 31: **#8 [Standards heuristic] Avoid duplicating token-derived CSS geometry in TypeScript**

These pixel totals mirror spacing and typography values that are independently defined in CSS. A token or line-height change can therefore leave the virtualizer estimates stale unless both places are updated together. Consider exposing authoritative row geometry from List/Virtualizer or measuring it instead of maintaining a second set of derived constants.

AI generated

src/components/Select/hooks/useAlignActiveOption.ts
```diff
@@ -0,0 +1,123 @@
...
+    /** The height a row comes out as — the very estimate the virtualizer is given */
+    getItemHeight: (row: FlattenOption, index: number) => number;
+}
+
+/**
+ * Brings the active option into view when the list opens — the core scrolls on a keyboard gesture of
+ * its own only. It scrolls the root of the list rather than calling `scrollIntoView`, so that a popup
+ * hanging off the viewport edge never drags the page along.
+ *
+ * TODO: this belongs to the list itself — once `ListVirtualizer` hands out its virtualizer
```
jhoncool, line 30: **#9 [Standards heuristic] Move item alignment behind the List abstraction**

This hook reconstructs List DOM IDs, walks its offset hierarchy, and implements virtualizer settling behavior; the TODO already notes that this responsibility belongs to List. Exposing an item-ID-based alignment/scroll API from List or ListVirtualizer would keep Select from depending on these implementation details.

AI generated

src/components/Select/Select.tsx
```diff
@@ -141,12 +144,18 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function
...
     });
     const filteredOptions = getSelectFilteredOptions(options) as FlattenOption[];
     const selectedOptionsContent = React.useMemo(() => {
-        return getSelectedOptionsContent(options, value, renderSelectedOption);
-    }, [options, value, renderSelectedOption]);
-    const virtualized = filteredOptions.length >= virtualizationThreshold;
+        return getSelectedOptionsContent(options, value, renderSelectedOption, getOptionText);
+    }, [options, value, renderSelectedOption, getOptionText]);
+
+    if (!virtualized && filteredOptions.length > VIRTUALIZATION_HINT_OPTIONS_COUNT) {
```
jhoncool, line 154: **#10 [Spec] Count options rather than section headers for the virtualization warning**

`filteredOptions.length` includes presentational group headers. For example, 149 options split across two groups have a flattened length of 151 and incorrectly trigger the warning, although the requirement says it applies above 150 options. Please count only entries for which `!isSelectGroupTitle(option)` and add a grouped boundary test.

AI generated

src/components/Select/components/SelectList/SelectList.tsx
korvin89: 🤖 AI generated

Fixed in the base PR of this stack — https://github.com/gravity-ui/uikit/pull/2825 (commit 4e8a341f).

`List.SectionHeader` is a border box now, the way the row view already was, so a height fixed on the header is the height of the row with its padding counted in — the same number the virtualizer is given. The header that follows other rows no longer grows by the spacing above its line either.

src/components/Select/Select.tsx
korvin89: 🤖 AI generated

Fixed in the base PR of this stack — https://github.com/gravity-ui/uikit/pull/2825 (commit 4e8a341f).

The list hands its children an empty virtualization context, so a list — or a component that holds one, such as a `Select` — rendered inside a row no longer sees the wrapper of the outer list and no longer inherits its `Root`, `measure` and `overscan`. The regression test is there: "a list inside a row does not inherit the virtualization of the outer one" (it fails without the fix, the inner rows come back virtualized).

src/components/Select/components/SelectList/SelectList.tsx
korvin89: 🤖 AI generated

Fixed in fb90019a.

`getSelectListNodeText` answers for the loading row itself now — it is our own row, it has no text of its own, and a getter written for the options of the consumer has no reason to be called with it. Test: "the loading row does not reach getOptionText" — the getter throws on an option without `data`, and opening a `loading` popup used to bring the whole render down with it.

src/components/Select/utils.tsx
korvin89: 🤖 AI generated

Fixed in fb90019a.

The prefix of a section id is chosen so that no value of an option starts with it: `__group_` is extended with underscores until it is free. Test: "an option whose value looks like the id of a section keeps its own row" — with a value of `__group_0` beside a group, the header used to disappear and two rows shared an id.

src/components/Select/utils.tsx
korvin89: 🤖 AI generated

Fixed in fb90019a.

Group membership is written down while flattening still knows it — a `Map` from an option to its group title, kept in the same non-enumerable bag as `filteredOptions` — and the tree of the list is built from that rather than from position. Filtering keeps it working: the options in the filtered array are the same objects.

Test: "an option that merely follows a group stays outside of it" — the standalone option is outside the section and carries no `aria-describedby`, while the option of the group still points at its header.

src/components/Select/hooks/useAlignActiveOption.ts
korvin89: 🤖 AI generated

Fixed in fb90019a.

The rows and the height getter are dependencies of the watch now. Test: "rows that moved start the watch again" — rows that appear above the active option move it out of the viewport without changing its id, and nothing used to schedule another frame. A restart costs nothing when the row is already in view: the alignment is a no-op and the watch ends after three quiet frames.

src/components/Select/Select.tsx
korvin89: 🤖 AI generated

Fixed in fb90019a.

The hint counts options, not rows — group headers no longer add to the number. Test: "the group headers do not count towards the hint" — exactly 150 options split across two groups, where the flat list is 152 rows long, no longer warn.

src/components/Select/types.ts
korvin89: 🤖 AI generated

Agreed, and that is the plan: the documentation is the third PR of this stack — the README of both languages, a `migration-guide.md` for the component, and the README of `useSelectOptions`. The stack does not reach `main` without it.

src/components/Select/constants.ts
korvin89: 🤖 AI generated

Agreed — the constants mirror geometry that CSS owns, and that is a duplication with no guard. The reason it is here is that the estimate has to exist before a row is rendered, and today the core gives the consumer no way to ask a list for the height of a row.

Keeping it as a follow-up for the list itself: row geometry should come from the core (or be measured), and these constants should disappear along with it. For now the arithmetic is spelled out next to each constant, so a token change at least has something to be checked against.

src/components/Select/hooks/useAlignActiveOption.ts
korvin89: 🤖 AI generated

Agreed — that is what the `TODO` in the hook says, and it is a planned follow-up rather than an oversight.

What blocks it today: the core deliberately does not import the virtualizer, and the bridge between them (`ListVirtualizationContext`) is render-only — there is no imperative handle, and `ListVirtualizer` does not expose the `scrollToIndex` of its virtualizer, which already implements the settling this hook reimplements by hand. Once that handle exists, the alignment becomes an item-id API on the list (`scrollToItem(id)` plus an opt-in for "align on mount"), and both this hook and the DOM-id arithmetic leave the Select.

src/components/Select/types.ts
jhoncool, line 118: The new public `getOptionText` API is not documented, while the component README still advertises the removed `virtualizationThreshold` and built-in virtualization. This README is used for package-facing component documentation, so the merged package would direct consumers to an API that no longer compiles. Please update the English and Russian virtualization sections, the properties table, and the custom-filter example to cover `getOptionText`, external `ListVirtualizer`, and the new `inputProps` ARIA contract.

AI generated

src/components/Select/hooks/useAlignActiveOption.ts
```diff
@@ -0,0 +1,125 @@
...
```
jhoncool, line 30: This hook reaches into List internals (`getItemDomId`, virtualizer estimates, DOM offsets, and frame-settling behavior) to implement a List scrolling concern. That couples Select to the List DOM and virtualization implementation, so changes on either side require coordinated updates. Could List/ListVirtualizer expose an item-id alignment API and let Select delegate to it?

AI generated

src/components/Select/utils.tsx
```diff
@@ -139,7 +283,7 @@ export const getSelectedOptionsContent = (
     } else {
         return selectedOptions
             .map((option) => {
-                return getOptionText(option);
+                return resolveOptionText(option, getOptionText);
```
jhoncool: When a controlled value is temporarily absent from `options` (for example while async options load), line 272 fabricates `{value}` and this call passes that synthetic object to the consumer getter. A valid getter such as `option => option.data!.label` then throws before the options arrive. Please invoke `getOptionText` only for options found in `optionsMap` and keep the raw value fallback for missing entries.

AI generated

src/components/Select/utils.tsx
```diff
@@ -48,69 +84,177 @@ export const getFlattenOptions = (options: SelectOptions): FlattenOptions => {
...
+    groupOfOption?: Map<SelectOption, GroupTitleItem>,
+): SelectListNode[] => {
+    const nodes: SelectListNode[] = [];
+    const sectionOfTitle = new Map<GroupTitleItem, SelectGroupNode>();
+    const prefix = getSectionIdPrefix(flattenOptions);
+    let previousTitle: GroupTitleItem | undefined;
+
+    flattenOptions.forEach((option, index) => {
+        if (isSelectGroupTitle(option)) {
+            const section: SelectGroupNode = {...option, id: `${prefix}${index}`, options: []};
```
jhoncool, line 195: Filtering can leave this section empty when all of its own options are removed but a following standalone option matches. For example, a group containing Alpha followed by standalone Beta and filter Beta produces an empty group header plus Beta. Please retain a title only when a filtered option maps to it, or prune sections with no children after building the nodes.

AI generated

src/components/Select/components/SelectList/SelectList.tsx
```diff
@@ -2,162 +2,306 @@
...
+                        );
+                    }
+
+                    // A group with an empty label separates the options with a line instead
+                    if (group.label === '') {
+                        return (
+                            <div
+                                {...getItemProps({
+                                    className: selectListBlock('item', {separator: true}),
+                                    style,
```
jhoncool, line 197: With an empty-label group and `getOptionGroupHeight`, this inline height is treated as the row estimate, but the separator CSS adds 8px of block margins and a 1px border outside it. A requested height of H therefore occupies H + 9 while `measure={false}` virtualization and active-row alignment still use H. Please keep the separator geometry inside the declared border-box height or include it consistently in both rendering and estimation.

AI generated

src/components/Select/components/SelectList/SelectList.tsx
```diff
@@ -2,162 +2,306 @@
...
+            return {
+                ...outerVirtualization,
+                estimateItemSize: (ctx: ListItemContext<SelectListNode>) =>
+                    getItemHeight(ctx.item, ctx.index),
+            };
+        }, [outerVirtualization, getItemHeight]);
+
+        const handleItemAction = React.useCallback(
+            (_id: string, node: SelectListNode) => {
+                if (isSelectGroupNode(node) || node.value === LOADING_OPTION_VALUE) {
```
jhoncool: `SelectOption.value` accepts any string, so a consumer option whose value is `__SELECT_LIST_ITEM_LOADING__` is mistaken for the internal sentinel here: it renders as a loader, skips `getOptionText`, and cannot be selected. Please distinguish the loading row by object identity or a discriminated internal node instead of reserving an undocumented public value.

AI generated

src/components/Select/__tests__/Select.base-actions.test.tsx
```diff
@@ -233,130 +235,93 @@ describe('Select base actions', () => {
...
                 expect(selectedItem.textContent).toBe(content);
                 await user.keyboard('[ArrowUp]');
             }
         });
     });
 
-    describe('find elements in flat list by "quick search"', () => {
-        test('with instant input', async () => {
-            const {getByTestId} = setup({options: QUICK_SEARCH_OPTIONS});
+    // The search by the first letters belongs to the List: a prefix, a buffer of 500 ms and the
```
jhoncool: This comment says 500 ms, but `TYPEAHEAD_TIMEOUT` and the documented behavior use a one-second buffer. Please change it to 1 second or refer to the constant without duplicating the numeric value.

AI generated

src/components/Select/utils.tsx
korvin89: 🤖 AI generated

Fixed in 869bb147.

The text of a selected value is asked of the getter only when the options actually hold that value; a value they do not hold yet is shown as it is, which is what happened before `getOptionText` existed. `renderSelectedOption` still receives the `{value}` object, as it always did — that part is the documented contract.

Test: "the trigger shows the value and getOptionText is not asked about it" — a getter that reads `option.data` with empty options.

src/components/Select/utils.tsx
korvin89: 🤖 AI generated

Good catch, and it is the flip side of the group-boundary fix: while a standalone option after a group was counted into it, a header always had something under it. Fixed in 869bb147 — a section with no options of its own is dropped after the tree is built. A group with an empty label stays: it is a separator and heads nothing to begin with.

Test: "a group left without options by the filter loses its header" — group [Alpha] followed by standalone Beta, filtered to Beta.

src/components/Select/components/SelectList/SelectList.tsx
korvin89: 🤖 AI generated

Fixed in 869bb147.

The separator draws its line inside its own box now — `box-sizing: border-box`, the spacing as padding and the line as a `::before` — so the row costs the list exactly the height it declared, whether that came from the default (9px: 4 + 1 + 4) or from `getOptionGroupHeight`. The eight pixels of margin and the border used to stand outside the box while the virtualizer and the alignment counted on the declared number.

src/components/Select/components/SelectList/SelectList.tsx
korvin89: 🤖 AI generated

Fixed in 869bb147.

The loading row carries a mark of its own (a symbol on the node) and is recognised by it everywhere — the render, the item action and the text getter. Its id is built the same way the ids of the sections are, from a prefix no value starts with, so it cannot take the place of a row in the list either.

Test: "an option that took the value of the loading row is still an option" — an option valued `__SELECT_LIST_ITEM_LOADING__` renders as an option and is selectable while the list is loading.

src/components/Select/__tests__/Select.base-actions.test.tsx
korvin89: 🤖 AI generated

Fixed in 869bb147 — the comment names `TYPEAHEAD_TIMEOUT` instead of repeating a number.

src/components/Select/types.ts
korvin89: 🤖 AI generated

Agreed — and it is the next PR of this stack rather than an omission: the README of both languages, a `migration-guide.md` for the component and the README of `useSelectOptions` are the whole of that PR. It will be linked here when it opens, and the stack does not reach `main` without it.

For what it is worth, the documentation is already written against this diff: the virtualization sections, the properties table, the custom-filter example with `inputProps`, a section on the text of an option, and the migration table for everything that changed.

src/components/Select/hooks/useAlignActiveOption.ts
korvin89: 🤖 AI generated

Agreed, and the `TODO` at the top of the hook says the same thing. It is blocked rather than overlooked: the core deliberately does not import the virtualizer, the bridge between them is render-only, and `ListVirtualizer` does not hand out the `scrollToIndex` of its virtualizer — the very API that already implements the settling this hook does by hand.

The shape it should take is an item-id API on the list (`scrollToItem(id, {align})`, plus an opt-in for aligning on mount), and then both this hook and the DOM-id arithmetic leave the Select. It is written down as a follow-up for the list rather than for this PR.

src/components/Select/utils.tsx
```diff
@@ -48,69 +96,181 @@ export const getFlattenOptions = (options: SelectOptions): FlattenOptions => {
...
+
+    if (loading) {
+        nodes.push({...LOADING_OPTION, value: `${prefix}loading`});
     }
 
-    if (typeof option.children === 'string') {
-        return option.children;
+    // A section whose options the filter took away has nothing left to head. A group with an empty
+    // label is a separator rather than a header, and heads nothing to begin with — it stays
+    return nodes.filter(
```
jhoncool: After this filter removes an empty named section, `nodes` no longer has the same flattened row sequence as `flattenOptions`, but `useAlignActiveOption` still receives the latter. With a virtualized Select, a filtered-out group followed by a standalone option shifts both the accumulated offset and every index-sensitive height callback; the initial alignment can scroll to the wrong window or leave the active row unmounted when several groups are pruned. Please derive the alignment rows from the same post-pruning `nodes` structure that is passed to `List`, and add a filtered-group virtualization regression test.

AI generated

src/components/Select/components/SelectList/SelectList.tsx
```diff
@@ -2,162 +2,306 @@
...
-
-        const valueSet = new Set(value);
-
-        return flattenOptions.reduce<number[]>((acc, option, index) => {
-            if ('value' in option && valueSet.has(option.value)) {
-                acc.push(index);
+/** The value is owned by useSelect and changes through onItemAction only */
+const noop = () => {};
+
+const getItemId = (node: SelectListNode) => (isSelectGroupNode(node) ? node.id : node.value);
```
jhoncool, line 60: `SelectOption.value` accepts any string, but this value is now used as the List row id and eventually passed to `encodeURIComponent`. A lone UTF-16 surrogate, for example `String.fromCharCode(0xd800)`, is a valid string value but makes `encodeURIComponent` throw `URIError: URI malformed`, so opening the Select crashes; the previous index-based IDs did not have this failure. Please use a total, collision-free encoder for arbitrary UTF-16 values and cover lone high and low surrogates.

AI generated

src/components/Select/utils.tsx
```diff
@@ -48,69 +96,181 @@ export const getFlattenOptions = (options: SelectOptions): FlattenOptions => {
...
-}) => {
-    const {getOptionHeight, getOptionGroupHeight, size, option, index, mobile} = args;
+/** A string and a number are the text of the option as they are; a node is not */
+const asText = (content: React.ReactNode): string | undefined => {
+    if (typeof content === 'string') {
+        return content;
+    }
 
-    let itemHeight = mobile ? MOBILE_ITEM_HEIGHT : SIZE_TO_ITEM_HEIGHT[size];
+    return typeof content === 'number' ? String(content) : undefined;
```
jhoncool, line 116: The new public contract says that only string content becomes the default option text and that all other content falls back to `value`, but `asText` also stringifies numbers. For `{value: "ten", content: 10}`, the trigger, filter, and typeahead use `"10"` instead of the documented `"ten"`; the current test misses this because its numeric content and value stringify identically. Please either remove the number branch and test distinct values, or update every public contract to explicitly include numeric content.

AI generated

src/components/Select/README.md
```diff
@@ -75,8 +75,64 @@ SANDBOX-->
...
+
+The text of an option is what the trigger shows for a selected option, what the filter matches and
+what the search by the first letters looks up. By default it is the content of the option when that
+is a string, and its `value` otherwise — so an option that renders anything but a plain string needs
+`getOptionText`, or it will be searched and named by its value.
+
+<!--GITHUB_BLOCK-->
+
+```tsx
+import {Select, getSelectOptionText} from '@gravity-ui/uikit';
```
jhoncool: This new public example cannot be copied as written: it uses `Flex` and `Icon` without importing them. The following comment is also stale after the latest fix: `getOptionText` is deliberately not called for a selected value missing from `options`, so this callback never receives that synthetic object without `data`. Please import all used public components and rewrite the comment around a real option without data in both the English and Russian READMEs.

AI generated

src/components/Select/utils.tsx
korvin89: 🤖 AI generated

Right, and it was a real shift: the pruning made the nodes of the list and the flat list of options two different row sequences, and the alignment was summing the second one. Fixed in 7381f5e9 — the rows of the alignment are the nodes of the list flattened back into the order it draws them, so a pruned header is not counted where nothing is drawn.

Test: "a group the filter emptied does not shift the scroll" — a virtualized Select whose first group is filtered away; without the fix the scroll lands one header lower (1046 against 1020).

src/components/Select/components/SelectList/SelectList.tsx
korvin89: 🤖 AI generated

Confirmed and fixed, in the core PR of this stack: https://github.com/gravity-ui/uikit/pull/2825 (commit ad73002f). A lone surrogate really did bring the list down — `encodeURIComponent` is not total.

The escaping of a DOM id is now total and one-to-one: every character outside `[A-Za-z0-9-]` becomes its code unit between underscores, the underscore included, so no two ids can meet. Tests: the existing injectivity case, plus "an id of any string at all still makes a DOM id" with a lone high surrogate, a lone low surrogate and a non-Latin word — it throws `URIError` without the fix.

src/components/Select/utils.tsx
korvin89: 🤖 AI generated

The number is deliberate — a `Select` whose options read `10`, `20`, `50` is a normal one, and the kit has such a Select of its own in `Pagination`; making it fall back to the value would give it a dev warning and the wrong text. So the contract is the one to fix, and it is fixed in 7381f5e9: the JSDoc of `getOptionText` and the READMEs of both languages say "a string or a number" now.

You are right about the test, though — it could not tell the two apart. It now asks for the text of `{value: 'ten', content: 10}` and expects `10`.

src/components/Select/README.md
korvin89: 🤖 AI generated

Both fixed in the working tree of the documentation PR, which is the next one in this stack — `Flex` and `Icon` are imported in the example, in both languages, and the comment no longer talks about a selected value missing from `options`: that case never reaches `getOptionText` any more, so the example explains the real one instead — options that are not all of the same shape, where the ones without `data` keep the default text.

korvin89:
🤖 AI generated

A note on where this lands, since the documentation has come up twice in the review.

@jhoncool  `next` is not the branch production versions of the kit are released from — it collects the v8 major, and nothing merged into it reaches a published package until that major is cut. A PR that lands here without its documentation therefore never leaves a consumer with a README that contradicts the code in an installed version.

The documentation is a PR of its own, the next one in this stack: the READMEs of both languages, a `migration-guide.md` for the component and the README of `useSelectOptions`. It is already written against this diff — the virtualization sections, the properties table, the custom-filter example with `inputProps`, a section on the text of an option and the migration table. It will be linked here when it opens, and the stack does not reach `main` without it.

src/components/Select/components/SelectList/SelectList.tsx
```diff
@@ -2,162 +2,304 @@
...
```
jhoncool, line 60: Duplicate option values now collapse to the last row because getItemId uses node.value while List stores rows in rowById. Both visible occurrences resolve to the same entry, so two options with the same value but different content render the latter twice and produce duplicate DOM IDs. The public Select contract does not require unique values, and the previous implementation preserved both positional rows. Please separate internal row identity from the selection value, or explicitly enforce and document uniqueness before rendering.

AI generated

korvin89: 🤖 AI generated

The uniqueness is the contract, and it is deliberate — but you are right that the diagnostic was wrong, and that part is fixed in c3496d1b.

Why the contract stays: the value of an option *is* the identity of a choice. Two options with one value were never two choices even before this PR — clicking either applied the same value, and in `multiple` both of them would have shown as selected. The rows were two only on screen. Separating an internal row identity from the value would keep those two rows alive, but the ambiguity would remain everywhere it matters: which of them is selected, which one the keyboard lands on, which one `renderOption` is called for when the consumer looks up the value.

The uniqueness is documented — "The `value` of an option identifies its row, so the values have to be unique across the whole list, groups included" in both READMEs, and a line in the migration guide of the component.

What was genuinely wrong: the warning came from `[List]`, a component the consumer never held, and it talked about "item ids". The Select now says it itself and names the value: "More than one option has the value "js". The value identifies the row of an option, so the list renders one row per value — make the values unique." Test: "options that share a value are named in a warning of their own".

jhoncool (approved):

src/components/Select/types.ts
korvin89: 🤖 AI generated

The documentation PR is open: https://github.com/gravity-ui/uikit/pull/2830 — the READMEs of both languages, a migration guide for the component and the README of `useSelectOptions`. It sits on top of this branch, so it merges after it.

src/components/Select/__tests__/Select.popup.test.tsx
```diff
@@ -78,8 +78,9 @@ describe('Select popup', () => {
         expect(popup).not.toBeNull();
     });
 
+    // The row view sizes itself; the heights are what it comes out as (on mobile it is size `l`)
```
amje: It should be XL on mobile

src/components/Select/constants.ts
```diff
@@ -12,27 +12,55 @@ export const selectListBlock = block('select-list');
...
-    s: 28,
+    s: 24,
     m: 28,
     l: 32,
     xl: 36,
 };
 
-export const MOBILE_ITEM_HEIGHT = 32;
+/** A row on mobile is a row of size `l` */
+export const MOBILE_SIZE: SelectSize = 'l';
```
amje: This should be XL on mobile

amje (changes requested):

src/components/Select/constants.ts
korvin89: 🤖 AI generated

You are right, fixed in 7d08c785 — `MOBILE_SIZE` is `xl` now.

Checking where the `l` came from: the old mobile row really was 36px with the bigger text — `SelectList.scss` on `next` gave `#{$block}_mobile &` a label of `$l-height` (36px) and `--g-text-body-2-font-size`, and a mobile group header the same 36px. In the row view of the new list that is `xl` (36px, `text-body-2`); its `l` is 32px with `text-body-1`.

The 32 I started from came from `MOBILE_ITEM_HEIGHT`, which the old code fed to the virtualizer only — it never matched the CSS beside it.

So on mobile an option is 36px again and a group header is 28 (the header of the new design is more compact than the row it heads). The height tests that pin what `renderOption`/`renderOptionGroup` receive follow: 32 → 36 for an option, 26 → 28 for a group.

amje (approved):

PR #2828, opened 2026-09-13
feat(TreeSelect): support default props and custom selected options content

BlackPoretsky:
- support `DefaultPropsProvider`
  - add `renderSelectedOption` prop to override the content of selected options
  - fix selected options with a ReactNode `title` being rendered as `[object Object]`


korvin89:
@BlackPoretsky Hi, the component has been moved to this repository https://github.com/gravity-ui/normalized-list. In the next major version, it will be removed, so it’s better to make the change directly there.

PR #2830, opened 2026-09-14
docs(Select): the list of v8 and the way to it

korvin89:
🤖 AI generated

The documentation of the `Select` that moved onto the new list. Third and last PR of the stack: the core is #2825 (merged), the migration itself is #2826, and this one carries what a consumer reads.

## What is here

**The reference**, in both languages, follows the component:

- **the text of an option** — what the trigger shows, what the filter matches and what the search looks up, and the new `getOptionText` that gives it. The `text` property of an option is gone, and the section says so where its replacement is explained;
- **the search by the first letters** — a prefix, a buffer that lives a second, a space that continues the query, and characters that do not reach the hotkeys of the application;
- **virtualization** — asked for by wrapping the `Select` in `ListVirtualizer` instead of counting options, with the entry point, the optional peer dependency and the development warning above 150 options;
- **groups** — a header is a caption rather than an option, and an empty label draws a line;
- **the uniqueness of the values** — a value identifies the row of its option;
- **the CSS contract** — the class names of the rows are not one. What is supported is the colours of a row through the variables of the row view, and the content through the render props; the geometry follows the size of the `Select`;
- **a custom filter** — wired through `inputProps`, which carries the combobox ARIA of the input along with its handlers. The example hands the parts to `TextInput` one by one, because a component that does not take a spread would otherwise lose the keyboard of the list.

**A migration guide** (`src/components/Select/migration-guide.md`) — a table of what has to be rewritten and why, then the same for the data, the DOM and the styles, and the keyboard. Every row points into the reference, so the two cannot drift apart. The README links to it from the top.

## Notes

- The guide is English-only, as the two guides already in the repository are.
- The heights of the rows changed with the design (an option of `size="s"` is 24px instead of 28, a group header 26 plus 8 above it, a separating line 9) — the guide names them, since `renderOption` receives those numbers.
- No code is touched here.

## Checks

`prettier` over every `.md` of the repository, every anchor of both READMEs and of the guide resolved against the headings they point at, and the statements checked against the code of #2826 file by file.


src/components/Select/migration-guide.md
```diff
@@ -0,0 +1,80 @@
...
+## The data
+
+- **The values of the options have to be unique**, groups included: a value identifies the row of an
+  option. Two options with one value share a DOM id and all such rows show the content of the last
+  one; the `Select` says so in a development warning. The old list kept rows by position and drew
+  both.
+- **The rows are the rows of the new design, and they are of other heights**: an option of `size="s"`
+  is 24px instead of 28, an option on mobile is 36px instead of 32 (there every row is of size `xl`,
+  whatever the `size` of the `Select`), the header of a group is 26px (28 in `xl`) plus 8px above it
+  when it follows other rows, and a group with an empty label — a separating line — takes 9px instead
```
jhoncool, line 51: **#4 Qualify the separator height**

An empty-label group costs 9px only when it follows another row. At index 0, `getPopupItemHeight` returns 0 and the CSS suppresses the separator line, so the current wording is still too broad. Please document the first-row exception, including the `itemHeight` passed to `renderOptionGroup`.

_AI generated_

korvin89: 🤖 AI generated

Right — fixed: the line now says 9px when the group follows another row, and nothing at all as the first row of the list, as it was before. The README already carried that exception in the section on grouped options; the guide did not.

src/components/Select/README.md
```diff
@@ -499,16 +588,17 @@ There are many ways to customize your `Select`.
...
 <!--SANDBOX
 import {Button, Select} from '@gravity-ui/uikit';
 
 export default function () {
     return (
         <Select
-            renderControl={({onClick, onKeyDown, ref}) => (
-                <Button ref={ref} onClick={onClick} extraProps={{onKeyDown}}>
+            renderControl={({ref, triggerProps}) => (
+                <Button ref={ref} extraProps={triggerProps}>
```
jhoncool: Passing `triggerProps` through `Button.extraProps` does not preserve the disabled state: `Button` later writes its own top-level `disabled` value to the native element, overriding `extraProps.disabled`. A disabled `Select` copied from this example therefore renders an enabled, focusable trigger whose click handler can still open the popup. Please destructure `disabled` and pass it directly to `Button`.

_AI generated_

src/components/Select/migration-guide.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+`@gravity-ui/uikit/virtualizer` needs `@tanstack/react-virtual` — it is an optional peer dependency
+of the package, so it has to be installed alongside it. A `Select` without the wrapper renders every
+option as a row, and above 150 of them says so in a development warning.
+
+The width of the popup follows from the same place. A list that virtualized itself by the old
+threshold was as wide as the control; a list that is not virtualized is as wide as its widest option
+(see [Popup width](./README.md#popup-width)). So a long list left without the wrapper can come out
+wider than it used to — `popupWidth="fit"` pins it to the control as before.
```
jhoncool: The old virtualized default enforced a 100px minimum popup width, while `popupWidth="fit"` uses the control width minus its borders. For controls narrower than 100px this recommendation shrinks the popup instead of preserving the previous behavior. Please qualify the advice or recommend `ListVirtualizer` without `popupWidth` when the legacy minimum must be retained.

_AI generated_

src/components/Select/README.md
```diff
@@ -75,8 +79,54 @@ SANDBOX-->
...
+groups included. The DOM `id` of the row is derived from it, and that id is what
+`aria-activedescendant` of the trigger points at — read it from there rather than building it by
+hand: the escaping of a value belongs to the list and is not a contract.
+
 #### Grouped list
 
+A group header is a caption rather than an option: it carries `role="presentation"`, stays out of
+the keyboard walk and out of the count of `role="option"` rows, and describes the options under it
+through `aria-describedby`. A group whose `label` is empty draws a separating line instead of a
+header — as the first row of the list it has nothing to separate and takes no space at all.
```
jhoncool: The zero-height first-row exception only applies to the default group rendering and height. `renderOptionGroup` runs before the empty-label separator branch, and `getOptionGroupHeight` overrides the default zero height, so a leading empty-label group can still render content or occupy space. Please scope this statement to the default behavior and make the same qualification in the migration guide.

_AI generated_

src/components/Select/README.md
korvin89: 🤖 AI generated

Confirmed and fixed. `Button` spreads `extraProps` and then writes `disabled={disabled || loading}` onto the element (`Button.tsx:138,142`), so anything the spread carried under that name is overwritten by the component's own default of `false`.

The example takes `disabled` from the arguments and hands it to `Button` directly now, and the prose beside it says why: a wrapper usually writes its own attributes over whatever was spread onto it.

src/components/Select/migration-guide.md
korvin89: 🤖 AI generated

Right — `getMinWidth` kept a floor of 100px while the list virtualized itself (`middlewares.ts:12-20`), and `popupWidth="fit"` is `controlWidth − 2 × BORDER_WIDTH` with no floor at all, so on a control narrower than 100px the advice made the popup smaller rather than the same.

The guide says both now: wrapping in `ListVirtualizer` brings the old rule back as it was, and `popupWidth="fit"` pins the popup to the control but drops the 100px floor with it.

src/components/Select/README.md
korvin89: 🤖 AI generated

Confirmed and scoped, in both the README and the guide. `renderItem` asks `renderOptionGroup` before it looks at the empty label, and `getPopupItemHeight` returns `getOptionGroupHeight` before the `label === ''` branch — so a leading empty-label group does render and does take space when either of them is given.

The sentence now says that the zero is the default drawing of a group, and that both callbacks are asked first, empty label or not.

jhoncool (approved):

PR #2834, opened 2026-09-16
fix: introduce shared overlay layers and keep tooltips above toasts

### LINKED ISSUE #2056: ClipboardButton — provide tooltip options
Components docs https://preview.gravity-ui.com/uikit/?path=/docs/components-utils-clipboardbutton--docs#copytoclipboardoptions

I would like to be able to set the direction for the tooltip and  rendering it without using a portal.

Right now, this is what it looks like inside the `<ToasterComponent/>`

Image

Demo in codesandbox
https://codesandbox.io/p/sandbox/aged-sun-qfnxp9

kseniya57:
Tooltips inside `ToasterComponent` could appear behind notifications because their z-index was lower than the toaster’s.

Introduce a shared `--g-layer-*` scale for overlays and drag previews. Preserve existing overlay levels while deriving the tooltip level above them. Explicit `Popup.zIndex` and `Popup.floatingStyles.zIndex` overrides retain their precedence.


ogonkov:
Can you make a screenshot test for a case with toaster + tooltip overlap?

PR #2841, opened 2026-09-17
feat(theme): support external default props

amje:
## Summary

- expose `useDefaultProps` and extensible `DefaultPropsMap` from the stable entrypoint
- keep `ComponentDefaultPropsMap` as a deprecated compatibility alias
- document and test package-qualified defaults for components from peer libraries

## Testing

- `npm run lint`
- `npm run typecheck`
- `npm test -- --runInBand` (113 suites, 1460 tests)
- `npm run build`
- `git diff --check`

## Snapshots

No snapshots changed.


src/components/theme/useDefaultProps.ts
```diff
@@ -1,9 +1,9 @@
 import * as React from 'react';
 
-import type {ComponentDefaultPropsMap} from './DefaultPropsProvider';
+import type {DefaultPropsMap} from './DefaultPropsProvider';
 import {DefaultPropsContext} from './DefaultPropsProvider';
 
-export function useDefaultProps<K extends keyof ComponentDefaultPropsMap, T extends object>(
+export function useDefaultProps<K extends keyof DefaultPropsMap, T extends object>(
```
korvin89, line 6: 🤖 **AI generated**

❓ **question**

Now that `useDefaultProps` is part of the public API, would it make sense to tie `T` to the registered entry? Right now `T` is unconstrained, so `useDefaultProps('Button', {totally: 'unrelated'})` compiles. A library that copy-pastes a call and gets the key wrong would silently get another component's defaults merged in. Tightening the signature after release would break typings, so it might be worth deciding now.

I tried this locally:

```ts
export function useDefaultProps<K extends keyof DefaultPropsMap, T extends NonNullable<DefaultPropsMap[K]>>(
```

With it, the full typecheck fails in only one place: `Accordion.tsx:20`, because of variance on the generic `onUpdate` (a cast there would do). `PasswordInput`, which goes through the `'TextInput'` key, still passes, and unknown fields in object literals are caught. If the loose typing is intentional, maybe the docs could say that the key only selects defaults and doesn't validate the shape of the props?

korvin89 (approved):
🤖 **AI generated**

**Severity legend**

PR #2844, opened 2026-09-18
fix(Popover): handle nested popup interactions

Feverqwe:
## What was fixed

Fixed incorrect dismissal of `Popover` when interacting with a nested portaled popup, such as `Select`.

Previously, `Popover` created its `useDismiss` interaction outside the `FloatingTree` created by `Popup`. As a result, the parent dismiss handler could not detect nested floating nodes and treated clicks inside them as `outside-press`.

`Popover` now creates its interactions inside the same `FloatingTree` used by its `Popup` and nested UIKit popups. `Popup` also reuses an existing tree when one is available.

## Testing

Added a regression test covering both `trigger="click"` and `trigger="all"`:

- opens the parent `Popover`;
- opens the nested `Select`;
- verifies that selecting an option calls `Select.onUpdate`;
- verifies that the parent is not dismissed before the option handler;
- verifies that clicking outside still closes the parent with `outside-press`.

Also verified existing `Popup` and `Select` popup tests, including Escape handling.

No public API changes.


src/components/Popover/Popover.tsx
```diff
@@ -140,3 +142,17 @@ export function Popover(rawProps: PopoverProps) {
         </React.Fragment>
     );
 }
+
+export function Popover(props: PopoverProps) {
+    const tree = useFloatingTree();
+
+    if (tree === null) {
+        return (
+            <FloatingTree>
```
amje, line 151: Popover renders Popup, which already does this. It's double wrapping in FloatingTree

amje (changes requested):

src/components/Popup/Popup.tsx
```diff
@@ -385,9 +385,9 @@ function PopupComponent(rawProps: PopupProps) {
 }
 
 export function Popup(props: PopupProps) {
-    const parentId = useFloatingParentNodeId();
+    const tree = useFloatingTree();
 
-    if (parentId === null) {
+    if (tree === null) {
```
amje, line 390: Let's do the same change for Modal then

amje (changes requested):

src/components/Popover/Popover.tsx
Feverqwe: The FloatingTree created inside Popup is too low in the React tree. Since Popover calls useDismiss above it, the hook cannot access that tree context.

src/components/Popup/Popup.tsx
Feverqwe: Done!

amje (approved):

PR #2845, opened 2026-09-21
fix(Select): stop forwarding opening arrow key events to the closing list

Pechenux:
## What was fixed

Fixed a `Maximum update depth exceeded` error when quickly reopening `Select` with ArrowUp after pressing Escape.

Previously, the arrow key event that opened the popup was also forwarded to the previous list, which could still be mounted during the closing animation. This could trigger conflicting active-item updates between `Select` and `List`.

The key handler now returns immediately after opening the popup. `useActiveItemIndex` determines the initial active option, making reopening during the closing animation consistent with opening after the previous list has unmounted.

## Testing

Added a parameterized regression test covering eight scenarios: ArrowUp and ArrowDown, empty and selected values, and reopening both during and after the closing animation.

Each scenario verifies the initial active option, subsequent arrow-key navigation, and selection with Enter.


korvin89 (commented):
🤖 **AI generated**

**Severity legend**

src/components/Select/Select.tsx
```diff
@@ -203,6 +203,7 @@ export const Select = React.forwardRef<HTMLButtonElement, SelectProps>(function
             if ([KeyCode.ARROW_DOWN, KeyCode.ARROW_UP].includes(e.key) && !open) {
                 e.preventDefault();
                 toggleOpen();
+                return;
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

Would you consider making this guard a bit more general? Right now only the opening arrow key is kept away from the closing list. Any other key still reaches the previous List during the 100 ms closing transition. For example, pressing Escape and then End (or Home / PageDown) followed by Enter within that window silently selects an option (`onUpdate` is called with `ruby` / `js` / `python`) and reopens the popup. This already happens on `main`, so this PR doesn't cause it, but it's the same kind of problem.

While the popup is closed, the list is either unmounted or still closing, so forwarding keys to it is never useful. Instead of the early `return`, I'd suggest forwarding keys only while the popup is open:

```ts
if ([KeyCode.ARROW_DOWN, KeyCode.ARROW_UP].includes(e.key) && !open) {
    e.preventDefault();
    toggleOpen();
}
if (e.key === KeyCode.ESCAPE && open) {
    toggleOpen(false);
}

// The previous List may still be mounted during the closing transition,
// so keys must not reach it while the popup is closed.
if (open) {
    listRef?.current?.onKeyDown(e);
}
```

I checked this variant locally: all Select tests still pass, including the new ones, and the stray selection is gone. A short comment like the one above would also keep someone from removing the guard later as redundant.

src/components/Select/__tests__/Select.base-actions.test.tsx
```diff
@@ -88,6 +88,69 @@ describe('Select base actions', () => {
...
+            jest.useFakeTimers();
+        });
+
+        afterEach(() => {
+            cleanup();
+            jest.runOnlyPendingTimers();
+            jest.useRealTimers();
+        });
+
+        describe.each([false, true])('after closing transition completes: %s', (finishClosing) => {
```
korvin89: 🤖 **AI generated**

🟠 **should-fix**

Could we trim this matrix a bit? I reverted the `return;` to see which cases actually catch the bug. Only 3 of the 8 fail with "Maximum update depth exceeded":

- during closing, ArrowUp with `[]`
- during closing, ArrowDown with `['python']`
- during closing, ArrowUp with `['python']`

The other 5 pass with or without the fix:

- **All four `finishClosing: true` cases.** By then the previous List is unmounted and `listRef.current` is `null`, so the key was never forwarded anywhere. The "open without a mounted list" baseline is also already covered by the first half of each test (the very first open after `user.tab()`).
- **During closing, ArrowDown with `[]`.** This one passes only by coincidence: the old List's ArrowDown from an empty state lands on index 0, which is also the initial index, so there is nothing to conflict.

I'd suggest keeping the three cases above and removing the other five, together with `describe.each([false, true])` and the `if (finishClosing)` branch inside the test:

```ts
describe('reopen by arrow key during closing transition', () => {
    // beforeEach / afterEach unchanged

    test.each([
        {key: 'ArrowUp', value: [], initialIndex: 0, nextIndex: 2},
        {key: 'ArrowDown', value: ['python'], initialIndex: 1, nextIndex: 2},
        {key: 'ArrowUp', value: ['python'], initialIndex: 1, nextIndex: 0},
    ])('$key with value $value', async ({key, value, initialIndex, nextIndex}) => {
        // same body, without the `if (finishClosing)` block
    });
});
```

This keeps the same regression coverage with fewer than half the cases, and the test title no longer renders as `after closing transition completes: false`.

korvin89 (approved):

PR #2847, opened 2026-09-21
fix(Alert): prevent content from shrinking

saxumcordis:
## Summary
- allow Alert content wrapper to take available width without shrinking custom message content
- add ContentWidth story covering full-width textarea inside message
- update Alert visual snapshots

## Tests
- npm run typecheck
- npm run playwright -- src/components/Alert/__tests__/Alert.visual.test.tsx


src/components/Alert/Alert.scss
```diff
@@ -88,6 +88,10 @@ $block: '.#{variables.$ns}alert';
         }
     }
 
+    &__content {
+        width: 100%;
```
ogonkov: i think that should be flex-grow: 1 + min-width: 0 here and may be somewhere in parent elements to have proper flexibility

IsaevAlexandr (approved):

